PR #31, opened 2015-11-17
Fix for wrong config file for multiple sites.

abiosoft:
Also, `commands.HugoCmd.Run` isn't using the parameters.


mholt:
:100:

hacdias:
Thank you!

PR #78, opened 2016-06-27
Fix source path in docs

rselbach:
README incorrectly points to source files under `$GOPATH` instead of `$GOPATH/src`. This PR fixes that. Please consider applying.


hacdias:
Thank you.

PR #79, opened 2016-06-27
add note about error when go get'ting

rselbach:
Just adding a quick note so that users don't get confused by the unexpected message during `go get`. Please consider applying this. Thanks.


hacdias:
Thanks ;)

PR #144, opened 2017-07-19
Fix shadow bug on Firefox

princemaple:
box-shadow: 0 0 0 0; leaves a slight edge on Firefox.
box-shadow does accept `none`, which works as expected.


princemaple:
before
image

after
image

hacdias:
Thanks @princemaple! I have already found that bug but I didn't really know where the problem was 😅 Thanks!

PR #147, opened 2017-07-19
Fix typo Costum:Custom

EduardoRT:
Fixed a typo in the frontend of the app, great work! love using this in my personal server.


hacdias:
Thank you 😄 This change will be available after deploying it to the caddyserver. It may take some time.

PR #157, opened 2017-07-27
Add Docker Support

llitfkitfk:
the Dockerfile use Multistage build (currently Docker hub not supported)

Use [PWD](http://play-with-docker.com/) instead to build docker image


hacdias:
Hello @llitfkitfk! I really appreciate what you did here, and I thank you for that. The checks are failing but it is my fault, so don't worry about it. I've this things to ask/comment:

1. I always assure that the files `rice-box.go` are always up to date so I don't think that the `npm run build` and the rice commands are needed. It would only be a problem if I forgot, but I have a `pre-push` git hook to build it for me so I don't forget, ever. What do you think?
2. Is godep really needed? Couldn't we just `go get ./...`? What are the advantages? Disadvantages?

I'll look forward for you answer. 😄 I'm really excited to merge this!!

llitfkitfk:
## rice

I think rice-box.go need to be removed if we use docker build, 
OR just leave it alone for Github releases?

## dep

**pros**: 
 * specify dependency version 
 * simple?  dep init & dep ensure
 * more in [dep](https://github.com/golang/dep) & [video](https://www.youtube.com/watch?v=5LtMb090AZI)

**cons**: 
 *

hacdias:
So, okay, I'll leave GoDep 😋 
About rice, I think that if:

```
FROM node:alpine
WORKDIR /src/app
COPY . /src/app
RUN yarn install && rm -rf assets/dist && npm run build
```

and

`rice embed-go && cd ./caddy/hugo && rice embed-go` 

Were removed, there won't be any problems, since it is on the repository already. I think those are extra unnecessary steps. Could you try it, please?

llitfkitfk:
will try it

llitfkitfk:
done

hacdias:
Hey! Thanks.

One more thing: I'll ask you to remove the godep please. Right now, file
manager is also a Caddy plugin and its build server doesn't support godep
(it go gets all dependencies). Since I'm using Travis I believe I can keep
filemanager keep up to date with the dependencies now. Do you mind?

I appreciate your effort on this, though. But I don't feel it's needed now.
After this change I'll merge the PR. :)

llitfkitfk:
OK

Instruction for docker setup and Options (Volume) for Dockerfile  will update in the future.
Also need a dockerhub organization account for automate build(currently not support Multistage build)

hacdias:
Thanks. The dockerfile still calls some godep commands.

llitfkitfk:
will fix

hacdias:
Thank you :) when I arrive home I'll publish it to docker hub and ping you here.

hacdias:
@llitfkitfk,

~So, I made some small updates (ef315082edf5fb7d28731f80d7e961468d69d6e7) in order to be supported on Docker Hub. And here it is: [hacdias/filemanager](https://hub.docker.com/r/hacdias/filemanager/).~

~I was thinking about changing ENTRYPOINT to:~

```
ENTRYPOINT ["/go/src/github.com/hacdias/filemanager/cmd/filemanager/filemanager", "--port", "80"]
```

~So the user only needed to add `-p somebeautifulport:80` when running since it was already attributed to 80. I don't think there's any problem since it is running inside a Docker image. What do you think?~

~**Edit**: after all, I think that changing the default `CMD` to `["--port", "80"]` would be better than having it to display the help but this should be done in a more automated way.~

~The `scope` should be a VOLUME.~
~The `database` should be an external file too, right?~
~The `config` should also be an external file. How do we get it?~

I'm a bit new to docker 😅 

~I'm trying some stuff here, but I won't push it now because you were the one who had the idea and I'd like your opinion on this.~

**Edit:** got a better idea. I'll post it as soon as I can 😄

hacdias:
@llitfkitfk,

What about:

```
FROM golang:alpine
COPY . /go/src/github.com/hacdias/filemanager
WORKDIR /go/src/github.com/hacdias/filemanager
RUN apk add --no-cache git
RUN go get ./...
WORKDIR /go/src/github.com/hacdias/filemanager/cmd/filemanager
RUN go build
VOLUME /data
ENTRYPOINT ["/go/src/github.com/hacdias/filemanager/cmd/filemanager/filemanager", "--port", "80"]
CMD ["--port", "80", "--scope", "/data/files/", "--database", "/data/database.db"]
```

And the user would run something like:

```
docker run -v="/mystuff:/data" -p 80:80 -it filemanager
```

Where:

- `/mystuff/files` would contain the files
- `/mystuff/database.db` would be the DB

by default. What do you think?

I'm not sure which are the best practices for this

hacdias:
```Dockerfile
FROM golang:alpine

COPY . /go/src/github.com/hacdias/filemanager

WORKDIR /go/src/github.com/hacdias/filemanager
RUN apk add --no-cache git
RUN go get ./...

WORKDIR /go/src/github.com/hacdias/filemanager/cmd/filemanager
RUN go install

VOLUME /srv
EXPOSE 80

COPY Docker.json /etc/config.json

ENTRYPOINT ["/go/bin/filemanager"]
CMD ["--config", "/etc/config.json"]
```

### Paths in the container

- Config: `/etc/config.json`
- Database: `/etc/database.db`
- Base scope: `/srv`

### Default config.json

```json
{
  "port": 80,
  "address": "",
  "database": "/etc/database.db",
  "scope": "/srv",
  "allowCommands": true,
  "allowEdit": true,
  "allowNew": true,
  "commands": []
}
```

### Usage

If the user wants to use the `config.json` file:

```shell
docker run \
    -v /path/to/sites/root:/srv \
    -v /path/to/config.json:/etc/config.json \
    -v /path/to/database.db:/etc/database.db \
    -p 80:80 \
    hacdias/filemanager
```

If the user doesn't use the `config.json` file:

```shell
docker run \
    -v /path/to/sites/root:/srv \
    -v /path/to/database.db:/etc/database.db \
    -p 80:80 \
    hacdias/filemanager
    --port 80
    --database /etc/database.db
    --scope /srv
    --other-flag other-value
```

Does this make sense to you, @llitfkitfk?

hacdias:
Check out this commit (02c9890) 😄

llitfkitfk:
Great!

llitfkitfk:
docker image size  is still large.

After Docker Hub supported, need modify Dockerfile use Multistage build
Or manually build in [pwd](http://labs.play-with-docker.com/) using multistage build and push to docker hub.

hacdias:
I just set it to built with Docker Cloud using the Edge channel and the build was successful (using multi stage 8117eb98d87d5676cde6938d7553fea1fed4f6e0). As you can see [here](https://hub.docker.com/r/hacdias/filemanager/tags/), the compressed size is now 8 MB. 😄 Try it out

llitfkitfk:
Great!

llitfkitfk:
Now Docker Hub  support multistage-build.

hacdias:
Thanks 😄

PR #174, opened 2017-07-29
Last Modified Sort

jazware:
Addresses issue https://github.com/hacdias/filemanager/issues/173

Updated the Listing.vue and file.go files to allow sorting by last modified time.

Also ended up adapting how sorting works so instead of it being by name or not by name, sort can have 3 different values and sorting logic follows that.


hacdias:
Thank you! 😄

PR #219, opened 2017-08-25
Update locale

Equim-chan:


hacdias:
Thanks for the updated :smiley:

PR #236, opened 2017-09-07
Update zh-tw translation

kairen:
Fix some details for Traditional Chinese.


hacdias:
Hello @kairen! Thanks for the updates 😄

PR #237, opened 2017-09-07
Docker image built from scratch

rigon:
What do you think?
Filemanager doesn't need alpine to work.

I've followed some hints from here: https://blog.codeship.com/building-minimal-docker-containers-for-go-applications/

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,16 +7,16 @@ RUN apk add --no-cache git
 RUN go get ./...
 
 WORKDIR /go/src/github.com/hacdias/filemanager/cmd/filemanager
-RUN go build -ldflags "-X main.version=$(git tag -l --points-at HEAD)"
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -7,16 +7,16 @@ RUN apk add --no-cache git
 RUN go get ./...
 
 WORKDIR /go/src/github.com/hacdias/filemanager/cmd/filemanager
-RUN go build -ldflags "-X main.version=$(git tag -l --points-at HEAD)"
+RUN CGO_ENABLED=0 go build -a -ldflags "-X main.version=$(git tag -l --points-at HEAD)"
```
hacdias: This is really nice! Thanks! Could you please just change it to `go build` please? We don't need the ldflags anymore 😄

hacdias (changes requested):

hacdias (approved):

hacdias:
Thanks!

PR #238, opened 2017-09-07
Add new triggers and improve environment variables

maxux:
This pull request is a chunk of my fork I use for some specific use case. I try to split my job to small pull requests to help to improve upstream :)

This pull request add `after_copy`, `after_rename`, `after_upload` and `after_delete` trigger.
Moreover, when a trigger is fired, the trigger name is added to `TRIGGER` environment variable, the username which did the trigger is set as `USERNAME` and the scope of the user is set as `ROOT` (because otherwise when using specific scope, the `cwd` of the script run by the trigger is not correct).

Warning: this pull request don't add the UI access to edit the script called by `after_copy` and `after_rename`

Warning (bis): this change the `file` to `FILE` variable environment. Usualy, environment variables are in uppercase.


hacdias:
Hey @maxux! I really like what you did here! I'm just asking myself: what if you added `before_` handlers for those actions? Do you think they would be useful?

maxux:
Thanks :) Actually, good point, I'll add this !

hacdias:
Thanks! 😄

hacdias:
image

maxux:
Yeah ! Looks perfect !

PR #253, opened 2017-10-13
i18n: auto detect locale from browser

Equim-chan:
How do you like the idea of auto detecting the default locale from the browser?

The implementation is a little complex and breaking though.

### DIFF
--- a/assets/src/i18n/index.js
+++ b/assets/src/i18n/index.js
@@ -36,8 +36,6 @@ switch (true) {
     locale = 'en'
 }
 
-console.log(locale)
-
 const i18n = new VueI18n({
   locale: locale,
   fallbackLocale: 'en',
--- a/assets/src/i18n/index.js
+++ b/assets/src/i18n/index.js
@@ -9,8 +9,35 @@ import zhTW from './zh-tw.yaml'
 
 Vue.use(VueI18n)
 
+let locale = (navigator.language || navigator.browserLangugae).toLowerCase()
+switch (true) {
+  case /en.*/i.test(locale):
+    locale = 'en'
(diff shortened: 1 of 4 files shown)


hacdias:
Isn't that going to replace the user's settings? This is a good idea if the user doesn't set its own language.

Equim-chan:
It should work that way, as `value.locale` is before `navigator.language`.
https://github.com/Equim-chan/filemanager/blob/0271b914a422dd6dbf7cbc843a05f0f45ec585ca/assets/src/store/mutations.js#L30

I also tested. After the user set his own language, the auto detection will be overrided.

The problem is the code isn't so DRY. The same switch case repeated twice in the code.

~~`switch (true)` style comes from go~~

assets/src/i18n/index.js
```diff
@@ -9,8 +9,37 @@ import zhTW from './zh-tw.yaml'
...
+    locale = 'zh-tw'
+    break
+  case /^zh.*/i.test(locale):
+    locale = 'zh-cn'
+    break
+  default:
+    locale = 'en'
+}
+
+console.log(locale)
```
hacdias: Sorry for being picky, I'll just ask to remove this, please

hacdias (changes requested):
LGTM. Just remove that console.log and I'll merge this. And thank you 😄 a lot!
After merging this I'll take care of the "DRY" thing.

assets/src/i18n/index.js
Equim-chan: oh, my bad

hacdias:
Thank you 😄

PR #254, opened 2017-10-13
Don't expose error

Equim-chan:
* When a resource has been moved, the share link should no longer be valid.
* Don't expose error.


hacdias:
Good idea! Thanks for this 😄

Issue #218, opened 2017-08-24
Symlinked or mounted folder not showing up


remoteweb:
Congrats for the great work you 've done for filemanager. 

I'd like to ask if there is any way i can have symlinked or mounted dirs to get shown in file manager.

Say i have the following folder /someplace/folder1 and filemanager has access to /filemanager/files.
When I symlink /someplace/folder1 to  /filemanager/files/folder1 i only see a folder1 file on filemanager.
However if i mount -bind /someplace/folder1 to /filemanager/files/folder1 , i end up with an empty folder though /someplace/folder1 has files.

Best regards,
Aris

hacdias:
I'll investigate this! Thanks 😄

hacdias:
Hello @arisxama! I'm not being able to reproduce this error. However, I'm on Windows so I'll probably install a Linux VM to try it out.

hacdias:
Closed by #281.

ThibautMarty:
Apparently the symlink issue had been solved, but what about the mounted folder issue?
The mount point appears to be empty in filebrowser while it does contain some files.
I'm using latest docker image.

Regards,
Thibaut

ThibautMarty:
Never mind, I guess it is a problem with the docker volume + mount. It should work with the `shared` volume options.

eine:
@ThibautMarty does it work if you start filebrowser in the host, instead of using the docker image?

PR #281, opened 2017-11-14
fix symbolic links issue on Linux.

oopsmonk:
fix issue #218
Cannot create files or folders in symlink directories on Linux.


hacdias:
Hello @oopsmonk! Thanks for this 😃

Issue #303, opened 2017-12-26
Large download kills Caddy


lstrz:
### 1. Have you downloaded File Manager from caddyserver.com? When have you done that?
Yes. November 26 2017.

### 2. What is your entire Caddyfile?
```text
https://my.domain.com
errors /my/home/caddy.err {
        rotate_size 10
        rotate_age 365
        rotate_compress
}
gzip
header / {
        Strict-Transport-Security "max-age=31536000;"
        X-XSS-Protection "1; mode=block"
        X-Content-Type-Options "nosniff"
        X-Frame-Options "DENY"
}
log /my/home/caddy.out {
        rotate_size 10
        rotate_age 365
        rotate_compress
}
root /my/home/public
timeouts {
        read 180s
}
git https://github.com/a_repo.git /my/home/some_dir {
        then hugo --config /my/home/some_dir/config.toml
}
filemanager /path /my/scope {
        database 16b77317de4813d8e2fcdfefa1e92903.db
}
forwardproxy {
        basicauth username pass
        probe_resistance
        hide_ip
}
minify
```

### 3. What are you trying to do?
Download a 1.66 GB file which was uploaded successfully.

### 4. What did you expect to see?
Once the file is selected and the download button is pressed, I'd expected to see a "Save file" kind of a pop-up.

### 5. What did you see instead (give full error messages and/or log)?
After selecting the file and pressing the download button, another page started loading and after a few seconds it stopped loading. Caddy was no longer running on the VPS it was running on previously and had to be restarted. Nothing was logged. The VPS has 512 MB RAM only, so the file doesn't fit in it -- maybe that caused the problem?

To me, it doesn't seem like a timeout issue, as I wouldn't expect Caddy to crash after timing out.

### 6. How can someone who is starting from scratch reproduce this behaviour as minimally as possible?
1. Upload a big file (potentially bigger than RAM available).
2. Try downloading the big file.

hacdias:
Probably the problem is exactly what you're saying (the RAM). I'll try to find a way to download it in chunks. THanks for reporting 😄

hacdias:
Hopefully this was fixed in #303. We're now streaming the file using `io.Copy` instead of using `http.ServeFile` which was, I think, putting it all in memory and then serving it to the user. I'm publishing the new version (1.3.12) and I ask you to test it and reopen this issue if the error persists (I hope it doesn't 😮)

hacdias:
[...]

PR #330, opened 2018-01-13
http: download: improvement for range support and I/O

Equim-chan:
The main benefit of [http.ServeContent](https://golang.org/pkg/net/http/#ServeContent)  over io.Copy is that it handles Range requests properly, sets the MIME type, and handles If-Match, If-Unmodified-Since, If-None-Match, If-Modified-Since, and If-Range requests. This is helpful when downloading under an unstable network condition where a download connection may be cut off. With http.ServeContent, suspended downloading process can be resumed properly.

[http.ServeFile](https://golang.org/pkg/net/http/#ServeFile) is just a wrapper over http.ServeContent which handles path resolve.

In downloadHandler, it's a bit hard to adopt http.ServeContent, but at the moment it's not necessary to store the archived filed in the first place; it's an overhead that we can reduce.

NOTE: try to reproduce #303 to make sure the stream is properly adopted.

### DIFF
--- a/http/download.go
+++ b/http/download.go
@@ -1,7 +1,6 @@
 package http
 
 import (
-	"io"
 	"net/http"
 	"net/url"
 	"os"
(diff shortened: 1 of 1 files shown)


hacdias:
Thank you @Equim-chan. I run this with `top` and downloaded a 2GB files and it didn't seem to use more than 1MB RAM so I think it's good!!

PR #333, opened 2018-01-18
http: header: fix wrong header

Equim-chan:
`X-Content-Type` should be `X-Content-Type-Options`
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Content-Type-Options


hacdias:
Thank you @Equim-chan!

PR #339, opened 2018-01-21
Minor fixes.

Equim-chan:
`build.sh` was using CRLF for line wrap, making it not gonna be working under Linux. So I ran `dos2unix` on it to change it into LF.

---

https://github.com/hacdias/filemanager/blob/93566aceef7fa68b7a5009257448716c643f3c0b/http/auth.go#L40-L45
It can just be
```go
var data struct {
	Success     bool        `json:"success"`
}
```
other fields aren't necessary.

---

https://github.com/hacdias/filemanager/blob/93566aceef7fa68b7a5009257448716c643f3c0b/http/http.go#L225-L233
Render with `ReCaptchaSecret`!? Are you serious? Isn't it "secret"? :smile:


hacdias:
Thanks! I'm going to merge this and release a new version. 😄 👍 

> Render with ReCaptchaSecret!? Are you serious? Isn't it "secret"? 😄

Oh, well... Never noticed that. No one would see it unless they changed the templates too so...

PR #340, opened 2018-01-23
http: download: use PathEscape instead of QueryEscape

Equim-chan:
Sorry it was my mistake.

For example `file+name with space` should be escaped into `file+name%20with%20space` (PathEscape) instead of `file%2Bname+with+space` (QueryEscape), which causes wrong unescaped filename.


hacdias:
Thanks!

PR #358, opened 2018-02-18
feat: edit every file

### LINKED ISSUE #314: Edit files without extension

Currently double clicking is the only way to open the editor in order to modify the content of a file. However, it does not work if the file has no extension. Instead, it shows a black box covering all the page which suggests to download it. I think opening any file as text should be allowed.

Somehow related to this, I don't know if using double click to open folders and files is good for smartphones and other touch devices. Since there is already a 'Select multiple' activation option and both `Ctrl+Click` and `Shift+Click` work on desktops, I suggest making a single click open the file/folder. Moreover, this is coherent with default browsers' behaviour.

> What if we try to open a binary file without extension?

There is no problem at all to open a binary file with a text edior. In the worst case you get non-understandable characters, e.g.:

```
E  èx"  …íÇ   …¶ýÿÿ1ÀH‡é¬ýÿÿH‰Ñÿ‡~  é&ýÿÿH‹E  H‹
ôD  Ç   è9"  émýÿÿE1íéæþÿÿ‰ÁèÍ!  @ Hƒì(H‹5E  Ç    èZ
  è¥üÿÿHƒÄ(Ãfffff.„     Hƒì(H‹E  Ç     è*
  èuüÿÿHƒÄ(ÃUH‰å]Ãf.„     UH‰åHƒì Hƒ= +   t0H
‡:  ÿ±}  H…Àt/Hƒ:  H‰Áÿ¤}  H…Àt	H
```

Moreover, some binary files use extensions such as `.bin`, `.run`, `.exec`... Therefore, I think it is better to not filter by name at all.

Shall JavaScript or golang provide some mechanism to identify if a file is going to be readable by a human, a filter can be applied based on that guess. The filter would show a warning letting the user know that the opened file is expected to be hardly readable.

hacdias:
Close #314.

### DIFF
--- a/lib/file.go
+++ b/lib/file.go
@@ -248,37 +248,17 @@ func (i *File) GetFileType(checkContent bool) error {
 		return nil
 	}
 
-	if strings.HasPrefix(mimetype, "text") {
-		i.Type = "text"
-		goto End
-	}
+	i.Type = "text"
 
-	if strings.HasPrefix(mimetype, "application/javascript") {
-		i.Type = "text"
-		goto End
-	}
-
-	// If the type isn't text (and is blob for example), it will check some
-	// common types that are mistaken not to be text.
-	if isInTextExtensions(i.Name) {
-		i.Type = "text"
-	} else {
-		i.Type = "blob"
-	}
-
-End:
 	// If the file type is text, save its content.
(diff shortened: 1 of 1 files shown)


hacdias:
Close in favour of #575

PR #368, opened 2018-02-26
alternative ReCaptcha, close #366

### LINKED ISSUE #366: proposal: alternative ReCaptcha
Since nearly the whole Google is blocked in China, `https://www.google.com/recaptcha` is not an exception. However, `recaptcha.net` is [offical](https://s26.postimg.org/x735zdiyh/image.png) and [accessible from China](https://en.greatfire.org/https/recaptcha.net). Just change
```
https://www.google.com/recaptcha/api.js?render=explicit
```

to
```
https://recaptcha.net/recaptcha/api.js?render=explicit
```

in front end, and
```
https://www.google.com/recaptcha/api/siteverify
```

to
```
https://recaptcha.net/recaptcha/api/siteverify
```

in back end, and it should work as expected.

We may add an option, `alternative_recaptcha` in Caddy plugin, `--alternative-recaptcha` in CLI, to switch to `recaptcha.net` from `www.google.com` as host in both recaptcha handling (backend) and serving (frontend).

Equim-chan:
See #366.

Sub-PRs:
* https://github.com/filebrowser/frontend/pull/3
* https://github.com/filebrowser/filebrowser.github.io/pull/2

### DIFF
--- a/caddy/parser/parser.go
+++ b/caddy/parser/parser.go
@@ -49,6 +49,7 @@ func Parse(c *caddy.Controller, plugin string) ([]*filebrowser.FileBrowser, erro
 		scope := "."
 		database := ""
 		noAuth := false
+		alterRecaptcha := false
 		reCaptchaKey := ""
 		reCaptchaSecret := ""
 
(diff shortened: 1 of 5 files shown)


Equim-chan:
Done, and proven to be working.

image

Equim-chan:
Just a bit off-topic though, I don't know whether I should feel happy or sad that we, and many other fellows, have to make so many extra efforts for those who are not yet able to embrace the freedom that they deserve.

hacdias (approved):

hacdias:
Thanks for this @Equim-chan 😄 This is awesome eheh

hacdias:
@Equim-chan I know I've just merged this, but I'd like to ask you to change the option 'alternative_recaptcha' to 'recaptcha_host' so it is more direct. And do the same on https://github.com/filebrowser/filebrowser.github.io/pull/2 😺

Equim-chan:
I don't really agree. `--recaptcha_host recaptcha.net` is longer than `--alternative-captcha`, and requires more parsing that is likely to fail if the user passes an invalid recaptcha host. The possible valid values are only `www.google.com` and `recaptcha.net`, so we can just use a boolean. recaptcha.net is also not very known.

hacdias:
Oh okay. I thought there would be more alternatives

sainiaj:
https://github.com/filebrowser/filebrowser/issues/3697
@Equim-chan can you please help me i keep getting the error message - unknown flag

PR #369, opened 2018-02-27
share: 6 bytes version of share link, close #331

### LINKED ISSUE #331: Propose: shorter share link
Currently the hash of share link is 32 bytes long and 64 characters long in hex. It's not really necessary to use a hash that long, even 16 bytes is enough. Or, adopt Base64 encoding instead of hex?

https://github.com/hacdias/filemanager/blob/94648c92a2d159d8edebb06adb9cd7462aa4374c/http/share.go#L70-L75

Equim-chan:
Well, seems easier than expected.

New share link's hash is 6 bytes long, with a capacity of 256 ** 6 == 281,474,976,710,656. Fair enough imo.

### DIFF
--- a/http/share.go
+++ b/http/share.go
@@ -1,7 +1,7 @@
 package http
 
 import (
-	"encoding/hex"
+	"encoding/base64"
 	"net/http"
 	"path/filepath"
 	"strconv"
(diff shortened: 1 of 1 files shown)


hacdias (approved):

hacdias:
Nice, let's go for it. Thanks 😄

Equim-chan:
ping @hacdias, I think it's ready to release a new version now.

PR #392, opened 2018-04-05
Fix typo preventing "--allow-edit" from being set

mdlawson:
As title. Would be great to get this fixed, spent a while trying to work out how I could possibly be doing this wrong before realising!


Equim-chan:
Thanks.

PR #397, opened 2018-04-17
fix309: merge CMD in the dockerfile to ENTRYPOINT

### LINKED ISSUE #309: A single command line argument overrides all the defaults

### 1. Have you downloaded File Manager from caddyserver.com? If yes, when have you done that? If no, and you are running a custom build, which is the revision of File Manager's repository?

No. I am using the 'vanilla' docker image: `hacdias/filemanager`.

### 3. What are you trying to do? and 4. What did you expect to see?

I am trying to set a single command line argument, but keep default values for every other parameter. I'd expect that changing a single argument won't affect any other.

### 5. What did you see instead (give full error messages and/or log)? and 6. How can someone who is starting from scratch reproduce this behaviour as minimally as possible?

For example, although the default port is `80`, setting `--no-auth` produces an apparently random port number:

```
$ docker run --rm -itv $(pwd):/srv -p 80:80 hacdias/filemanager
Listening on [::]:80
^C
$ docker run --rm -itv $(pwd):/srv -p 80:80 hacdias/filemanager --no-auth
Listening on [::]:34665
^C
$ docker run --rm -itv $(pwd):/srv -p 80:80 hacdias/filemanager --no-auth
Listening on [::]:42178
```

Workaround:

```
$ docker run --rm -itv $(pwd):/srv -p 80:80 hacdias/filemanager --no-auth -p 80
```

However, this does not happen with ports only:

filemanager_defaults

---

> I have to check this to see what's going on. Not sure if it's caused by Viper or something I've done wrong but I don't remember having any issues with that.

eine:
Fix #309.

This PR moves `"--config", "/config.json"` in the Dockerfile from CMD to ENTRYPOINT, in order to prevent these arguments being overriden.

Note that right now Viper does not support multiple config files:

> https://github.com/spf13/viper#reading-config-files
> Viper can search multiple paths, but currently a single Viper instance only supports a single configuration file.

Therefore, if a user provides an additional `-c` flag, parameters in `.config.json` are completely ignored. Only the last passed config file will be used.

Something like spf13/viper#442 would be really useful in this use case.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -21,5 +21,4 @@ EXPOSE 80
 
 COPY Docker.json /config.json
 
-ENTRYPOINT ["/filebrowser"]
-CMD ["--config", "/config.json"]
+ENTRYPOINT ["/filebrowser", "--config", "/config.json"]


hacdias:
Thank you @1138-4EB 😄

Issue #342, opened 2018-01-23
Consider moving to its own organisation


hacdias:
Let's consider moving this repository to its own organisation, where we can separate better the front-end from the back-end and, eventually, add more contributors 😄

Equim-chan:
Wow! I was _just_ about to propose this!! Yeah just now when I opened the issue page! :smile:

I'm making more changes at the moment, but I can hold on until the organisation is possibly established.

There will be a few things to do if an organisation is to be established:
* [ ] Change import paths in Go source files
* [ ] Update documents for all `go get` cases
* [ ] Migrate the project's main page to its new home
* [ ] Re-submit Caddy plugin

hacdias:
Hello @Equim-chan, so: you can keep doing your changes and this will be moved later. There are still things to settle such as CI flow and such because I'm thinking about splitting this main package into some smaller ones 😄 

Also, we'd have to choose a name for the organisation since `filemanager` is not available.

Equim-chan:
Most of my changes are trivial, so I don't prefer them to be in one single PR, nor merged into one single commit. So I think I'd wait.

hacdias:
Any suggestion for the name?

Equim-chan:
I'm also thinking of it at the moment...

Equim-chan:
If we adopt a good organisation name, do we need to change the project title as well?

Equim-chan:
* **file-manager**? _simply adding a bar_
* **filemand**? _file manager daemon, sounds classic_
* **easyfile**? _clear but not somewhat off-topic_
* **efile**? _short but not clear_
* **teamfile**? _a bit random_
* **fileer**? _almost meaningless, but short_
* **filebrowser**? _this matches the original name more_
* **gophiles**? _gopher and files, I like it_
* **gophila**? _a more proper name for an organisation, go with phila (love)_
* **gofile**? _neat and clear name too_

I'm not really good at giving names though. :smile:

hacdias:
I vote on `file-manager` or `filebrowser`. Don't vote for `gophiles` because the front-end could be easily (I think) reused to other back-ends (like IPFS, which is one of my goals).

arteta22000:
I vote for filebrowser

hacdias:
Vote here, please #343

hacdias:
Ping @Equim-chan @arteta22000

Equim-chan:
May I ask for access permissions on the repos? I should be able to assign some issues.

[...]

Issue #224, opened 2017-08-31
Document Filemanager API


ferronrsmith:
The Web API is nice. It would help if this was documented so that it can be integrated into other systems. I have a distributed FS and use `filemanager` has the front-end for all of this. 

When a user is created in our system we create one in filemanager user with a scope. So that it acts like a multi-tenant system. It's beautiful. We had to read through the web app to + code find endpoints.

Should have been documented.

hacdias:
😮 Yes, I have to write the docs for the web API. There was someone else who asked me the same too.

PR #399, opened 2018-04-18
drop package.json, use yarn instad of npm, add build_in_docker.sh

eine:
I found no info about how one should proceed to modify the frontend and rebuild filebrowser. It seems that `package.json` is used, which requires to push resources either to the npm or to a git repo. Moreover, because `node_modules` is removed in `build.sh`, any local changes are deleted.

This PR follows a different approach. `frontend` is expected to be a subdir of this repo, but ignored through `.gitignore`. When `build.sh` is executed, the frontend is cloned if not present. `dist` is always cleared. Then, dependencies of the frontend are downloaded/updated and the frontend is built. Last, because the path to the rice box is changed in ` filebrowser.go`, the content of `dist` can be packed directly.

For those users that don't want to install nodejs and npm/yarn locally, `build_in_docker.sh` is provided. It runs a `golang:alpine` container, installs the required dependencies and executes `build.sh`. Indeed, an image name `hacdias/filebrowser:dev` can be provided, in order to avoid installing the tools each time.


hacdias:
@1138-4EB how do you find this simpler/easier? `.sh` scripts don't run natively on Windows, for example.

eine:
Well, it is not necessarily simpler/easier. Indeed, the objective of this issue/PR is to discuss it, as it seems that there is not info to help contribution (yet).

Right now you assume that a contributor has `nodejs`, `npm`, `go`, `git` and docker. My final approach is to require `docker` and a `shell`. All the other (nodejs, npm/yarn, go and git) are distributed in a single container (`filebrowser/filebrowser:dev`): see [this](https://github.com/1138-4EB/filebrowser/commit/49a8abdf41646d5a32711a3c6dc17fcba8b53efe) commit, that is one ahead of this PR.

My proposed workflow is: https://github.com/1138-4EB/community/blob/fa41dbb7036ee4958e0c4e16ad7eb0d0536f31dd/contributing.md If the user wants to install all the tools locally, `build.sh` can be used. If he/she wants to install nothing but docker, `build_in_docker.sh` produces exactly the same result. After the first execution of any of them, frontend will be checked out, so the user/contributor can edit any source (either from the frontend or the backend) and have everything rebuilt without internet connection.

Also, instead of downloading go dep and rice each time, these are part of the dev image. Moreover, when dep ensure is executed, the vendor folder is kept in the host, in order to avoid downloading all the dependencies each time.

Besides, I find it really weird to use a package.json for the backend. Executing npm install in the filebrowser repo installs many resources in node_modules, apart from filebrowser-frontend, which is the single dependency. How are you supposed to edit something in the frontend and have it rebuilt along with the backend? Are you supposed to move to filebrowser/node_modules/filebrowser-frontend and run npm build there? Or do you edit the frontend in a different path, then push to GitHub and update from filebrowser?

In my proposal, frontend is a child of filebrowser, but it is directly cloned with git in a script. Therefore, the location is still fixed, but all the nodejs stuff is kept in frontend. On the whole, I find it more "natural" to move to subdir frontend in order to build the frontend and then move to filebrowser to build the backend.

---

> .sh scripts don't run natively on Windows, for example.

A GitHub user/contributer is likely to have Git for Windows, MinGW, MSYS2, Cygwin... On Windows 10 there is even the Linux Subsystem available. Furthermore, any docker container can be used to run a shell script. If you think that it is required, I can add the equivalent to `docker run --rm -tv $(pwd):/src -w /src -v /var/run/docker.sock:/var/run/docker.sock busybox sh -c "./build_in_docker.sh` in PowerShell.

hacdias:
> How are you supposed to edit something in the frontend and have it rebuilt along with the backend?

There is `npm link`/`yarn link` for this kind of cases.

---

I think it's a good idea, although I'd like to get some feedback from someone else. @Equim-chan what do you think about this?

eine:
> There is `npm link`/`yarn link` for this kind of cases.

I didn't know about these. Then, if you prefer, I can revert https://github.com/1138-4EB/filebrowser/blob/3acc5fb43feabea42eefece9a4fda313b6155bbf/build.sh#L7-L15 and keep `package.json`. I still think that it can be slightly misleading, but that's just a very personal opinion. In the end, I think the most important contribution in this PR is to provide `build_in_docker.sh`, so that specific dependencies or compile sequences do no matter.

Let's hear what @Equim-chan and others think about it!

Equim-chan:
While I do agree the use of yarn is a great idea, I have several concerns about it.

* May the new Dockerfile breaks the automatic build?
* Is it a better idea to make frontend as a git submodule? Like, in this repo we make the frontend as a submodule locked on some release tag, and when developing we pull the submodule to its HEAD? When we are going to tag a release, we also tag the frontend and pull the submodule to the tag.

eine:
> May the new Dockerfile breaks the automatic build?

Do you mean the automatic build feature in hub.docker.com? If so, I think that you are ok, and this breaks it. There are two possible solutions:

- Rename the Dockerfile I propose to Dockerfile.prod and keep an all-included Dockerfile to be used by hub.docker.com.
- User travis-ci or any other CI service that allows to build and also run tests. On successful builds/tests, artifacts can be pushes to both GitHub releases and hub.docker.com.

I like the second approach because it allows more flexibility and the local workflow can be reproduced. The first solution requires to keep two parallel workflows: an optimal one (from the network usage point of view) and the all-in-one.

However, pushing images from travis from hub.docker.com requires to set a dockerid (user and password) through travis encrypted variables. It is a one-time configuration: https://github.com/ghdl/ghdl/blob/master/dist/linux/travis-enc-dockerhub.sh

BTW, a organization named `filebrowser` can be created in hub.docker.com, just as you did here. That would make it easier to provide both development and production images. Also, it'd be straightforward to provide multiple production images from travis, e.g., one based on scratch and another one based on alpine.

> Is it a better idea to make frontend as a git submodule? Like, in this repo we make the frontend as a submodule locked on some release tag, and when developing we pull the submodule to its HEAD? When we are going to tag a release, we also tag the frontend and pull the submodule to the tag.

I did not explicitly use the term 'submodule' because I have never used this feature of git. However, yes, I was thinking of a submodule when I moved `frontend` to be a subdir of `filebrowser`.

---

About versions, note that the proposed changes include getting the latest available version of dep, not a hardcoded one. I don't know if this is ok or if you prefer to manually bump dependency versions.

hacdias:
I'm starting to think that moving this project to its own organisation (#342) was not that a good idea. Since everything is now on different repositories, the development slowed down and it is actually harder to do.

I didn't want to take back that decision, but I'm asking your opinion @1138-4EB and @Equim-chan. If we moved it back, I'd still leave the same people with Pull Access but we could also call it File Manager again, which was the original and better name.

I'm just dropping some thoughts here. I'll wait for your opinion.

eine:
@hacdias, @Equim-chan, this is going to be long, but I think it is worth. Please, bear with me.

In general, I think that moving the project to its own organization was a good idea, and we should stick with it. It provides a generic name, not linked to any specific user, so it is easier to remember. It is also coherent with many other projects in GitHub, such as ghdl/ghdl, VUnit/vunit or portainer/portainer.

However, I believe that the migration to a organization structure is not finished yet. #342 was just the first step. Now, we need to improve and polish it, in order to make development and collaboration easier.

We might merge filebrowser/filebrowser and filebrowser/frontend back. That would make local development 'easier' at first sight. Yet, it wouldn't be 'clean'. Those two repos are clearly decoupled, as one contains frontend dependencies (Vue, Js, CSS, HTML) and the other one is the backend (Go). By keeping them separated we are stating that alternative implementations can be made for any of them. For example, a different frontend can be written using Angular, but keeping the same backend. Equally, a different backend can be written using Python or Ruby, but keeping the same frontend. This is very valuable, because it lets collaborators focus on/reuse only the parts the need, instead of forcing them to analyze the whole project. However, for it to be a real possibility we should provide a good documentation of the API between them (see #224). I think we should use some service such as [Swagger](https://en.wikipedia.org/wiki/Swagger_(software)) ([explore](https://app.swaggerhub.com/search) existing projects). That's just one option. Docker's API docs ([docs.docker.com/engine/api/v1.30](https://docs.docker.com/engine/api/v1.30/)) used to be built with a Hugo Theme ([moxiegirl/old-docs-base](https://github.com/moxiegirl/old-docs-base)). An interesting up-to-date theme for Hugo is [DocuAPI](https://themes.gohugo.io/docuapi/).

Furthermore, it'd be desirable to create a third repo named filebrowser/caddy, in order to make it clear that filebrowser is a standalone application. I believe that many users might find it useful without Caddy, even though it is most known for being a plugin of it.

Indeed, related to you comment about the name, I think that filebrowser is better than filemanager, because the most notable feature as a standalone app is that this is a web-based solution. But, at the same time, it makes sense to use 'filemanager' in the context of Caddy. Anyone using Caddy already knows that everything is web-based, so 'browser' is redundant. As a result, I would stick with filebrowser for the project name, but use filemanager as the name of the Caddy plugin. Then, the filebrowser/caddy repo proposed above could be named filebrowser/filemanager.

Should we push in that direction, we need to ensure that contributing to any of the three legs feels comfortable enough. At least as comfortable as it was when all of them where hosted in a single repo. This is not hard to achieve, because there was no documentation at all, so the first contact has always been raw. Therefore, this is a good point to elaborate helper scripts until we get a pleasant experience, before actually writting and updating the docs to provide the first version of contribution guidelines.

---

Following this motivation, and according to the conversation we had a couple of months ago (previous comments in this thread), I have further elaborated this PR:

- I created an organization in [hub.docker.com/u/filebrowser](https://hub.docker.com/u/filebrowser/). I am the owner now. But, if you want to take control over it, just let me know and I will transfer it to you.
- I set up [travis-ci.com/1138-4EB/filebrowser](https://travis-ci.com/1138-4EB/filebrowser/) for my fork. I cannot do it with this repo because I am neither a owner nor a collaborator.
- I encrypted my docker credentials with travis, in order to support pushing docker images from travis-ci.com to hub.docker.com. See `./build/enc-dockerhub.sh`.
- I added the frontend repo as a git submodule of the backend. See https://blog.github.com/2016-02-01-working-with-submodules/#advice-on-using-submodules-or-not
- I explicitly separated the following tasks:
  - Getting dependencies and building the frontend (`build_assets.sh`).
  - Building the backend (`build.sh`).
  - Building the docker image (`build_img.sh`).
- I created an image named [filebrowser/dev](travis-ci.com/1138-4EB/filebrowser), to support development/collaboration from hosts with a single dependency: docker. A script is provided to build it (`build_img_dev.sh`).

---

This is the proposed workflow:

- A CRON job in travis-ci builds filebrowser/dev once a week. If successful, it is pushed to [hub.docker.com/r/filebrowser/dev](https://hub.docker.com/r/filebrowser/dev/). This is also executed when a new commit is pushed to branch `img-dev`.
- A CRON job in travis-ci builds filebrowser/filebrowser every day. If successful, it is pushed to [hub.docker.com/r/filebrowser/filebrowser](https://hub.docker.com/r/filebrowser/dev/). This is also executed when a new commit is pushed to branch `master`.

See https://travis-ci.com/1138-4EB/filebrowser/builds

---

With the proposed scheme, all of the following development environments are supported:

## No dependencies

If the developer does not want, or cannot, install any dependency at all, sources can be edited through GitHub.

## All the dependencies

If the developer is willing to install `nodejs`, `yarn`, `go`, `git`, `docker` and `bash`:

- `./build/build_assets.sh` gets frontend and backend dependencies, builds the frontend and updates `rice-box.go`.
- `./build/build.sh` builds the backend.
- `./build/build_img.sh` builds the `filebrowser/filebrowser` image.

Alternatively, a script named `build_all.sh` executes `build_assets.sh` and `build.sh`, one after the other.

NOTE: the repo must be cloned in a path inside GOPATH.

## Docker and bash

If the developer wants to install `docker` and `bash` only:

- `USE_DOCKER="true" ./build/build_all.sh` executes `build_assets.sh` and `build.sh` inside a `filebrowser/dev` container, instead of running them on the host.
- `./build/build_img.sh` builds the `filebrowser/filebrowser` image.

NOTE: the repo can be cloned anywhere.

## Docker and bash, interactive

If the developer wants to install `docker` and `bash` only, but desires to execute `build_assets.sh` and `build.sh` separately, he/she can work inside a container:

```
  WORKDIR="/go/src/github.com/filebrowser/filebrowser"

  $(command -v winpty) docker run -it \
    -v /$(pwd):${WORKDIR}:z \
    -w /${WORKDIR} \
    -p 8000:80 \
    sh -c "\
      cd build && \
      dos2unix build_assets.sh && \
      dos2unix build.sh \
      sh
    "
```

Exposing a port allows to execute filebrowser inside the container after the execution of `build.sh`, and browse it from the host.

NOTES:
  - The repo can be cloned anywhere.
  - Since the repo in the host is mounted inside the container, the developer can use any text editor/IDE in the host to edit the sources (say Atom, notepad++, Visual Studio Code, Sublime...)

## play-with-docker

Any of the three options above can be used in a [play-with-docker](https://labs.play-with-docker.com/) playground.

## No connection

After a first execution of `build_assets.sh`, dependencies (`vendor` and `frontend/node_modules`) are saved in the host. Therefore, as long as `filebrowser/dev` is already pulled, no internet connection is required for development. This applies to any of the options above.

## PowerShell

If the developer is using a Windows box without any bash shell, the interactive option above can be used by slightly modifying the syntax. Then, there is a single dependency: docker.

---

Additional issues:

- Is `goreleaser` affected by any of these changes?
- How is `publish.sh` affected by these changes?
- Should we modify `build_img.sh` in order to programmatically produce two copies of `dockerfiles/filebrowser`? One based on `scratch` and another one based on `alpine`. Only the first line (`FROM`) needs to be changed with sed.
- Should we provide ARM variants of the images? This can be applied to `filebrowser/filebrowser` and/or to `filebrowser/dev`.
- Right now, the latest versions of `node_modules`, `vendor` and `dep` are got. I don't know if this is desirable, or if it'd be better to use fixed/tagged versions and manually decide when to update dependencies.

Last, I think that this message can be a valid starting point as a `Contributing` section, either in [filebrowser/community](https://github.com/1138-4EB/filebrowser-community/blob/patch-1/contributing.md) or in filebrowser.github.io.

---

Unfortunately, I am working abroad for some months, so I won't have much time to actively develop the other pending issues I have (see [filebrowser/issues/created_by/1138-4EB](https://github.com/filebrowser/filebrowser/issues/created_by/1138-4EB) and [1138-4EB/hwd-ide](https://github.com/1138-4EB/hwd-ide)). However, I am open to discussing this PR and providing support to the issues that might arise as a result of merging it.

Indeed, because this PR is already functional in my fork ([1138-4EB/filebrowser](https://github.com/1138-4EB/filebrowser) and travis-ci.com/1138-4EB/filebrowser) you/we can start using it there to see how it works, before actually merging the PR here.

---

Caveats:

- `dockerfiles/dev` relies on the GitHub API to retrieve the latest version of `dep`. If it is triggered frequently, it might hit the rate limit: https://developer.github.com/v3/rate_limit/ This is not critical, because it is set to run once a week. But I think it is good to point it out, just in case. A possible workaround is to fall back to a hardcoded version when it happens. However, I have not implemented this yet.

hacdias:
*Sorry* for this delay. Now I finally have some time to dedicate to this project!

Yes, you could add me permission for that Docker Hub thing, but you can stay there too, of course. 😄 

> Is goreleaser affected by any of these changes?

I don't believe there are any issues with it.

> How is publish.sh affected by these changes?

I think it can remain the same.

> Should we modify build_img.sh in order to programmatically produce two copies of dockerfiles/filebrowser? One based on scratch and another one based on alpine. Only the first line (FROM) needs to be changed with sed.

Hmm, why would we?

> Should we provide ARM variants of the images? This can be applied to filebrowser/filebrowser and/or to filebrowser/dev.

Yes, but sometimes there are some problems compiling for some combinations of PLATFORM+ARM.

> Right now, the latest versions of node_modules, vendor and dep are got. I don't know if this is desirable, or if it'd be better to use fixed/tagged versions and manually decide when to update dependencies.

The node_modules are set on the package.json so they are fixed. For Go, we have Gopkg.

eine:
> Sorry for this delay. Now I finally have some time to dedicate to this project!

I hope you had a good year!

> Yes, you could add me permission for that Docker Hub thing, but you can stay there too, of course. :smile:

You should be in the 'owners' group/team now. Please, let me know if you can edit anything, e.g. https://hub.docker.com/u/filebrowser/dashboard/settings/

About `publish.sh`, I think that a single line needs to be changed, because `build.sh` will not exists anymore when this PR is merged: https://github.com/filebrowser/filebrowser/blob/master/publish.sh#L5 It is a somehow minor issue, but let's not oversee it.

> Hmm, why would we?

There are some use cases where having a shell in the container is desirable. See https://github.com/filebrowser/filebrowser/pull/441#issuecomment-400317290 and https://github.com/filebrowser/filebrowser/pull/441#issuecomment-400473947. I would go with the last option: provide a single scratch image and document how to create others. But I am open to any discussion about it.

> Yes, but sometimes there are some problems compiling for some combinations of PLATFORM+ARM.

Then, let's keep it apart for now. We can follow the discussion about building ARM variants in #440. Should you have any specific information about those conflicts/problems, please add references there.

> The node_modules are set on the package.json so they are fixed. For Go, we have Gopkg.

For node_modules, I was considering to add `yarn.lock` to the frontend repo. Just as we do with `Gopkg.lock` for the backend.

Also, see the previous comment about getting the `go dep` binary:

> dockerfiles/dev relies on the GitHub API to retrieve the latest version of dep. If it is triggered frequently, it might hit the rate limit: developer.github.com/v3/rate_limit...

hacdias:
> There are some use cases where having a shell in the container is desirable. See #441 (comment) and #441 (comment). I would go with the last option: provide a single scratch image and document how to create others. But I am open to any discussion about it.

I think the instructions are enough.

> For node_modules, I was considering to add yarn.lock to the frontend repo. Just as we do with Gopkg.lock for the backend.

Sure. We just have to decide between yarn or npm.

.travis.yml
```diff
@@ -0,0 +1,23 @@
...
+  provider: script
+  skip_cleanup: true
+  script:
+    - echo "$DOCKER_PASS" | docker login -u "$DOCKER_USER" --password-stdin && docker push filebrowser/filebrowser && docker logout
+  on:
+    repo: 1138-4EB/filebrowser
+    branch: master
+env:
+  global:
+  - secure: YtxLrM3z2A/wDh5yLXrMXgsMRr+c7v5auS/2SpNxra0E6fcFKNPtk2g6JB4JfpFRp6feeuwtT61b16OdHjNVcwtZPLEdM1o7nUaLd+UxxKYlXwMSl+Hqaclx5KbBwlVCU3BKJjPVY/RujjSgOGG7C9b40E8YBssVIxyM51qV+u4C7Odu99ZC42OnFG8n+SsdP2xdI7BFw5qJTmNAgnns4QEAiZi++YMtsYmoo828mNpYX093N1VeW81JHeHo82Xwifpeij2fxuOkGAf2JMOuYrH+Mn+jgAiSUGkNRKhaA5do3ov6tSiCWZtQjrAIFk56knXaeeIO+ysryJ+BO9w2qquI1IYktpIemZjjoz24qKeUY0suxPdEe7cH17/P6n2LHUyXiUyUpjRX2bFqo6kfAoMc6oPqgOWu8Ls5GwuEHqCwsZXTVy4ZqNYiAIgk85lUecm9uy+Ubtu4bCdJpDlkC9jGow4Wn/6lHNxsuGWfBR1P94WtC+qmrZf/xje3A5xKATtZtLrfc0tLdsH5VYDyIZ4WYWOIrCemayjXK3RJxN+qQDxaPWPG0cmHAxPslUyHE1eDxy1hCkCZx82Qjr1dX8Xx5tO4SoBayV9pmJo4+IyEb711NQyYrA/EdDswMmrVJNLBY+yDK6umkMlcIuRkVQanOydgRrusbKuQoAGOQRE=
```
hacdias: @1138-4EB just a bit confused: what's this?

eine: In order to push docker images from travis-ci to hub.docker.com, docker credentials need to be set. These are added as encrypted variables through the travis CLI (a ruby gem you use locally). Those two `secure` strings are the hashes corresponding to envvars `$DOCKER_PASS` and `$DOCKER_USER`. When a build is executed, travis decrypts those and "ensures" that are not shown in the logs.

For further info see (in this order):

- https://docs.travis-ci.com/user/docker/#Pushing-a-Docker-Image-to-a-Registry
- https://docs.travis-ci.com/user/environment-variables#Encrypting-environment-variables
- https://docs.travis-ci.com/user/build-stages/share-docker-image/

1138-4EB: It is to be noted that those specific hashes correspond to my repo and the docker user I am using. Therefore, when this PR is merged, two new hashes should be created by i) a GitHub user that has write permissions to write in this repo and ii) using a Docker user that has write permissions to hub.docker.com/u/filebrowser.The same person needs to know both credentials, in order to avoid sharing passwords between different people.

build/enc-dockerhub.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/sh
+
+# - Run `$(command -v winpty) docker run --rm -it alpine sh -c "REPO='https://github.com/<user|organization>/<repo>'; USERNAME=<username>; PASSWORD=<password>; $(cat travis-enc.sh)"`
```
eine, line 3: Related to the comment about secure envvars above:

It is to be noted that those specific hashes correspond to my repo and the docker user I am using. Therefore, when this PR is merged, two new hashes should be created by i) a GitHub user that has write permissions to write in this repo and ii) using a Docker user that has write permissions to hub.docker.com/u/filebrowser.

The same person needs to know both credentials, in order to avoid sharing passwords between different people. Therefore, you should be the one that does it, or you should temporally give write permissions in this repo to me.

This script (`enc-dockerhub.sh`) is meant to make this configuration step easier. The procedure should be as follows:

- Remove the `secure` part of the `.travis.yml` file.
- Either:
  - Push the previous change to the repo and run this script as is.
  - Do not push anything and modify the run command to mount the local folder.
  - Just run the content of this script locally, i.e., install the travis CLI in your machine.
- Follow the interactive steps explained by the CLI.
- Push the resulting `.travis.yml` with the two new hash strings to the repo.

eine: A couple of additional notes on this topic:

- Encrypted environment variables in travis only work for own branches, not for PRs. This ensures that third parties won't be able to push potentially harmful images to the docker hub.
- In my fork I am using travis-ci.com, because they are migrating open-source projects that used to be hosted in travi-ci.org (see https://blog.travis-ci.com/2018-05-02-open-source-projects-on-travis-ci-com-with-github-apps). However, this repo still is active in travis-ci.org. I suggest to change this before creating the new hashes.

hacdias: > Therefore, you should be the one that does it, or you should temporally give write permissions in this repo to me.

You have **write** permissions to all repositories on this organisation 😄

> However, this repo still is active in travis-ci.org. I suggest to change this before creating the new hashes.

Just disabled it on the .org website. Although, on .com, it still says it is active on .org. Maybe we need to wait.

hacdias:
@1138-4EB just noticed you ignored rice-box. That is something I really wanted to do. Although, what is holding me back is Caddy: it can't build the dependencies on the fly on the build server.

Before merging this PR, we should create a repository which holds the content of [caddy](https://github.com/filebrowser/filebrowser/tree/master/caddy) folder and on each release, it must contain the most update `rice-box` possible.

Altough I'm not sure if that's possible since `rice-box` was supposed to be part of the filebrowser package and not the caddy one.

build/enc-dockerhub.sh
eine: > You have write permissions to all repositories on this organisation :smile:

Had not tried it. Thanks!

> Just disabled it on the .org website. Although, on .com, it still says it is active on .org. Maybe we need to wait.

I dropped them an email.

eine:
> @1138-4EB just noticed you ignored rice-box. That is something I really wanted to do. Although, what is holding me back is Caddy: it can't build the dependencies on the fly on the build server.

Yes, I did it because in the current proposed scheme the frontend is built in a travis stage (test) and, on releases, it is cached to be used in a different stage (release). As a result, goreleaser produced an error because it considered the repo to be in a dirty state.

It is possible to automatically push that rice-box to the new caddy repo, by using a deployment key. This would be desirable as it would allow to automatically trigger a build in caddy each time we push tag a to filebrowser. However, I need some more info about the caddy build server. Could you, please, point me in the right direction?

> Before merging this PR, we should create a repository which holds the content of caddy folder and on each release, it must contain the most update rice-box possible.

I already moved the content to the new caddy repo (locally). But I'll wait until I have a clear understanding of the big picture before pushing it.

eine:
I am thinking that a workaround for the caddy/rice-box thing can be to activate travis for the frontend too, and have the artifacts (a tarball of the 'static' folder and the rice-box file) deployed to GitHub Releases. But I don't know if caddy supports fetching files that are not in the git repo. Indeed, this problem affects not only rice-box, but filebrowser as a whole. If we can't add filebrowser as a submodule of caddy, it might be hard to update the plugin with each release. Any thoughts?

hacdias:
@1138-4EB caddy build server only `go get`s the repo so we have to work with that unfortunately. That's why I've been keeping rice-box on version control.

> Indeed, this problem affects not only rice-box, but filebrowser as a whole.

If we have rice-box already compiled, we don't need the frontend on caddy build server.

benitogf:
@1138-4EB  why use yarn instead of npm?

eine:
@benitogf, mostly because of npm/npm#18434 and npm/npm#17621. But I am not a js developer so I might be outdated. Any strong feeling to prefer npm?

hacdias:
*From https://github.com/filebrowser/filebrowser/issues/420#issuecomment-403294003:* I don't have any idea for the Caddy build server issue unless we keep ´rice-box´ on the version control.

benitogf:
@1138-4EB not sure if those issues are active on old versions but on npm 6.1.0 / node 10.6.0 works fine perhaps indicating the required node/npm version would be enough?

eine:
@hacdias the last test I did was putting `rice-box.go`in filebrowser/caddy, instead of saving it in filebrowser/filebrowser. In the last proposed change for `publish.sh` (which is renamed (to `build/release.sh`](https://github.com/filebrowser/filebrowser/blob/develop/build/release.sh) in branch [`develop`](https://github.com/filebrowser/filebrowser/tree/develop)), the release procedure is as follows:

- Manually tag the commit that is to be used in filebrowser/frontend.
- Execute `release.sh`, which will:
  - Check if the tag exists in filebrowser/frontend.
  - Update the submodule to the tag.
  - Replace `untracked` with the semver/tag in `filebrowser.go` (through `sed`).
  - Tag filebrowser/filebrowser.
  - Revert the semver/tag in change in `filebrowser.go` and commit again.
- When the tag in filebrowser/filebrowser is pushed, travis-ci detects it and:
  - The frontend is built, thus `rice-box.go` is created.
  - All the releases are built and published with goreleaser.
  - A single image is created, tagged/named twice and published with goreleaser.
  - [HERE]

Therefore we could add a last step in [HERE], that automatically adds a commit on top of filebrowser/caddy in order to add `rice-box.go` and do any other compatibility task.

However, I have not been able to build filebrowser/caddy, being filebrowser/filebrowser a dependency of it (even without removing `rice-box.go`). Should you help me achieve this, I think that the next steps are quite straightforward.

Note: once we get filebrowser/caddy and filebrowser/filebrowser decoupled, the semver/tag replacements with `sed` could be removed from `release.sh` and executed in travis.

eine:
@benitogf, as long as NPM 6 works as expected, I don't mind. Are the populated `node_modules` folders compatible? Is it possible to use both of them interchangeably?

Edit: the proposed development image is based on `golang:alpine`, which is itself based on `alpine`. The latest `nodejs` version available now is [8.11.3](https://pkgs.alpinelinux.org/package/edge/main/x86/nodejs), and `npm 5.6.0`. Moreover, in the official docker image for nodejs 10.6.0 based on alpine, yarn is still being installed:

- https://hub.docker.com/_/node/
- https://github.com/nodejs/docker-node/tree/master/10/alpine

Therefore, should both tools not be compatible, as long as this PR is concerned, I think that we should wait until npm 6.x is introduced in alpine, before going back to npm.

Alternatively, if the relevant content/layers from image `node:alpine` could be added to image `golang:alpine`, we would probably have npm 6.x available now.

hacdias:
@1138-4EB did you check if it works adding the ´rice-box´ to filemanager/caddy only?

eine:
@hacdias, the problem I had was due to conflicting `vendor` folders. If `filebrowser/caddy` is just a subdir of `filebrowser/filebrowser`, there is a single vendoring space. However, if it is split and dependencies are fetched for both of them, when you try to build/run `filebrowser/caddy` it complains.

This is my scheme:

- filebrowser-hub/
  - caddy/
    - .git/
    - ...
  - community/
    - .git/
    - ...
  - filebrowser/
    - .git/
    - frontend/
      - .git/
      - ...
    - ...
  - logo/
    - .git/
    - ...

I start a `golang:alpine` container from `filebrowser-hub`:

``` bash
docker run --rm -it \
  -v $(pwd):/go/src/github.com/filebrowser \
  -w /go/src/github.com/filebrowser/caddy/filemanager \
  golang:alpine sh
```

Just `apk add git` and run `go get ./...`:

``` bash
/go/src/github.com/filebrowser/caddy/filemanager # apk add git
fetch http://dl-cdn.alpinelinux.org/alpine/v3.7/main/x86_64/APKINDEX.tar.gz
fetch http://dl-cdn.alpinelinux.org/alpine/v3.7/community/x86_64/APKINDEX.tar.gz
(1/5) Installing libssh2 (1.8.0-r2)
(2/5) Installing libcurl (7.60.0-r1)
(3/5) Installing expat (2.2.5-r0)
(4/5) Installing pcre2 (10.30-r0)
(5/5) Installing git (2.15.2-r0)
Executing busybox-1.27.2-r11.trigger
OK: 19 MiB in 19 packages
/go/src/github.com/filebrowser/caddy/filemanager # go get ./...
# github.com/filebrowser/caddy/parser
../parser/parser.go:251:32: cannot use db (type *"github.com/asdine/storm".DB) as type *"github.com/filebrowser/filebrowser/vendor/github.com/asdine/storm".DB in field value
../parser/parser.go:252:31: cannot use db (type *"github.com/asdine/storm".DB) as type *"github.com/filebrowser/filebrowser/vendor/github.com/asdine/storm".DB in field value
../parser/parser.go:253:31: cannot use db (type *"github.com/asdine/storm".DB) as type *"github.com/filebrowser/filebrowser/vendor/github.com/asdine/storm".DB in field value
```

I can remove the vendor folder from `filebrowser/filebrowser/vendor` and it will then get dependencies successfully and also build:

``` bash
/go/src/github.com/filebrowser/caddy/filemanager # rm -rf ../../filebrowser/vendor/
/go/src/github.com/filebrowser/caddy/filemanager # go get ./...
/go/src/github.com/filebrowser/caddy/filemanager # go build filemanager.go
/go/src/github.com/filebrowser/caddy/filemanager # ls /go/pkg/linux_amd64/github.com/filebrowser/caddy/
filemanager.a
```

But I don't know how to test that. Note that the library was built without building filebrowser itself. Indeed, rice-box does not exist, because the checked out version is `master` in my fork.

I pushed the modified sources I am using to the develop branch in `filebrowser/caddy`, so that you can reproduce the setup I explained.

hacdias:
@1138-4EB I'm not sure if Caddy server supports vendor folders either. Let us ask @mholt 😄

mholt:
>  I'm not sure if Caddy server supports vendor folders either. Let us ask @mholt 😄

It sort of does, but with some caveats -- you have to be very careful that none of the vendored packages are vendored by Caddy itself or other plugins it is built with, and also make sure that the vendored packages don't export any types used by Caddy.

Kind of hoping/wondering if Go modules (vgo) will help solve the problem a little bit.

eine:
@hacdias, @mholt, my main doubt is not about vendor folders. Although not ideal, we can work around it by removing all of them and using "absolute" dependencies only. What I need to understand is how much of filemanager/filebrowser is required in order to build the plugin for Caddy. It seems that the plugin can be built even with an incomplete filebrowser codebase:

> https://github.com/filebrowser/filebrowser/pull/399#issuecomment-403868140
> Note that the library was built without building filebrowser itself. Indeed, rice-box does not exist, because the checked out version is [master in my fork](https://github.com/1138-4EB/filebrowser).

```
/go/src/github.com/filebrowser/caddy/filemanager # go build filemanager.go
/go/src/github.com/filebrowser/caddy/filemanager # ls /go/pkg/linux_amd64/github.com/filebrowser/caddy/
filemanager.a
```

Does this mean that Caddy builds the plugin and filebrowser separately? Or should we expect filebrowser to be completely embedded into Caddy?

The former would be ideal for us, since we would only need to instruct the Caddy builder to build `filebrowser/caddy/filemanager` and get the latest released binary from `filebrowser/filebrowser`.

mholt:
All Caddy plugins are completely embedded into Caddy, rather than distributed separately. (If that helps to clarify.)

eine:
I think I found a solution!

```
docker run --rm -itw /go/src/github.com/filebrowser -p 2015:2015 golang:alpine sh -c "$(cat test.sh)"
```

where `test.sh` is:

``` bash
apk update
apk add git vim

git clone --recurse-submodules https://github.com/filebrowser/filebrowser
rm -rf filebrowser/caddy

git clone https://github.com/filebrowser/caddy
cd caddy
go get -v ./...

mv ../filebrowser/rice-box.go ./filemanager/
sed -i.bak 's#package filebrowser#package filemanager#' ./filemanager/rice-box.go

go get -v github.com/mholt/caddy/caddy
go get -v github.com/caddyserver/builds
cd /go/src/github.com/mholt/caddy/caddy
go get -v ./...

sed -i.bak 's#// This is where other plugins get plugged in (imported)#_ "github.com/filebrowser/caddy/filemanager"#' caddymain/run.go

go run build.go

echo "Hello world!" > index.html

./caddy -host localhost filemanager
```

Therefore, I think we can safely remove subdir `caddy` and `rice-box.go` from this repo, as long as `rice-box.go` is ensured to exist in `filebrowser/caddy/filemanager`. Now we need to decide how to update it when a new rice-box is built in travis:

- Use a encrypted deploy key to update the file and create a commit in `filebrowser/caddy`. This is potentially dangerous, as we are giving travis write access to that repo. We can push it to a different branch. E.g., we keep the development related commits in `develop` and `master` is where the `rice-box.go` is added.
- Deploy it to GitHub Releases of `filebrowser/frontend`. In order to support this, `build.go` in `mholt/caddy/caddy` should support copying the file at compile time.

hacdias:
@1138-4EB Awesome!

I think the first solution is the best: a commit to a branch and create a Pull Request automatically. We'd just need to approve it. We just need to make sure that key *isn't able* to commit to master to prevent someone to change Travis.yml and make a change on a pull request.

hacdias:
That for Filemanager plugin. WHat about Hugo and Jekyll? Would we need to replicate the file on three directories?

eine:
> I think the first solution is the best: a commit to a branch and create a Pull Request automatically. We'd just need to approve it.

AFAIK, that's not possible. The scope of 'Deploy keys' is the repository. These can be read-only or read/write, as if it was any other contributor/admin. There is no option to allow writing to a single branch.

Moreover, I don't know how to automatically create a PR. What I propose is to create a pair of SSH keys locally, upload the public one to GitHub (settings of repo filebrowser/caddy); then, encrypt the private one and add it to the codebase in filebrowser/filebrowser; last, remove the created pair so that it does not exist unencrypted anymore. I don know if this SSH key would allow to create a PR. I suppose that it can be done with GitHub's API, but I believe that it requires a different workflow (such as a personal access token).

In order to achieve the behaviour you suggest, we would need to use two separate repos. The sources would be hosted in `filebrowser/caddy` and that's where the development would be done. Then, we would push to a different one, say `filebrowser/filemanager`. The latter would be a deployment-only repo (without issues or PRs), which means that we don't mind if the content is messed up, because we can always remove it without losing any relevant data.

See https://developer.github.com/v3/guides/managing-deploy-keys/#deploy-keys

> We just need to make sure that key isn't able to commit to master to prevent someone to change Travis.yml and make a change on a pull request.

This is supposed not to be a problem, because encrypted variables and files do not work for PRs from forks. The risk would be maintainers not being cautious. We should take care when accepting PRs that change the build procedure (i.e. what is executed in travis). See:

- https://docs.travis-ci.com/user/encrypting-files/
- https://github.com/1138-4EB/ghdl-io/blob/master/travis/travis-enc-deploy.sh
- https://docs.travis-ci.com/user/pull-requests#Pull-Requests-and-Security-Restrictions

Note that this is slightly different compared to how we encrypt docker credentials:

- https://docs.travis-ci.com/user/environment-variables/#defining-encrypted-variables-in-travisyml

However, the risks concerning PRs and possibly wanted pushes are the same.

> That for Filemanager plugin. WHat about Hugo and Jekyll? Would we need to replicate the file on three directories?

I have never used those plugins, so I don't know what's the interface. Do they rely on the GUI of filebrowser or do they just execute commands after an API request?

hacdias:
What if we create a user, like, FilebrowserBot, which has read and write access except to master?

hacdias:
The other plugins are basically the same as filemanager's but with other functionalities enabled so they also need the assets to run.

eine:
> What if we create a user, like, FilebrowserBot, which has read and write access except to master?

That's a good idea. Indeed, it seems that [Machine users](https://developer.github.com/v3/guides/managing-deploy-keys/#machine-users) are meant just for that. We can combine it with [branch restrictions](https://help.github.com/articles/about-branch-restrictions/) in order to let the bot write to a single branch. Furthermore, should we create it, we could use it for hub.docker.com too.

However, we need an e-mail account in order to register the bot both in GitHub and in Docker. @hacdias can you please handle it?

> The other plugins are basically the same as filemanager's but with other functionalities enabled so they also need the assets to run.

I made some additional tests:

First, I added the rice-box in `/go/src/github.com/mholt/caddy/caddy`, by slightly modifying the script I showed [above](https://github.com/filebrowser/filebrowser/pull/399#issuecomment-409804234). This works: [fb_caddy.sh.txt](https://github.com/filebrowser/filebrowser/files/2258761/fb_caddy.sh.txt)

Then, I did the same, but plugging in the three plugins (filemanager, hugo and jekyll): [fb_caddy_all.sh.txt](https://github.com/filebrowser/filebrowser/files/2258765/fb_caddy_all.sh.txt). It is properly compiled too. Could you please check if the functionality is correct?

Note that, even though it might work, it is not a desirable solution, because we cannot push to mholt/caddy. It is just a test to verify that it doesn't really matter in which package we add the rice-box. It is only required for it to be compiled once into the binary.

I also tried adding it three times, once in each of the plugins: [fb_caddy_all_three.sh.txt](https://github.com/filebrowser/filebrowser/files/2258777/fb_caddy_all_three.sh.txt). This does not work. The same content is embedded three times. Therefore, it compiles ok, but it fails at runtime:

``` bash
panic: EmbeddedBox with name `./node_modules/filebrowser-frontend/dist` exists already

goroutine 1 [running]:
github.com/GeertJohan/go.rice/embedded.RegisterEmbeddedBox(0x1078559, 0x28, 0xc4201fda80)
	/go/src/github.com/GeertJohan/go.rice/embedded/embedded.go:77 +0x18d
github.com/filebrowser/caddy/hugo.init.1()
	/go/src/github.com/filebrowser/caddy/hugo/rice-box.go:221 +0x224b
```

So, last, I tried adding the rice-box to `filebrowser/caddy/parser`: 
[fb_caddy_all_parser.sh.txt](https://github.com/filebrowser/filebrowser/files/2258803/fb_caddy_all_parser.sh.txt). It is properly compiled and there is no error at runtime. I think that we should go with this solution. Do you agree?

eine:
Just a note to remember enabling `gosec` with excluded options (those related to using MD5):

- https://github.com/securego/gosec#selecting-rules
- https://github.com/securego/gosec/issues/148#issuecomment-355534317

Need to check how to provide options in `.gometalinter.json`.

hacdias:
> However, we need an e-mail account in order to register the bot both in GitHub and in Docker. @hacdias can you please handle it?

Will do.

> Note that, even though it might work, it is not a desirable solution, because we cannot push to mholt/caddy. It is just a test to verify that it doesn't really matter in which package we add the rice-box. It is only required for it to be compiled once into the binary.

We can do a "dot" import or "_" import (not sure which one it is) of the package containing the assets so it isn't repeated twice.

hacdias:
@1138-4EB created the user. Created two teams on the repo (maintainers and ci) and maintainers have write access everywhere and ci only to caddy repo and conditionally:

github com_filebrowser_caddy_settings_branches_master

There must be a way to create a pull request from the CI. Many bots do that. [Take a look at this](https://hub.github.com/hub.1.html).

hacdias:
@1138-4EB we could do better: have a repo with just the built assets. The only repo @FilebrowserBot would have write access. No problems with permissions. And we could just import it since the package just needs to be present, right? 😄

hacdias:
I'll just try and see if it works here

hacdias:
@1138-4EB could you move what you have on https://github.com/1138-4EB/filebrowser/tree/master to a branch here and change this pull request to be based on that branch? That way we could already start configuring Travis and I could also commit if needed.

hacdias:
@1138-4EB I'm merging this on [chore/build](https://github.com/filebrowser/filebrowser/tree/chore/build) and then I'll open a new PR with this as a reference. That way we an work directly on this repo with Travis and everything else correct, okay?

eine:
@hacdias branch `develop` in this repo is a single commit ahead of branch `master` in my fork. So, it is/was meant for that. Anyway, I will:

- Apply the commit on top of `chore/build`.
- Remove `develop` and `master` from my fork.
- Push branch `img-dev` from my fork to here as `chore/devenv`.

Then, I will follow with encrypting the credentials of the bot, and have Travis properly set up.

PR #405, opened 2018-04-23
fix #311, fix #312 (see filebrowser/frontend#7)

### LINKED ISSUE #311: Commands: response for non-existing commands

Somehow related #310, when executing a command in the container (say `git` or `svn`), the response is an empty object, `{}`. It makes sense because the command does not exist. However, I think that the interface should show an error notification/dialog to let the user know about it.

> It's strange that you receive an empty object. I'll check that and show a better message.

### LINKED ISSUE #312: Commands: parser

I believe that the first word of the text box is checked in order to decide if it is a command or a search term. As a result, if `hugo` is enabled as a command, I cannot use `hugo` as the folder name for the sources of hugo sites within projects.

At the same time, commands need to be available in the path, because `/opt/bin/hugo` is considered a search.

I'd suggest using a prefix, say `cmd:` to tell commands and search terms apart, and allow the admin to set regular expresions as cmd filters. I.e., put all the custom tools in `/opt/custom` and allow `--commands "/opt/custom/.*"`

>[hacdias](https://github.com/hacdias/filemanager/issues/302#issuecomment-354128743)
> A prefix sounds good and makes sense.

eine:
Close #311.
Close #312.
See filebrowser/frontend#7.

### DIFF
--- a/http/websockets.go
+++ b/http/websockets.go
@@ -57,13 +57,14 @@ func command(c *fm.Context, w http.ResponseWriter, r *http.Request) (int, error)
 	allowed := false
 
 	for _, cmd := range c.User.Commands {
-		if cmd == command[0] {
+		if regexp.MustCompile(cmd).MatchString(command[0]) {
 			allowed = true
+			break
 		}
 	}
 
 	if !allowed {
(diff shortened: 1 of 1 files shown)


hacdias:
Thanks! 😄

PR #406, opened 2018-04-23
Rename manager to browser

eine:
Just replace remaining ocurrences of `Manager` to `Browser`, and `fm` to `fb`.

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,11 +1,11 @@
 ### Instructions (remove before submitting):
 
-1. Are you asking for help with using Caddy or File Manager? Please use our forum instead: https://forum.caddyserver.com.
+1. Are you asking for help with using Caddy or File Browser? Please use our forum instead: https://forum.caddyserver.com.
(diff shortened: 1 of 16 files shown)


hacdias:
@1138-4EB did you keep the LF line endings? It seems to have changed the whole file.

eine:
I am using a windows box, but git should automatically handle it. Indeed, I am trying to change it, but as soon as I `git add` the files with LF line endings, they are changed and nothing is staged.

---

Note that this only happens with markdown files, not with go sources. However, I used the same machine/context to submit both commits. Furthermore, with `--ignore-space-change` differences are properly shown.

eine:
@hacdias, I think that these markdown files have CRLF line endings, and I changed them to LF. This is the commit that changed previous LF endings to CRLF (oct 30): https://github.com/filebrowser/filebrowser/commit/99c8c92c6c6d1225380dbbfc5b61d4263a129156#diff-04c6e90faac2675aa89e2176d2eec7d8

hacdias:
Thank you @1138-4EB! 😄

PR #462, opened 2018-07-17
Update .dockerignore to ignore the .git directory

mozzbozz:
This results in a smaller build context being sent to the Docker daemon and therefore also in a faster build process. No need for the git-history and metadata to be used by Docker.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,2 +1,3 @@
 testdata/
 .github/
+.git/


hacdias:
Thanks! 😄

DeyV:
it should be **.git
to ignore all .git in all subfolders

https://github.com/filebrowser/filebrowser/commit/3d3191b1541710bb416d3bcf2e8b7e7682f1a0ab#r29813412

hacdias:
Done @DeyV! Thanks!

Issue #471, opened 2018-07-27
Support SSO through HTTP Headers


maweck:
Hello together and thank you for providing this great project.

we are planning to use "filebrowser" internally, but sadly miss some features. The first one is SSO. All our tools are running behind a proxy which handles the authentication and authorization. If the user is allowed to access the tool, the proxy sets the "X-Forwarded-User" header (and some additional ones). The tools are looking for this header and fully trust its value if found - Meaning that if the header is provided, the user will automatically be logged in.

I already did a quick and dirty POC with "filebrowser" and from my current point of view it could be implemented quite easily. My first questions would be: 
- Is this feature also interesting for the filebrowser project?
- Are there things we would need to take care of right now when we would implement this?

Thank you and best regards

eine:
If I don't get it wrong, what you are suggesting would be potentially dangerous as it should only be used through a properly configured proxy. If not, any user providing any `X-Forwarded-User` will have access. Am I wrong?

Nevertheless, I think that this is an interesting feature. Should you be willing to contribute a PR to implement this, I suggest to please:

- Be explicit about the risk in the codebase.
- Make it default to be disabled.
- Add a message/log when filebrowser starts in this mode.
- Consider adding an example configuration for a proxy such as NGINX or Traefik (or the one you are actually using).

Should it be possible to use any mechanism that allows filebrowser to check whether it is being executed behind a (known) proxy, that'd be great. But I don't think this is compulsory.

maweck:
Hay @1138-4EB , 

> If I don't get it wrong, what you are suggesting would be potentially dangerous [..] If not, any user providing any X-Forwarded-User will have access

Yes, if e.g. the proxy is not correctly configured or if the proxy can be bypassed. So I am more than fine with adding a warning about that feature.

> Should you be willing to contribute a PR to implement this

That would be the plan :) My question here would be: How do you want the options to look like? Should I create a new option "proxy-auth" in parallel to "no-auth" or should there be something like "auth" that can be set to:
- no-auth
- proxy-auth
- ... maybe more to come?

> Be explicit about the risk in the codebase.
Make it default to be disabled.
Add a message/log when filebrowser starts in this mode.

Fine with those :)

> Consider adding an example configuration for a proxy such as NGINX or Traefik (or the one you are actually using).

[...]

PR #485, opened 2018-07-31
Proxy Auth support

maweck:
This should more or less solve #471  - As I am a Java guy there may be several things to improve so a review is more than welcome :) 

If this PR is fine I would create some documentation afterwards.

### DIFF
--- a/cmd/filebrowser/main.go
+++ b/cmd/filebrowser/main.go
@@ -2,15 +2,7 @@ package main
 
 import (
 	"fmt"
-	"io/ioutil"
-	"log"
-	"net"
-	"net/http"
-	"os"
-	"path/filepath"
-	"strings"
 	"github.com/asdine/storm"
-	"gopkg.in/natefinch/lumberjack.v2"
 	"github.com/filebrowser/filebrowser"
 	"github.com/filebrowser/filebrowser/bolt"
 	h "github.com/filebrowser/filebrowser/http"
@@ -18,6 +10,14 @@ import (
 	"github.com/hacdias/fileutils"
 	flag "github.com/spf13/pflag"
 	"github.com/spf13/viper"
+	"gopkg.in/natefinch/lumberjack.v2"
+	"io/ioutil"
+	"log"
+	"net"
+	"net/http"
+	"os"
+	"path/filepath"
+	"strings"
 )
 
 var (
@@ -35,17 +35,17 @@ var (
 	recaptchakey    string
 	recaptchasecret string
 	port            int
-	auth       	struct {
-	        method          string
-        	loginHeader     string
+	auth            struct {
+		method      string
+		loginHeader string
 	}
-	noAuth          bool
-	allowCommands   bool
-	allowEdit       bool
-	allowNew        bool
-	allowPublish    bool
-	showVer         bool
-	alterRecaptcha  bool
+	noAuth         bool
+	allowCommands  bool
+	allowEdit      bool
+	allowNew       bool
+	allowPublish   bool
+	showVer        bool
+	alterRecaptcha bool
 )
 
 func init() {
@@ -65,7 +65,7 @@ func init() {
 	flag.BoolVar(&allowEdit, "allow-edit", true, "Default allow edit option for new users")
 	flag.BoolVar(&allowPublish, "allow-publish", true, "Default allow publish option for new users")
 	flag.StringVar(&auth.method, "auth.method", "default", "Switch between 'default' and 'proxy' authentication.")
-        flag.StringVar(&auth.loginHeader, "auth.login-header", "X-Forwarded-User", "The header name used for proxy authentication.")
+	flag.StringVar(&auth.loginHeader, "auth.login-header", "X-Forwarded-User", "The header name used for proxy authentication.")
 	flag.BoolVar(&allowNew, "allow-new", true, "Default allow new option for new users")
 	flag.BoolVar(&noAuth, "no-auth", false, "Disables authentication")
 	flag.BoolVar(&alterRecaptcha, "alternative-recaptcha", false, "Use recaptcha.net for serving and handling, useful in China")
@@ -87,8 +87,8 @@ func setupViper() {
 	viper.SetDefault("AllowPublish", true)
 	viper.SetDefault("StaticGen", "")
 	viper.SetDefault("Locale", "")
-    viper.SetDefault("AuthMethod", "default")
-	viper.SetDefault("LoginHeader", "X-Fowarded-User");
+	viper.SetDefault("AuthMethod", "default")
+	viper.SetDefault("LoginHeader", "X-Fowarded-User")
 	viper.SetDefault("NoAuth", false)
 	viper.SetDefault("BaseURL", "")
 	viper.SetDefault("PrefixURL", "")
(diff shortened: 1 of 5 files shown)


maweck:
A quick test can e.g. be done by using a browser plugin like: https://chrome.google.com/webstore/detail/modify-headers-for-google/innpjfdalfhpcoinfnehdnbkglpmogdi

eine:
Hi @maweck! Can you please:

- Make a separate PR for the first commit (the one that modifies the Dockerfile).
  - Reordering the `COPY`, `WORKDIR` and `RUN dep` lines is OK, but note that you removed the 'ca-certificates' in `RUN apk`. Your should keep that line untouched.
- Run `gofmt -w ./` in the root of the repo, to have the style of all the go sources automatically fixed.

maweck:
Docker change has been reverted
gofmt has been executed.

http/auth.go
```diff
@@ -51,31 +51,33 @@ func reCaptcha(host, secret, response string) (bool, error) {
...
-	var cred cred
-	if r.Body == nil {
-		return http.StatusForbidden, nil
-	}
-
-	err := json.NewDecoder(r.Body).Decode(&cred)
-	if err != nil {
-		return http.StatusForbidden, nil
-	}
+	} else if c.AuthMethod == "proxy" {
```
eine: I think this is an unnecessary branch. The previous `if` contains a return statement, so there is no need for an `else` here. The flow will automatically execute `if c.AuthMethod == "proxy"` after evaluating `c.NoAuth` to false. Please, remove `else` and start in a new line.

http/auth.go
```diff
@@ -51,31 +51,33 @@ func reCaptcha(host, secret, response string) (bool, error) {
...
+		// Receive the Username from the Header.
+		cred.Username = r.Header.Get(c.LoginHeader)
```
eine: Can you merge the following lines here?

```
	// Checks if the user exists.
	u, err := c.Store.Users.GetByUsername(cred.Username, c.NewFS)
	if err != nil {
    }
```

I.e.:

```
	// Receive the Username from the Header and check if it exists.
	u, err := c.Store.Users.GetByUsername(r.Header.Get(c.LoginHeader), c.NewFS)
	if err != nil {
    }
```

This will allow not to evaluate `c.AuthMethod` again. See below.

Furthermore, if you put the return after the check, you can avoid the last `else`. See above.

```
c.User = u
return printToken(c, w)
```

http/auth.go
```diff
@@ -85,9 +87,11 @@ func authHandler(c *fb.Context, w http.ResponseWriter, r *http.Request) (int, er
 		return http.StatusForbidden, nil
 	}
 
-	// Checks if the password is correct.
-	if !fb.CheckPasswordHash(cred.Password, u.Password) {
-		return http.StatusForbidden, nil
+	if c.AuthMethod != "proxy" {
```
eine: If the check and return instructions are added inside the previous block (see comment above), this additional condition needs not to be checked.

eine:
Overall LGTM. Please, have a look at the comments and let me know what you think.

BTW, did you build and test it with your proxy setup? Since I cannot test it, it is important that you test and try to 'break' it.

@hacdias, can you please have a look and let us know if you find any caveat?

http/auth.go
```diff
@@ -51,31 +51,33 @@ func reCaptcha(host, secret, response string) (bool, error) {
...
 // authHandler processes the authentication for the user.
 func authHandler(c *fb.Context, w http.ResponseWriter, r *http.Request) (int, error) {
-	// NoAuth instances shouldn't call this method.
 	if c.NoAuth {
+		// NoAuth instances shouldn't call this method.
 		return 0, nil
-	}
-
-	// Receive the credentials from the request and unmarshal them.
-	var cred cred
```
eine: I think that you removed this line. So trying to use `cred.Username` and/or `cred.ReCaptcha` afterwards should fail.

cmd/filebrowser/main.go
```diff
@@ -2,25 +2,22 @@ package main
...
 	"github.com/filebrowser/filebrowser/bolt"
 	h "github.com/filebrowser/filebrowser/http"
 	"github.com/filebrowser/filebrowser/staticgen"
 	"github.com/hacdias/fileutils"
 	flag "github.com/spf13/pflag"
 	"github.com/spf13/viper"
+	"gopkg.in/natefinch/lumberjack.v2"
+	"io/ioutil"
+	"log"
+	"net"
```
hacdias: I wonder why all of these imports moved here. 

By the way, @1138-4EB just found out CircleCI had the option to build forked PRs disabled. I just enabled it, but I believe we'll need changes on this PR to trigger it.

hacdias (dismissed):
Seems good overall. Take a look at @1138-4EB's annotations please 😄 And **thanks**!

Could you rebase it on master?

cmd/filebrowser/main.go
eine: They are sorted in alphabetical order now. I think this is because `gofmt` was used. Should we use `goimports` instead? See http://goinbigdata.com/goimports-vs-gofmt/

maweck:
Hello together,

I hopefully implemented all the requested changes.

@1138-4EB : I have build a new image locally and did some tests. Both, proxy and default seem to work as expected. You can also do a quick test on your own as stated earlier. Simply start the with "--auth.method proxy" and request its page with a "X-Forwarded-User" Header set to an existing username (e.g. admin). 

https://chrome.google.com/webstore/detail/modify-headers-for-google/innpjfdalfhpcoinfnehdnbkglpmogdi

Best regards,

Marvin

cmd/filebrowser/main.go
```diff
@@ -63,6 +64,8 @@ func init() {
 	flag.BoolVar(&allowCommands, "allow-commands", true, "Default allow commands option for new users")
 	flag.BoolVar(&allowEdit, "allow-edit", true, "Default allow edit option for new users")
 	flag.BoolVar(&allowPublish, "allow-publish", true, "Default allow publish option for new users")
+	flag.StringVar(&auth.method, "auth.method", "default", "Switch between 'default' and 'proxy' authentication.")
+	flag.StringVar(&auth.loginHeader, "auth.login-header", "X-Forwarded-User", "The header name used for proxy authentication.")
```
hacdias: I don't know how I feel about the dot in the flag honestly. Probably replace it by `-` for consistency. @maweck why did you choose to use `.`?

maweck: Because that is how it is done in most of the tools I work with (Maven, SonarQube, Jenkins, ..... ). In addition it allows you to "group" specific configs. Here is a sample from SonarQube:

>sonar.web.sso.enable=false
>sonar.web.sso.loginHeader=X-Forwarded-Login
>sonar.web.sso.nameHeader=X-Forwarded-Name
>sonar.web.sso.emailHeader=X-Forwarded-Email
>...
>sonar.security.realm=LDAP
>ldap.url=ldap://localhost:10389
>ldap.bindDn=cn=sonar,ou=users,o=mycompany

But sure, I can change that if you want :)

hacdias: Looking at that tools they also seem to use camelCase. It would not be a bad idea. There are some of the other options which should be also grouped.

Could you add them with dots and camelCase please? I'll make another PR to update the other ones because it actually makes sense.

hacdias: Could you also rework it in a way to include noAuth on auth object? Maybe could be renamed auth.disable

hacdias: Ref.: #497

eine (approved):
The changes I suggested have been applied. It's ok on my side. Just the renaming suggested by @hacdias is pending.

maweck:
Hello @hacdias , hello @1138-4EB ,

so I renamed the var and prepared the "auth.method" config to accept "none" as a parameter, but I do not feel like I am the correct person to remove the "no-auth" option right now, because I am not that deep in your code and I am not aware of side effects at all :/ Maybe this can be done in a separate pull request from someone who is more into the filebrowser code than me?

hacdias:
Yes, I'll merge this and I can do that later. Thanks @maweck 😄

PR #490, opened 2018-08-04
chore: disable gosec

hacdias:
Yes, I like gosec and I find it useful, but it is complaining because we're using MD5... We need that.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -24,7 +24,8 @@ jobs:
               -D vetshadow \
               -D errcheck \
               -D golint \
-              -D gas
+              -D gas \
+              -D gosec
   build:
     docker:
       - image: circleci/golang:1.10


eine:
According to alecthomas/gometalinter#505 `gas` was renamed to `gosec`. So, we don't need `gas` anymore.

xor-gate:
I think you can suppress this specific case with /* #nosec */. It is very usefull to have gosec linter because it can catch some uncommon cases.

eine:
@xor-gate, thanks for pointing it out. Would you like to propose a PR?

I am concerned with passing arguments to gosec, because we do not run it directly, but through gomeralinter.

- https://github.com/filebrowser/filebrowser/blob/master/.gometalinter.json
- https://github.com/filebrowser/filebrowser/blob/master/build/run_gometalinter.sh

PR #494, opened 2018-08-05
fix: wildcard on commands

### LINKED ISSUE #466: Wildcards in command arguments are not handled/expanded properly
1. download from release page
2. no caddy

### 3. What are you trying to do?
gofmt after save

### 4. What did you expect to see?
success format the file

### 5. What did you see instead (give full error messages and/or log)?
internal server error

### 6. How can someone who is starting from scratch reproduce this behaviour as minimally as possible?

config global setting, add command ```gofmt -w *.go```, and try edit a go file and click save.

version: filebrowser version 1.9.0
os: CentOS Linux release 7.3.1611 (Core)

hacdias:
Fixes #466.

@1138-4EB @chinglin could you check if this works on your end?

### DIFF
--- a/lib/http/websockets.go
+++ b/lib/http/websockets.go
@@ -37,7 +37,7 @@ func init() {
 			path, err = exec.LookPath("sh")
 		}
 
-		if err != nil {
+		if err == nil {
 			commandShell = []string{path, "-c"}
 		}
 	}
@@ -105,7 +105,7 @@ func command(c *fb.Context, w http.ResponseWriter, r *http.Request) (int, error)
 	buff := new(bytes.Buffer)
 
 	if len(commandShell) != 0 {
-		command = append(commandShell, command...)
+		command = append(commandShell, strings.Join(command, " "))
 	}
 
 	// Sets up the command executation.
--- a/lib/http/websockets.go
+++ b/lib/http/websockets.go
@@ -9,6 +9,7 @@ import (
 	"os/exec"
 	"path/filepath"
 	"regexp"
+	"runtime"
 	"strings"
 	"time"
 
@@ -25,6 +26,22 @@ var (
 	cmdNotImplemented = []byte("Command not implemented.")
 	cmdNotAllowed     = []byte("Command not allowed.")
 )
+
+var commandShell []string
+
+func init() {
+	if runtime.GOOS != "windows" {
+		path, err := exec.LookPath("bash")
+
+		if err != nil {
+			path, err = exec.LookPath("sh")
+		}
+
+		if err != nil {
+			commandShell = []string{path, "-c"}
+		}
+	}
+}
 
 // command handles the requests for VCS related commands: git, svn and mercurial
 func command(c *fb.Context, w http.ResponseWriter, r *http.Request) (int, error) {
@@ -86,6 +103,10 @@ func command(c *fb.Context, w http.ResponseWriter, r *http.Request) (int, error)
 	path := c.User.Scope + "/" + r.URL.Path
 	path = filepath.Clean(path)
 	buff := new(bytes.Buffer)
+
+	if len(commandShell) == 0 {
+		command = append(commandShell, command...)
+	}
 
 	// Sets up the command executation.
 	cmd := exec.Command(command[0], command[1:]...)
--- a/lib/http/websockets.go
+++ b/lib/http/websockets.go
@@ -104,7 +104,7 @@ func command(c *fb.Context, w http.ResponseWriter, r *http.Request) (int, error)
 	path = filepath.Clean(path)
 	buff := new(bytes.Buffer)
 
-	if len(commandShell) == 0 {
+	if len(commandShell) != 0 {
 		command = append(commandShell, command...)
 	}
 


hacdias:
@1138-4EB could you take a look at this?

http/websockets.go
```diff
@@ -26,6 +27,22 @@ var (
 	cmdNotAllowed     = []byte("Command not allowed.")
 )
 
+var commandShell []string
+
+func init() {
+	if runtime.GOOS != "windows" {
```
eine: Should we really add this condition? Filebrowser can be executed inside MSYS/MinGW, where `bash` is available.

http/websockets.go
```diff
@@ -26,6 +27,22 @@ var (
...
+		path, err := exec.LookPath("bash")
+
+		if err != nil {
+			path, err = exec.LookPath("sh")
+		}
+
+		if err == nil {
```
eine: I fixed this condition.

http/websockets.go
```diff
@@ -87,6 +104,10 @@ func command(c *fb.Context, w http.ResponseWriter, r *http.Request) (int, error)
 	path = filepath.Clean(path)
 	buff := new(bytes.Buffer)
 
+	if len(commandShell) != 0 {
+		command = append(commandShell, strings.Join(command, " "))
```
eine: If `bash -c` or `sh -c` are used, the original command must be passed as a single argument. I.e., command `ls -la ./*` should be executed as `bash -c "ls -la ./*"`. Else, it is executed as `bash -c ls`. That's why I added `strings.Join`.

eine (approved):
I fixed two issues and I tested it in Fedora 28. On top of the comment about Windows, some thoughts:

- Should we default to executing the command without commanShell and use `bash`|`sh` as a fallback?
  - Should we provide two different keys in the frontend (`$` and `#`) to make the options explicit?
- This fix will not work in the default filebrowser docker image, because it is based on `scratch`, so no `bash`|`sh` is available. Should we consider using `Glob` as a fallback in this context?
  - If we implemented two keys, as proposed above, one of them could be disabled if no `bash`|`sh` is found in the 'host'.

Anyway, we can discuss these after the PR is merged, and work on them in further PRs.

chinglin:
Hi Henrique,FYI you're sending this to the wrong Ching Lin, you must have picked the wrong GitHub user.                                                                                                                                                                                                                                                                                                                                        Sent from my BlackBerry 10 smartphone.

http/websockets.go
hacdias: Yeah, we could just use if it's available.

hacdias:
Oops, sorry @chinglin! It was to @chinglinwen 😄

chinglinwen:
Sorry for the late, I build with this branch

No error box shows, with setting ( after save command )

Since I can't edit  go suffix file, I can't test it

I try the searchbox to execute "gofmt -w main.go", it seems doesn't work, only show a file found, content doens't change.

@hacdias

eine:
@chinglinwen thanks for testing. This PR should not affect the execution of commands without wildcards. Can you, please, try `$gofmt -w main.go` and `$gofmt -w ./*` in the 'search' box?

chinglinwen:
It shows ```This is a not supported command.``` for both commands provided above. @1138-4EB

chinglinwen:
Also my primary concern is automatic format on save. not execute command to format on search box manually. ( hope this helps to understand the issue #466  )

@1138-4EB @hacdias

eine:
> It shows `This is a not supported command`. for both commands provided above.

This means that you didn't add `gofmt` to the list of supported commands. You need to do it either in the config file or in the user settings. Please, do it and try again.

> Also my primary concern is automatic format on save. not execute command to format on search box manually.

If/when it works in the search box, it should also work in the hook (automatic execution on save).

hacdias:
@chinglinwen @1138-4EB now I wonder if we should merge this because:

1. The behaviour inside a shell might be different.
2. The user could just allow to run `bash` or `sh`.

Or:

We could add an option to run the commands inside the shell they want. Eg: to run inside bash, the user would give us the template `/usr/bin/bash -c "{{ .Command }}"`

eine:
I think we could have three radio buttons next to the search box in the new version. Those three buttons would be 'Search', 'Execute' and 'Shell'. Then, the user can easily choose the behaviour. Furthermore, the template for option 'Shell' could be configurable in the settings. Equivalently, each hook command would have a select box to choose between 'Execute' and 'Shell'.

hacdias:
@1138-4EB yeah, that's a great idea. I'll think about the best way to do this 😄 Liked the option on the interface to pick

hacdias:
Well, I was thinking, and the shell option in the interface would probably give more confusion. I think in the interface we should leave Search and Run/Execute.

There would be a global setting telling if it should be run inside a shell or just executed. I doubt that anyone would like to run commands specifically without being in a shell..

WHat do you think?

hacdias:
@1138-4EB actually I think we should **separate** search and execution. I have an idea. Will show you when it's better done. That way they can be improved on their own and separately and not depending on each other.

hacdias:
image

hacdias:
Closing this in favour of #575 😄

eine:
@hacdias, after the screenshot you posted, would it be feasible to integrate xtermjs or some alternative?

See e.g. [freman/goterm](https://github.com/freman/goterm).

hacdias:
@1138-4EB woah. That's great. Will take a look tomorrow, but yes 😄

hacdias:
@1138-4EB although it seemed a good idea at start, since our use case is *really* simple it seems it will need too much configuration for very little reason. Since we won't have it directly connected to a terminal, we would need to parse keys (backspace, arrows, line feeds, carriage returns... etc) ourselves, which would be a lot *lot* of code. What do you think?

Look at this echo example https://github.com/wavesoft/local-echo

eine:
I think that the most straighforward solution is to enable it only if a shell is available on the host (let it be a container, or a native host). I.e., a runtime check as the one in this PR.

hacdias:
Hmm... but then we need to have stylings for the default shell or for the other commands.

eine:
Styling with xtermsjs should not be needed, it is bind to a div. What we should handle is the styling of the other commands, which we need to do anyway, don't we?

hacdias:
The thing is: if we're not connecting directly to the terminal, I mean, if we're not running a shell proccess that's running behind (we're just doing shell -c yourCommand), what's the advantage of xterm?

eine:
Mostly format. Even if we are just running a command, not an interactive shell, xtermjs will format the output. Plus, when we really want an interactive experience, the interface is already there. E.g. Visual Studio Code's Terminal and Output tabs (which are based on xtermsjs BTW).

hacdias:
@1138-4EB hmmm... What if we leave that for 2.1 or soemthing? It seems to actually require some work. You could even open an issue

PR #492, opened 2018-08-05
chore: new build process

hacdias:
**Please read #399 before any of the following conversations.**

PR #496, opened 2018-08-06
add go.mod for go1.11+

Equim-chan:
Please note that this also updates bbolt to v1.3.1-coreos.6, instead of v1.3.0 which was released two years ago (I think it's a bug in semver).

After Go1.11 release maybe we can drop Gopkg.{toml,lock}?


hacdias:
Uh, this seems interesting. Since @1138-4EB is taking care of #492, I believe it is better to be him talking to you about this 😄

eine:
It looks interesting, but I have to read about it. Anyway, Go 1.11 is in beta yet, so we have some margin until we use it. @Equim-chan, should you have any handy reference about this `go.mod` approach, I'll like to have a look at it. I.e., are these produced by [golang/dep](https://github.com/golang/dep) or is there a new tool?

Equim-chan:
It's a new tool and already embedded into go1.11 toolchain. It can parse Gopkg.lock (and it's how this commit comes from). The biggest benefit, in my opinion, is that you don't need to put it under GOPATH any more. I've seen there are too many newcomers stumbled on GOPATH when they try to build something in Go.

eine:
According to https://github.com/golang/go/wiki/Modules#semantic-import-versioning, we might need to take some further action to support `go.mod` considering the upcoming breaking changes (see #497, #506).

Also I did need to update `Gopkg.lock` in a different branch (see #505).

@Equim-chan could you please have a look at these?

eine:
Now #505 and #506 (including #497) are merged.

On the one hand, as commented in https://github.com/filebrowser/filebrowser/pull/506#issuecomment-414395143:

> I had problems with `dep ensure -vendor-only`. Then, I removed `dep`, `Gopkg*` and `vendor`, in order to use `go mod`. However, building with `go mod` is failing too. So, master does not use any vendoring/module tool right now. `go get -v ./...` is used.

On the other hand, should we create a new branch `v2` and move all the commits after `v1.10.0` there?

---

> @Equim-chan:
> The biggest benefit, in my opinion, is that you don't need to put it under GOPATH any more.

I tried this, but it seems that we need to add import comments to most of the sources for it to work.

EDIT: we are using go1.11 now (#530), but new modules are not being used, because the project is mounted in he GOPATH inside the docker container, and `GO111MODULE=on` is unset.

PR #497, opened 2018-08-06
refactor: rename settings (breaking change)

hacdias:
Related to https://github.com/filebrowser/filebrowser/pull/485#discussion_r207921156, filebrowser/filebrowser.github.io#9 and filebrowser/caddy#4

License: MIT
Signed-off-by: Henrique Dias <hacdias@gmail.com>

### DIFF
--- a/cmd/filebrowser/main.go
+++ b/cmd/filebrowser/main.go
@@ -2,6 +2,14 @@ package main
 
 import (
 	"fmt"
+	"io/ioutil"
+	"log"
+	"net"
+	"net/http"
+	"os"
+	"path/filepath"
+	"strings"
+
 	"github.com/asdine/storm"
 	"github.com/filebrowser/filebrowser"
 	"github.com/filebrowser/filebrowser/bolt"
@@ -11,13 +19,7 @@ import (
 	flag "github.com/spf13/pflag"
 	"github.com/spf13/viper"
 	"gopkg.in/natefinch/lumberjack.v2"
-	"io/ioutil"
-	"log"
-	"net"
-	"net/http"
-	"os"
-	"path/filepath"
-	"strings")
+)
 
 var (
 	addr            string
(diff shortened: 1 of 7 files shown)


eine:
This PR modifies `caddy/parser/parser.go`, which is removed in #492.

- If #492 is merged first, these changes need to be done in `filebrowser/caddy` directly.
- If this PR is merged first, changes need to be done in `filebrowser/caddy` as soon (or before) #492 is merged.

hacdias:
Wait for #492

eine:
A thought before actually reviewing it: since this is a breaking change, and the filebrowser docker image is automatically updated now, should we publish a release (`v.1.10.0) before merging this? That would provide users, who cannot update their configs ATM, a stable image to use.

NOTE: when we publish a release now, a single docker image is created and it is tagged twice (`latest` and `$version`). So, if we publish a release now and a breaking one in a few days, we will have one tagged `v.1.10.0` and another one tagged `v2.0.0` and `latest`.

Furthermore, I don't know if the changes introduced in #485 should be considered breaking changes too.

eine (approved):

hacdias:
Yes @1138-4EB, please do the honors 😄

Equim-chan:
Might need a db mitigation, since the structure has been changed?

hacdias:
@Equim-chan no, the DB wasn't changed at all 😄 just the CLI options and the config file

Equim-chan (approved):

cmd/filebrowser/main.go
```diff
@@ -53,70 +56,82 @@ func init() {
...
+	flag.StringVar(&locale, "defaults.locale", "", "Default locale for new users, set it empty to enable auto detect from browser")
+
+	// Recaptcha settings
+	flag.StringVar(&recaptcha.host, "recaptcha.host", "https://www.google.com", "Use another host for ReCAPTCHA. recaptcha.net might be useful in China")
+	flag.StringVar(&recaptcha.key, "recaptcha.key", "", "ReCaptcha site key")
+	flag.StringVar(&recaptcha.secret, "recaptcha.secret", "", "ReCaptcha secret")
+
+	// Auth settings
+	flag.StringVar(&auth.method, "auth.method", "default", "Switch between 'none', 'default' and 'proxy' authentication")
+	flag.StringVar(&auth.header, "auth.header", "X-Forwarded-User", "The header name used for proxy authentication")
```
Equim-chan, line 81: Can we make some deprecated flags? For example when user tries to use the old `--allow-new` flag, we give a hint that `--allow-new is deprecated, use --defaults.allowNew instead`.

hacdias: I think we could, but we'd need to do it for every flag. If you're at it, you can update this PR itself 😄

eine:
@hacdias, can you please have a look at the comment about [BindPFlags](https://godoc.org/github.com/spf13/viper#BindPFlags) in #506?

Also, I don't know if you prefer to merge this PR on its own, or as part of #506, since it is almost ready too.

hacdias:
@1138-4EB yes, it's ok for me to close this in favor of #506 😄 and using BindPFlags as long as it works

cmd/filebrowser/main.go
eine: I am at it. Will push soon.

https://godoc.org/github.com/spf13/pflag#FlagSet.MarkDeprecated

eine: Done. See https://github.com/filebrowser/filebrowser/pull/506/commits/1314e432be7235af20c35b5d815b7b4a73cda059.

Issue #489, opened 2018-08-04
Configuration and database rework


eine:
This issue summarizes the requested features for the workflow related to the configuration file, the database, how user profiles are set up, etc. Some of these are frequently shown in multiple issues and this is an effort to put them all together. It is also the place where we can add and track new related requests. 

There was some previous discussion about this in #371. However, rather than merging the configuration file and the database, I think that the configuration file should be extended ([#381issuecomment-373983968](https://github.com/filebrowser/filebrowser/issues/381#issuecomment-373983968)) to be used as a template to initialize/set the database (probably periodically).

## Config file

- [x] Define multiple users in the config file [#371issuecomment-373985636](https://github.com/filebrowser/filebrowser/issues/371#issuecomment-373985636). See go-yaml/yaml#394.
  - Use (additional) JSON file to configure users [#424issuecomment-392551151](https://github.com/filebrowser/filebrowser/issues/424#issuecomment-392551151)
- [x] allow to set commands for events in config file [#246issuecomment-349563766](https://github.com/filebrowser/filebrowser/issues/246#issuecomment-349563766)
- [x] Add option "default sort mode" [#326issue-287085355](https://github.com/filebrowser/filebrowser/issues/326#issue-287085355) (for each user)
- [x] Rules disallow/allow in config file or globally #424 (for each user)
- [x] More flexible rights/permission scheme #502
  - [x] Add user group features to set some permissions [#401issue-316476440](https://github.com/filebrowser/filebrowser/issues/401#issue-316476440). *We can try using YAML variables to define groups. But I don't know if these groups are readable*.
		- We support YAML. Please see #614 for actual user groups.
- [x] Customizing admin credentials via config #257. *Let `admin` be a group and not a user?*
- [ ] ~Support for password changing via config #381 (for each user)~ via CLI
- [ ] ~Allow disabling user submitted stylesheets #507 (per user)~ no per-user CSS anymore.

## CLI flags/options

- [x] Customizing admin credentials via flags #257, #381, #435.
- [ ] ~Add option to enabe/disable [watching-and-re-reading-config-files](https://github.com/spf13/viper#watching-and-re-reading-config-files).~
- [x] Subcommand/option to [#401issue-316476440](https://github.com/filebrowser/filebrowser/issues/401#issue-316476440):
- [x] Add users from the config file to the database without removing existing ones. If conflict:
  - Overwrite.
  - Show both in the terminal and ask which one to keep.
- [x] Remove one/some/all users from the database.
[...]

Issue #360, opened 2018-02-22
Configuration using environment variables


bcardiff:
For easy deployment of the docker image it would be great to allow the config.json file to be generated from environment variables upon start. 

Another alternative, that the code itself checks if certain environment variables are defined and then fallback to the config.json (that is a pattern widely used AFAIK). Depending on all the configuration options this might be even a better approach than expanding a template to a .json.

Thanks for the great project!

eine:
@bcardiff, @veteran29, @davidpelayo, @derekoharrow, this is now merged in master, along with other enhancements (#506). It will be available in the next release.

You can see implementation details in #504 -> #506. See also [spf13/viper#working-with-environment-variables](https://github.com/spf13/viper#working-with-environment-variables) and [precedence](https://github.com/spf13/viper#why-viper).

bcardiff:
Probably it's on its way, but is hard to know the allowed config options. Between  #504 and #506 I am a bit lost what is the current rule to map `"some.config"` to an env variable. Is it `FB_SOME_CONFIG`/`FB_SOME__CONFIG`/`FB_SOME.CONFIG`/`FB_some.config`? I am not used to the codebase and packages used. From a user perspective knowing the convention will be really helpful.

Thanks for resolving the issue and keep developing FB. I will eagerly wait for the next release.

eine:
The relevant lines, which were introduced in #506, are [cli/cmd/root.go#L65-L67](https://github.com/filebrowser/filebrowser/blob/master/cli/cmd/root.go#L65-L67). Note that it is the same as [#504/files](https://github.com/filebrowser/filebrowser/pull/504/files), with an additional replacement. Therefore the format is:

- Prefix `FB_`.
- Convert key/flag to uppercase.
- Replace `.` with `_`.

The first of the options you commented is correct: `FB_SOME_CONFIG`.

Anyway, note that flags take precedence over environment variables. Values in the config file do not. E.g.:

``` bash
$ docker run --rm -it -p 5566:1234 -e FB_PORT=1234 -e FB_AUTH_METHOD=none filebrowser/filebrowser
2018/08/23 00:34:16 Using config file: /.filebrowser.json
2018/08/23 00:34:16 Listening on [::]:1234
```

``` bash
$ docker run --rm -it -p 5566:1234 -e FB_PORT=8888 -e FB_AUTH_METHOD=none filebrowser/filebrowser -p 1234
2018/08/23 00:35:24 Using config file: /.filebrowser.json
[...]

PR #504, opened 2018-08-08
Allow to configure filebrowser through environment variables (#360) (#489)

maweck:
I also had the need to provide the settings as Environment variables. A quick research pointed out that Viper already supports the feature requested in #360 . So I did some tests and it works fine.

Not sure if you are fine with "FB" as a prefix or if you want to see this PR at all, because you may want to solve everything mentioned in #489 in one PR. However, at least I can let you guys know that it would work like this :)

Best regards

### DIFF
--- a/cmd/filebrowser/main.go
+++ b/cmd/filebrowser/main.go
@@ -120,6 +120,9 @@ func setupViper() {
 
 	viper.SetConfigName("filebrowser")
 	viper.AddConfigPath(".")
+
+	viper.SetEnvPrefix("FB")
+	viper.AutomaticEnv()
 }
 
 func printVersion() {


eine:
@maweck thanks a lot!

Related to other enhancements in progress, flags/parameters with dots (#497) cannot be set through envvars. A replacement command is required. So, I already merged this commit into #506, and added the replacement there.

eine:
@maweck this is now merged in master, along with other enhancements (#506). Please, let us know if you find any problem with it. Thanks again!

maweck:
Looks very good! Thank you for the effort @1138-4EB - Do you have some kind of "pre-release" planned that can be tested?

eine:
We'd like to have some PRs merged before publishing a new release (i.e. #494, #496, #515). However, there are two options for you to test it:

- Execute `USE_DOCKER="true" ./build/build_all.sh` in the root of the repo. A `filebrowser` binary will be generated.
- Use the latest docker image: `filebrowser/filebrowser:latest`.

In both cases you only need docker installed locally. So, alternatively, you can do it in play-with-docker.

See https://github.com/filebrowser/community/blob/master/development.md.

PR #505, opened 2018-08-09
chore: use commit SHA as version if not release

eine:
This PR replaces the `(untracked)` version with the first eight characters of the commit that is being built when `USE_DOCKER="true" ./build/build_all.sh` is used. If executed outside of a docker container envvar `COMMIT_SHA` can be set to any value and that will be used.

Since `build.sh` is not executed when commits are tagged, this does not apply to releses.

Also, `dep ensure --update` was executed in order to update `Gopkg.lock`.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -3,29 +3,25 @@
 
 [[projects]]
   branch = "master"
+  digest = "1:af3215cd158218c92f0706bb32f67b0ce3d37da4c4d135e7ed5e2dc36a90fca5"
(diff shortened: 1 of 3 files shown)


eine:
I don't know why, this and some other PRs (#497, #504) are not being executed in travis.

hacdias:
@1138-4EB on travisstatus.com it says «Partially Degraded Service». It might be because of that.

hacdias (approved):

PR #508, opened 2018-08-09
Ensure that the config path provided through flags is used

### LINKED ISSUE #461: Dockerhub docker image -> command line parameters don't work
### What are you trying to do?
Install filebrowser via docker. Give a path to the config file via command line. Basically my command to run the container looks like this:

`docker run -v /home/mozzbozz/data:/srv -v /home/mozzbozz/config:/config -p 8080:80 hacdias/filebrowser --config /config/config.json`

My `/home/mozzbozz/config/config.json` looks like the following:

```yaml
{
  "port": 80,
  "address": "",
  "database": "/config/database.db",
  "scope": "/srv",
  "allowCommands": true,
  "allowEdit": true,
  "allowNew": true,
  "commands": []
}
```

### What did you expect to see? What did you see instead?
A file with the name `/home/mozzbozz/config/database.db` should be created on my host system if it doesn't exist.

Well, that's not the case. Therefore stopping and running the container again, changes like new users, ... are not persistent. It doesn't matter if I pull the latest image or the v1.8.0 tag (also tried some older images).

### So what's your problem?
If I clone this git repository and run a simple `docker build -t test .` inside the root directory and issue the following command (basically the same as shown above, just the image name is replaced):

`docker run -v /home/mozzbozz/data:/srv -v /home/mozzbozz/config:/config -p 8080:80 test --config /config/config.json`

...everything works as expected. Restarting the container doesn't result in a loose of the changes and the `database.db` file is created where expected.

I have honestly no idea, why this happens or how this can be fixed. I've even compared the Dockerfile shown on Dockerhub with the one inside this git repo and couldn't find any differences. Tested this on two different computers with different Linux distributions and different docker versions...

eine:
Close #461.

The root path (`.`) was automatically added to viper. As a result, when using the Docker image, if the user provided a different path for the config file, but with the same filename, the default file in the image (`/config.json`) was used instead of the one provided by the user.

This PR ensures that no default search path is added when the user provides the config through a flag.

Also, the viper initialization is moved to a function.

### DIFF
--- a/cmd/filebrowser/main.go
+++ b/cmd/filebrowser/main.go
@@ -117,9 +117,6 @@ func setupViper() {
 	viper.BindPFlag("AlternativeRecaptcha", flag.Lookup("alternative-recaptcha"))
 	viper.BindPFlag("ReCaptchaKey", flag.Lookup("recaptcha-key"))
 	viper.BindPFlag("ReCaptchaSecret", flag.Lookup("recaptcha-secret"))
-
-	viper.SetConfigName("filebrowser")
-	viper.AddConfigPath(".")
 }
 
 func printVersion() {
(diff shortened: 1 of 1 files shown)


hacdias (approved):
It LGTM!

Issue #415, opened 2018-05-08
Case Insensitive & Fuzzy Search 


collegeimprovements:
Would it be possible to make search `case:insensitive` by default or via some config option ?
And would it be possible to have fuzzy search ?

bkanuka:
I opened issue #336 which, if implemented, would allow this and other search customization. Unfortunately I gave it my best shot and couldn't get it.

DeyV:
I suggest  change default behavior. Everyone use google and other search tools, and most of them is case insensitive. It is especially important in file search, when nobody think about letters size (excluding programmers)

1138-4EB:
Since case insensitive search is implemented already, I suggest we take one of these paths:

- Change the title of this issue to "Fuzzy Search" and use it to track that.
- Close this issue and open a new one for Fuzzy Search.
- Close this issue and add a comment about Fuzzy Search in #336.

@collegeimprovements, it is up to you.

Issue #252, opened 2017-10-13
Search improvement ideas


francislavoie:
It seems like it only searches filenames. It would be nice if it would also match against folders/paths. I use it as a dead-easy "fake plex", and a lot of TV show filenames are just random hashes, but the folder names are correct.

Also, the search acts weird based on context. Say I'm in the settings page and I try to use the search, there'll never be any file results, because the page I tried to search from isn't an actual file list. The search view should probably show the current path context it's searching in.

I think the results might look better if they were styled similarly to the files list view, with the modification time and such included.

Maybe if possible, sorting the search results? By relevance by default, but user could pick by name/time?

eine:
We'll use this issue to track ideas, suggestions and enhancements related to the 'Search' feature:

- #336. External search commands.
- #415. Fuzzy search.

helloray:
Please also consider to add an auto complete for search. Just like google.

hacdias:
I'm closing this since this repository will be archived. Please read the information on the README for more information. Thanks for all your contributions!

PR #506, opened 2018-08-09
feat: use cobra to provide subcommands, move sources to lib

eine:
Close #360, close #497 and close #504.

---

This is a proof-of-concept to use cobra in order to provide subcommands. Two main subcommands are provided:

- `filebrowser serve` or `filebrowser server` keeps the existing behaviour. The root command, `filebrowser`, is aliased to this.
- `filebrowser db` is expected to be used for interacting with de DB. See #489.

Also:

- Sources related to the core are moved to subdir `lib`.
- Subdir `cmd` is removed.
- Subdir `cli` is created, which is a standard cobra structure.
- #497 and #504 are merged.
- Deprecated flags are added. See https://github.com/filebrowser/filebrowser/pull/497#discussion_r209428120.
- [`viper.BindPFlags`](https://godoc.org/github.com/spf13/viper#BindPFlags) is used in order to reduce the verbosity in `serve.go`.

---

- [x] Add proper short and long descriptions to the subcommands.
- [x] Make `filebrowser` behave as `filebrowser serve`, accepting the same flags. (spf13/cobra#725)
- [x] Add subcommand `version` that behaves as `--version`. (spf13/cobra#724)
- [x] Modify the build scripts accordingly. I.e., rice-box is moved to subdir `lib` too.
- [x] Modify caddy plugins accordingly. (filebrowser/caddy#11)
- [x] Modify `initConfig`? See #508.

PR #515, opened 2018-08-21
fix search not working

### LINKED ISSUE #484: Searched filename in UPPER CASE is found but result is lower case, thus file not found :(
**Description**
I searched this File3.txt and is able to find but presented file is file3.txt

**Expected behaviour**
Should be able to find (yes this is achieved) however the presented file (in URL) is in lowercase. Upon clicking I get this "This location can't be reached".

hugomassing:
Closes #484 

Fix search not working because the path was in lower case and resulted in a 404 when clinking on the search result. 

This is a quick fix while a better solution is found for the fuzzy search / case insensitive search problems described in #252 and #415.

### DIFF
--- a/lib/http/websockets.go
+++ b/lib/http/websockets.go
@@ -279,7 +279,7 @@ func search(c *fb.Context, w http.ResponseWriter, r *http.Request) (int, error) 
 		var (
 			originalPath string
 		)
-		
+
 		path = strings.TrimPrefix(path, scope)
 		path = strings.TrimPrefix(path, "/")
 		path = strings.Replace(path, "\\", "/", -1)
--- a/lib/http/websockets.go
+++ b/lib/http/websockets.go
@@ -276,13 +276,19 @@ func search(c *fb.Context, w http.ResponseWriter, r *http.Request) (int, error) 
 	scope = filepath.Clean(scope)
 
 	err = filepath.Walk(scope, func(path string, f os.FileInfo, err error) error {
-		if !search.CaseSensitive {
-			path = strings.ToLower(path)
-		}
-
+		var (
+			originalPath string
+		)
+		
 		path = strings.TrimPrefix(path, scope)
 		path = strings.TrimPrefix(path, "/")
 		path = strings.Replace(path, "\\", "/", -1)
+
+		originalPath = path
+
+		if !search.CaseSensitive {
+			path = strings.ToLower(path)
+		}
 
 		// Only execute if there are conditions to meet.
 		if len(search.Conditions) > 0 {
@@ -326,7 +332,7 @@ func search(c *fb.Context, w http.ResponseWriter, r *http.Request) (int, error) 
 
 		response, _ := json.Marshal(map[string]interface{}{
 			"dir":  f.IsDir(),
-			"path": path,
+			"path": originalPath,
 		})
 
 		return conn.WriteMessage(websocket.TextMessage, response)


eine:
@hugomassing, can you please rebase this PR on top of master? Note that `http/websockets.go` is now located at `lib/http/websockets.go`.

hacdias (approved):

hacdias:
Thanks @hugomassing. Could you just rebase, as @1138-4EB said?

hugomassing:
Looks good now :)

eine:
Thanks @hugomassing! :wave:

Just a note for future contributions: I ran `gofmt lib/http/websockets.go` on top of your commit in order to remove some whitespaces.

ChrisGet:
I have just installed this great piece of software using the install script method:
curl -fsSL https://filebrowser.github.io/get.sh | bash

Version is 1.10 but I still have the search issue with results being all lower case so the links are not found. How can I include the fix as stated above?

eine:
Hi @ChrisGet, v1.10.0 was released on 2018-08-16, and this fix was added on 2018-08-22. So it will be available in the next release.

If you want to have it now, you need to build filebrowser from sources. See https://github.com/filebrowser/community, [build](https://github.com/filebrowser/community/blob/master/builds.md) and [development](https://github.com/filebrowser/community/blob/master/development.md).

Alternatively, you can use docker image `filebrowser/filebrowser:latest`, which is always built with the latest commit in branch master. So, the fix should be already applied there.

ChrisGet:
@1138-4EB Thanks for your response. Do you have any idea when the next release will be published?

I have tried using the install files build_all.sh but I am hitting problem after problem with installing go, rice, and everything else.

I have managed to install docker. I am on Debian 8.10 Jessie. How do I use that docker image that you referred to? I have no experience with docker I'm afraid.

Thanks.

ChrisGet:
@1138-4EB Sorry I found the documentation. I'll try the docker method. Thanks.

eine:
> Sorry I found the documentation. I'll try the docker method. Thanks.

Let us know if you have any problem with it. The image should be automatically downloaded, so you just need to run a container.

> Do you have any idea when the next release will be published?

That is actually hard to reply at now. If you have a look at https://github.com/filebrowser/filebrowser/commits/master, you will see that, indeed, we released `v1.10.0` a month ago because we were going to introduce some breaking changes (#506). Therefore, the next release should be `v2.0.0`, according to https://semver.org/. However, golang just released version 1.11 (#530), which introduces some changes about how packages/dependencies are handled: https://github.com/golang/go/wiki/Modules#semantic-import-versioning:

> (...) In contrast, a package that has opted in to modules must include the major version in the import path for any imported v2+ modules)
> There are two ways to release a v2 or higher module version. Using the example of creating a v2.0.0 release, the two options are:
> - Update the go.mod file to include a /v2 at the end of the module path. Tag the release with v2.0.0.
>    - To avoid confusion with this approach, consider putting the v2.*.* commits on a separate v2 branch.
>    - If instead you have been previously releasing on master and would prefer to tag v2.0.0 on master, that is a viable option, but consider creating a v1 branch for any future v1 bug fixes.
> - Alternatively, create a new v2 subdirectory (e.g., my/module/v2) and place a new go.mod file in that subdirectory. The module path must end with /v2. Copy or move the code into the v2 subdirectory. Tag the release with v2.0.0.
>    - This provides greater backwards compatibility. In particular, Go versions older than 1.9.7 and 1.10.3 are also able to properly consume and build a v2 or higher module created using this approach.

This is yet to be solved in #496.

At the same time #532, so it is difficult to know when will other breaking changes be done.

Meanwhile, we might cherry-pick some commits after `v1.10.0` (#508, #514, this and #531) and either:

- Apply them on top of tag `v1.10.0` as hotfixes (`v1.10.1`).
- Rebase/reorder master branch and tag/publish a new release (`v1.10.1`) just before the breaking changes.

The first option is not desirable, because we would have a bunch of duplicated commits, really close to each other. But, the second one requires to rebase the master branch.

@hacdias , @Equim-chan , any ideas?

hacdias:
@1138-4EB I have as a rule of thumb not to force push to master and I believe it's a best practice so we shouldn't rebase it.

How would we do the first one?

eine:
> How would we do the first one?

We can either [cherry-pick](https://git-scm.com/docs/git-cherry-pick) or:

``` bash
git checkout -b tmp origin/master
git rebase -i v1.10.0
# Remove the commits we don't want to keep in `v1.10.1`.
git tag v1.10.1
git checkout master
git branch -D tmp
git push origin v.1.10.1
```

PR #531, opened 2018-09-08
fix small text file been recognize as blob issue

chinglinwen:
**Description**

relate to #314 

I've realize it's some go file can't edit, and some file can.
During read the code, I think the problem is at `GetFileType` method.

After this change, the go file(even it's small) can be correctly recognize, and I have tested it.

Test by using two go file, one is small then 512 bytes, the other is over 512 bytes, this change make both file editable through UI.


eine:
Thanks @chinglinwen :+1:

PR #533, opened 2018-09-08
docs: remove old disclaimer and add new one

hacdias:
Remove old disclaimer and add new one.


eine (approved):
I think that it is useful to keep the comment about renaming the project and docker images.

Also, IMO 'warning' may be 'too much' as it could refrain people from using it. I find 'INFO' to be more adequate.

hacdias:
Thanks @1138-4EB 😄

Issue #381, opened 2018-03-12
Support for password changing via command line or config


maran:
I'm trying to add support for Filebrowser to my one-click-install hosting platform.

Ideally I would love some way to change the password before the first login so that I can install and configure the application for my end-user without having to give everybody the admin/admin details. (they might be lazy and not change it leading to security issues).

Would you be open to accepting a PR that implements some form of setting the initial password via a config or command line option? Is there a better way you can think of to do this?

Equim-chan:
For a further goal, I think a full featured HTTP API can be implemented to help.

hacdias:
Hello @maran and @Equim-chan,

Here are my two cents: I think we have too much command line flags. When I say this, I mean: we don't have all of the possible features withing command line flags and we don't have all the possible configuration in a configuration file.

I think we should invest on creating a better configuration file. See #371.

ghost:
I also vote for this issue. The current solution with a default admin/admin user makes automated deployments impossible or unsafe. There is always a manual step required.

I am deploying with Docker in Kubernetes and it's common to specify such information in a secret, which is then used to populate environment variables for the container.

bcardiff:
Now that #360 is in master I can vote for this also :-). I was also interested in a one-click-install, in my case for rancher, but without admin password config option is not a good fit for public usage.

eine:
Indeed, with #506 we removed possible inconsistencies between the config file, flags and envvars. Also a placeholder for database related CLI (sub)commands was added.

Therefore, it is relatively straightforward to add default `user` and `password` fields to the config file, flags and envvars at the same time. The user is not a problem, but as commented by @hacdias in [371-issuecomment-370157182](https://github.com/filebrowser/filebrowser/issues/371#issuecomment-370157182) it is not desirable at all to store passwords in plain text.

---

We might add several subcommands to `filebrowser db`:

- `filebrowser db init` would initialize a db file (i.e. execute func `Setup`) without starting the server (i.e. do not execute func `http.Serve`).
- `filebrowser db set pass <user>` could prompt for a password (twice) and update the database. 
- `filebrowser db hash` would take a string and execute `HashPassword`. Then, the user would use the hash in the config/flag/envvar when executing `filebrowser serve`. However, I don't know if this is secure enough.

---

Moreover, a very demanded feature is to allow programmatically adding/removing/editing user entries in the database. A possible approach is to use a YAML file to define users and groups. E.g.:

``` yaml
- &g-admin
[...]

Issue #257, opened 2017-10-18
Customizing admin credentials via config/flags


jzaefferer:
I'm interested in using this project in a managed hosting context, where customers could use filemanager to access files. For that the default admin/admin credentials are a problem. Is it reasonable to add config/flags to set the default values for those?

It looks like it  could be a simple code change (plus a bit more effort for tests + documentation): https://github.com/hacdias/filemanager/blob/26e1e44d602b9bca5b1e20771095579f4a1c2b43/filemanager.go#L193-L196 - instead of using the string "admin" here, read a value from the config (not sure yet how to do that).

MikeMichel:
config via env vars would be also nice https://github.com/kelseyhightower/envconfig

jzaefferer:
@hacdias since you added the label, does that mean you approve the idea?

Could you point me in the right direction for reading the username and password from the config?

hacdias:
Hey @jzaefferer!

Yes, I like the idea and would like to implement it. Unfortunately, I haven't had much time for this. If you'd like to give a hand, you could check this files:

- [Flags](https://github.com/hacdias/filemanager/blob/master/cmd/filemanager/main.go)
- [Where the default pass is set](https://github.com/hacdias/filemanager/blob/master/filemanager.go#L192)

If you need anything, just ping me and I'll try to answer ASAP.

jzaefferer:
Thanks for the pointers. I'll give that a try (or ask a colleague).

jzaefferer:
Its possible that the only required change is this: https://github.com/var2dan/filemanager/commit/351dd8ee1244e231eb77b6efec57cb95dc2ab07b

Not sure if that still merges into current master; probably also needs updates to tests and docs.

eine:
portainer/portainer is a partially open source web app where the backend is written in golang (although the frontend is based on Angular, instead of Vue.js). Having a look at their codebase might be interesting for anyone trying to implement this. See http://portainer.readthedocs.io/en/stable/configuration.html#admin-password

ncresswell:
Funny enough, im actually looking to use this project within Portainer as a way to provide users with a way to download files from existing containers persistent volumes.

eine:
@ncresswell, indeed, I got to know this project when you mentioned it in portainer/portainer#1380 in november. Do you think if the know-how you have in portainer about this issue can be somehow ported here?

ncresswell:
[...]

PR #562, opened 2018-11-08
Manage default user login/password with env var

Airmanbzh:
**Description**
Description
Manage default user/password with env var. If there is no var found, the default will be admin/admin.
This feature will help to deploy a filebrowser instance with docker (no need to connect to the instance to update the main account).


eine:
Thanks for the proposal @Airmanbzh! This was requested in #257 and #381.

I think that your implementation is a good starting point, but it would be better if `viper` was used. We are already using it to set multiple parameters either through configuration files, cli arguments or environment variables. Would you mind adapting it?

You can see how we use `viper` in:

- https://github.com/filebrowser/filebrowser/blob/master/cli/cmd/serve.go
- https://github.com/filebrowser/filebrowser/blob/master/cli/cmd/server.go

PR #565, opened 2018-11-08
fix mp4 in chromium

vitaliy-kuzmich:
Server process octet stream in wrong way


vitaliy-kuzmich:
huh?

eine:
Hi @vitaliy-kuzmich! Does this PR have any follow-up? Was it closed because it was not functional?

Issue #554, opened 2018-10-25
Build failed after hugo updated: undefined: parser.ReadFrom


helloray:
How to resolve it?

helloray:
I use local vendor directory solve this issue.  
cd $GOPATH/src/github.com/filebrowser/filebrowser
mkdir -p vendor/github.com/gohugoio
cd vendor/github.com/gohugoio
git clone https://github.com/gohugoio/hugo -b release-0.49.2
After that, everything is OK. Maybe the better way is to use dep for  dependency management.

eine:
Related to #496.

vitaliy-kuzmich:
> I use local vendor directory solve this issue.
> cd $GOPATH/src/github.com/filebrowser/filebrowser
> mkdir -p vendor/github.com/gohugoio
> cd vendor/github.com/gohugoio
> git clone https://github.com/gohugoio/hugo -b release-0.49.2
> After that, everything is OK. Maybe the better way is to use dep for dependency management.

/lib/http/download.go:54:17: type archiver.Zip is not an expression
../lib/http/download.go:56:17: type archiver.Tar is not an expression
../lib/http/download.go:58:17: type archiver.TarGz is not an expression
../lib/http/download.go:60:17: type archiver.TarBz2 is not an expression
../lib/http/download.go:62:30: undefined: archiver.TarXZ
../lib/http/download.go:75:11: ar.Write undefined (type archiver.Archiver has no field or method Write)

hacdias:
Hello! I actually suggest blurring the line between front-matter and content and leave it in the same editor. That way we can stop depending on Hugo.

The archiver thing I'll open a PR soon, but we also need to define if we're removing Hugo as a dependency or not.

pandarong:
git clone https://github.com/gohugoio/hugo -b release-0.49.2
git clone https://github.com/mholt/archiver
cd archiver
git tag -l
git checkout v2.1.0
replace the archiver ,hugo folder in gopatch,it will build successful.

i will fix the filebrowser with Latest dependency options

benfff85:
Just a friendly inquiry, is this impacting the publication of an updated docker image? I tried this out locally and love it but #522 is preventing it from being used behind a reverse proxy.

Syavne:
Docker build is indeed affected, the following command ends up with the error message:
`docker build --build-arg plugins=filemanager github.com/abiosoft/caddy-docker.git`

[...]

PR #567, opened 2018-11-11
feat: archiver 3

hacdias:
- Remove dependency on go hugo by merging front-matter editor with content.
- Update to use archiver 3.

Ref.: #554 https://github.com/filebrowser/frontend/pull/39


hacdias:
[WIP] not working yet

eine:
I'm not really sure about what's going on here. Let me word what I understood:

- The API of Hugo changed and that broke the build of filebrowser, because we are not using any vendoring/module system ATM.
- At the same time, the archiver dependency also changed, so the build also broke.
- Some users have worked around it by cloning specific versions of hugo and archiver.

If that is ok, the most straigthforward solution would be #496. By using go modules, the build would work without further modifications.

However, at the same time @hacdias suggested getting rid of hugo as a dependency. I think that it makes sense, because AFAIK it is only used for a slightly improved editing experience. I.e., should the site be served, hugo is required as an external dependency anyway.

I am not sure about the relation between archiver and hugo. @hacdias, should these commits be split to separate PRs? Moreover, in the last comment you said that it is not working, yet. Is it because of the frontmatter/hugo part or because of the archiver?

hacdias:
@1138-4EB not working because of archiver. The resultant archives are empty and I still don't know why. Yes, it would be better to separate this into two different PRs. Also, using a vendoring system ASAP would be good.

Does Caddy build server support go.mod @mholt?

lib/http/download.go
```diff
@@ -45,21 +45,21 @@ func downloadHandler(c *fb.Context, w http.ResponseWriter, r *http.Request) (int
...
 		extension string
-		ar        archiver.Archiver
+		ar        archiver.Writer
 	)
 
 	switch query {
 	// If the format is true, just set it to "zip".
 	case "zip", "true", "":
-		extension, ar = ".zip", archiver.Zip
+		extension, ar = ".zip", &archiver.Zip{}
```
mholt, line 54: I recommend using `archiver.DefaultZip` (or if you use more than one, probably `archiver.NewZip()`) -- it sets some sane defaults. Probably especially important for TarGz, which has a default compression level of 0 without sane defaults :wink:

mholt:
> Does Caddy build server support go.mod @mholt?

No, but you can technically vendor some dependencies you need... just make sure they don't conflict with any others that are vendored...

lib/http/download.go
hacdias: Thank you @mholt 😄

hacdias:
Closing in favour of #575

PR #568, opened 2018-11-12
Update README.md

saveshodhan:
Added link of #532 to README so that people landing here can easily go to the issue

Issue #301, opened 2017-12-19
LDAP & Active Directory (AD) Sync


berserkwarwolf:
First of all, this is an awesome App.

I'm writing to ask if there are any plans to implement AD sync. File Manager makes sense for small businesses, and some of us have AD, it would be cool if we could sync our users.

Thank you.

maweck:
I would also be quite interested in having the LDAP for Authorization to combine it with #471 

So in my case the username would be provided through a header, but the filebrowser instance would need to check all other properties (Full name, email, general access rights) from an LDAP. 

We also have some snippets for an LDAP connection in GO available, but I think this ticket is more than just adding it and I guess it also includes some changes in the UI.

@1138-4EB once you have some time to think about this, could you maybe provide some thoughts about the following points:
- Should there be some kind of parameter that allows you to specify if the internal DB or an LDAP should be used as the user base?
- What are the places where this switch has to happen?
- From my point of view filebrowser should show the user management at all when LDAP is configured - Do you agree? (So adding and removing users would be disabled in that case -> It has to be done in the LDAP itself).

eine:
Hi @maweck! IMHO user management and interactions with the database require a rework. See #489. The flows need to be clarified, in order to allow further extensions.

It would be rather straightforward to rely on an LDAP for authentication, as you did with the proxy. However, I don't think that the database can be replaced, because it contains more info than just usernames and passwords. Disabling the addition of new users might be an option; but parameters for existing users should still be editable. Therefore, a mechanism would be required to decide what to do when:

- A user exists in the DB but not in LDAP.
- A user exists in LDAP but not in the DB. I.e. automatic addition of users should be optional (not default).
- A user exists in both the DB and LDAP, and some parameters need to be updated in the DB.
- (Potentially) Some users are set up in the config file and are also found in LDAP.

Furthermore, I think that it should be possible for LDAP and non-LDAP users to coexists in the DB. So LDAP would be an (optional) authentication option and password would be another (optional) authentication method. Indeed, this could be extended to the proxy 'authentication'. So, there would be a precedence order:
- `If` noauth is enabled; go in.
- `Else if` the proxy header is present:
  - `If` the user has proxy access enabled; go in.
  - `else` error or jump to next option (if available).
- `Else if` LDAP authentication is enabled for the user:
  - `If` it exists in the LDAP service
    - `If` it is properly logged in; go in.
    - `else` error (not allowed) and/or jump to next option (if available).
[...]

Issue #574, opened 2018-11-28
Version 2.0


hacdias:
## TODO

- [x] merge: https://github.com/filebrowser/filebrowser/pull/575
- [x] finish docs: https://docs.filebrowser.xyz/
- [x] cleanup https://filebrowser.xyz/
- [x] check if docker image is OK

## After Release

- Caddy plugin release will be tied with next Caddy release which might take a while yet.

---

Hi everyone!

I was looking and the repo and noticed we have already more than 3800 stars and it makes me sad to know that the project is no longer active or even working on master...

In one/two months I'll have a break between uni semesters and I was thinking about giving this a clean up (and to the front-end too: it has a lot of old deps, and I'd like to move it to React).

With that revamp we'd remove the integration with Caddy, at least directly to the code. We could create a plugin that uses or even that downloads the binary. The integration with Caddy is more troublesome that it seems to be and it has always put some constrains on how we structure the code and the development.

In other news, the BoltDB DB is no longer active and I thought we could move to something more straightforward, with more background and with more tooling available such as SQLite.

The point would be to clean-up this repo and frontend, use go.mod and remove Caddy direct integrations (it can always be proxied by Caddy).

Also, with this rewrite, take in account the storage "issues" (#371).

I'm waiting to hear your feedback: @1138-4EB @Equim-chan

hacdias:
- Also, make **filebrowser** mainly a binary and move **main.go** to the root of the repo.
- That would also block FileBrowser from being used as a middleware.
- Use github.com/spf13/afero as a FS layer instead of our own.
- Add an interactive first run experience through the command line to configure the stuff on the database and get rid of almost all flags.
	- This would still let users create instances through scripts, they would just need to provide the stuff in the same order as it is asked.
	- Also add commands to interact with the DB programatically.

helloray:
Hi @hacdias 

I really like this project. Thanks for your great work. I really hope this project can be continued.
Because you are studying in the uni, I think you may not have so much time to rewrite the project. 
I don't know how much difference between VUE and React. If VUE works well, why spending so much time to rewrite it? Those frontend frameworks are almost changing everyday. The most popular framework can be die at any time. I suggest you keep working on features and bug fix. Do not rewrite it.

At the moment, the most important thing is to introduce vendor management for go. Without that,  the project can not be compiled if any go lib changing their APIs. Other contributors can gain benefits from it too.

They are still many features and improvements could be made for this project.
1. Folder upload(I can help to contribute it)
[...]

PR #575, opened 2018-11-29
[WIP] feat(BREAKING CH.): coding cleanup and v2.0.0

### LINKED ISSUE #77: Hugo Plugin - Next Moves
This issue contemplates the next moves on caddy-hugo plugin. The list bellow shows what am I going to do next, and after that, there is a bigger explanation for each item.

- [ ] Theme picker
- [x] Website preview
- [x] Image, audio and video preview.
- [x] Add support for multiple hugo sites within the same site.

### Image, audio and video preview

~~This one is very obvious. The same way we can click on a file on Browse to edit it, we could do the same to preview images/audio and video.~~

**Edit:** this is part of the integration with File Manager plugin.
### Page/post preview mode

~~This page preview mode would replace the current "preview" button that exists on the Edit screen. After clicking on "Preview", the plugin would generate the site in a temporary path. That temporary path would be served at, for example, `/admin/preview/`, and you would be able to preview your page with your theme, styles, css and so on. Then, after closing the tab/window, the plugin would delete the temporary folder.~~

### LINKED ISSUE #246: RFE: allow to set commands for events in config file
I would like to set the commands in the configfile that should be executed on specific events like 'before_save'? IT is not possible AFAIK, is it? What do you think?

### LINKED ISSUE #257: Customizing admin credentials via config/flags

### LINKED ISSUE #314: Edit files without extension

### LINKED ISSUE #326: add option "default sort mode"
It would be nice to be able to configure the default sort mode (byName,bySize,byModified) in config file.

### LINKED ISSUE #350: User Management "Download file" rights suggestion
Hello,

Can new permission flag be added to limit downloading of files, for scenarios that you want multiple users to be able to share same account just to upload files, but not download the files.

Thanks!

### LINKED ISSUE #371: Merge the database and the configuration into a single config file
Hello,

As requested by @hacdias after a discussion on IRC, I am opening this issue to consider the possibility to merge:
- The configuration that is stored in the configuration file ; with
- The configuration stored in the database

Having a single "entry point" to configure filebrowser would be handier.

### LINKED ISSUE #381: Support for password changing via command line or config

### LINKED ISSUE #401: More elegant and efficient way of user management
Filebrowser can only add user each at a time, which is not very elegant not efficient if I want to manage users on a large scale. So I purpose the following suggestions.
1. Add user group features to set some permissions (maybe not all permissions) to users within a certain group globally.
2. Create command line tools (or admin settings) that create users in batches. If user information is stored in `filebrowser.db`, then maybe some command line tool is more convenient.
3. Display customized banner/info for specific user (group)

### LINKED ISSUE #402: Disallow rules do not apply when downloading a folder
### 1. Have you downloaded File Manager from caddyserver.com? If yes, when have you done that? If no, and you are running a custom build, which is the revision of File Manager's repository?
Standalone build, behind nginx reverse proxy, Filebrowser 1.7.0, Ubuntu 16.04 x64
### 3. What are you trying to do?
I set a new user from the admin and create some rules as follows:
```
disallow regex [\\\/]\..+
disallow regex [\\\/].*\.(zip|tgz|xz|tar\.gz)
disallow /bkup
disallow /admin
disallow regex [\\\/].*\.sh
```
These rules pervent that user from viewing all hidden files, compressed file with expected formats, `/bkup` and `/admin` folder in the base folder, and all shell scripts. These rules work perfectly as I logged in from that new user as I see none of those files.
### 4. What did you expect to see?
When I download click `download` without selecting any files from that user, which should download everything displayed on the screen into a compressed format, it gives me **everything in that folder including the files that are disallowed by the rules!!!**
### 6. How can someone who is starting from scratch reproduce this behaviour as minimally as possible?
Try what I have done.

### LINKED ISSUE #420: Custom branding, logo/favicon, turn off github link.
Greetings, 

This is an excellent project and I would like to ask if it would be possible to allow some custom branding in either the command line or config files?  It would be great if it was possible to point the application to a custom favicon file and logo file when integrating this software into a client solution.  The other item that would be nice to have would be the option to turn off hyperlinks.  Having FileBrowser there is fine, but the hyperlink causes issues with external web calls.  It would be great if the "Help" under that was a slightly larger font.  I am not trying to be picky as I have no ability right now to contribute to the code other than cups of coffee.  However, as suggested by Henrique, I am posting these feature requests here.

Thank you for keeping the package fairly light weight as I am currently testing this as a solution for a problem on some embedded devices that I need to remotely manage a couple of files in a very "user friendly" fashion.

Best wishes all!

C

### LINKED ISSUE #424: Rules disallow/allow in config file or globally
It is possible to disallow/allow in config file?
Or way for setup default Rules for new / all users?

### LINKED ISSUE #435: Admin password
Hi there, is there any way to set the admin password when starting filebrowser under Docker?

You already support "--other-flag other-value" but can that be used as "--adminpassword Letmein" for example?

You only list  --port 80, --database /database.db, --scope /srv as flags in the documentation, but is there a complete list of all supported flags?

### LINKED ISSUE #439: Add user register page or API for adding users
Though there is a multi user managment, it is hard to add many users.
So I hope you can add a register page or some api for adding user.

And maybe it is necessary that filemanager should create a folder for a user and change it's scope after which is added.

### LINKED ISSUE #469: Custom upload view/template - query
I'm integrating filemanager with my web service and it's working beautifully.

However, I want to add a custom function, ie to have a custom upload form which I can open in a modal on my website portal. Basically I want to have the "upload button" with the filemanager at a specific directory in the filesystem, to upload to a pre-determined location each time - and that's it, with none of the rest of the usual interface.
I wanted to dive in and hack a template or two, and then I discovered that it's all Go/Rice/Hugo/VueJS funkiness - and I don't have a clue where to start as I'm not familiar with any of the above stack....

Can anybody give me any tips as to where to start?

### LINKED ISSUE #489: Configuration and database rework

### LINKED ISSUE #502: More flexible rights/permission scheme
**More flexibilty to assign rights**

It could be great to explode the right "rename/edit and delete a file".
In fact we can wish to give rights to rename a file but not to delete it.

I don't think it's really hard and if nobody do it until I take a look at go I'll do a PR ;)

In the same order, but more difficult, would be to give rights by folders (more flexibility again but more difficult).

What do you think about that ? :)

### LINKED ISSUE #507: Disable user submitted stylesheets
I'm using filebrowser in a configuration where I have some administrative accounts who submit content and multiple shared user account with different permission levels who access said content. I would like to be able to disable user submitted stylesheets as I would like all shared accounts to have the same user experience and prevent some users to disrupt the experience of other users they share the account with.

A good way to handle that would be to allow or disallow stylesheet submission per user.

Alternatively, it would be good to disable custom stylesheets for the whole filebrowser instance, if that is easier to do in the meantime.

If I have missed that feature in the current version, I apologise for the wrong request. I looked into the documentation and that feature seemed missing to me.

Thanks for your consideration and the work you did on filebrowser.

### LINKED ISSUE #523: User scope in relation to Hugo website root
See #451.

### LINKED ISSUE #555: Hugo: Restrict user scope (access) to subfolder of content. Unexpected "Hugo New" behaviour.
# Is your feature request related to a problem? Please describe.

## Our setup:
We have a small community homepage where kind of a lot of people are allowed to make blog posts. Not edit the rest of our Hugo-Site. 

## Our temporarily solution
So I give them a filebrowser account with scope to ./content/blog/username . I do that because I don't want them to be able to edit each others blog posts. When they intuitively press "Hugo New" they can't just enter the title they have to type /blog/username/title in the title field, otherwise they still are able to create the basic file anywhere in the ./content directory, no matter how scope is defined, but they won't be able to delete or edit this later, so I have to do the cleanup. I can teach them to do that right, but then they see:

home --> ... --> blog --> username --> title.markdown
"This location can't be reached." 

Nevertheless they can select home now and edit their content.

#  Describe the solution you'd like
I would like the limeted-content-scope-users to be able to skip the /content/blog/ part in the title-field  aka set a custom title prefix. Which would be enough to serve this according to the hugo syntax.

Afterwards they should get redirected to their content without the error message, which is no real error.

# Describe alternatives you've considered
Please tell me.

### LINKED ISSUE #572: Symlinks should not always treated as files
**Description**
In the file lists, a symlink is always represented as a file regardless of it being pointing at a file or folder.

**Expected behaviour**
If the symlink is pointing at a folder, it shows it as among the folders with folder icon and possibly with an icon to show this is a symlink (something link a shortcut sign on a folder"

Showing the symlinks that are pointing at a folder among files

### LINKED ISSUE #576: logfile is too simple to be parsed by fail2ban etc
I'm trying to use fail2ban to protect filebrowser brutal-force login attack.

however the log file only output:

2018/11/30 22:19:17 /auth/get: 403 Forbidden

which does not contain the remote-IP, thus fail2ban has no way to know which IP to block.

can we make the logfile more useful?

### LINKED ISSUE #577: POLL: do you use staticgen?
As we are renovating and rewriting for the next version, do you think we should keep Static Gen? Do you use it or not?

React with 👍 if you use it, or 👎 otherwise.

If we removed it, would you be satisfied with:
- Using the command hooks (before/after save,rename, etc)
- Or even custom buttons you could add commands to.

Ping @1138-4EB @Equim-chan @Bitti09 @torch2424 @sinni800 @saveshodhan @chinglinwen @hugomassing @maweck @DeyV @VLEFF @mozzbozz @tokvv @mdlawson @igargi @oopsmonk @dvrkps @pandarong @genofire @maxux @rigon @kairen @coco-perso @ericvolp12 @thomasyxy @llitfkitfk @EduardoRT @princemaple @matm @ptman @thgh @Timmmm @abiosoft @bronzehedwick

### LINKED ISSUE #593: $FILE variable doesn't set properly "Before Delete"
**When did you download File Browser from caddyserver.com?**
2018-12-31

**What is your entire Caddyfile?**

```text
demo.com:443 {
  filemanager / /mnt/CloudDisk {
    database /etc/ssl/caddy/filemanager/database.db
    locale zh_cn
    allowCommands true
    allowEdit true
    allowNew true
  }
}
```

**Description**
A clear and concise description of what the issue is about. What are you trying to do?
I'm trying to implement a function like Rubbish Bin or Recently Delete, move the file to .trash folder instead of delete it directly.

**Expected behaviour**
What did you expect to happen?
File is moved to .trash folder.

Please, give full error messages and/or log.
My script is `cp $FILE .Trash`, and the file to be delete is named `new`, just right in the path `/mnt/CloudDisk/`
When I hit the delete button, server reported Internal Server Error
Here is the logs
```
Jan 01 15:51:18 caddy[14233]: 2019/01/01 15:51:18 [INFO] Blocking Command:"cp $FILE .Trash"
Jan 01 15:51:18 caddy[14233]: cp: cannot stat '$FILE': No such file or directory
Jan 01 15:51:18 caddy[14233]: 2019/01/01 15:51:18 /new: 500 Internal Server Error
Jan 01 15:51:18 caddy[14233]: 2019/01/01 15:51:18 exit status 1
```

And then I saw #483, so I changed my script to `bash -c "cp $FILE .Trash"` with no working:
```
Jan 01 15:58:24 caddy[14233]: 2019/01/01 15:58:24 [INFO] Blocking Command:"bash -c cp $FILE .Trash"
Jan 01 15:58:24 caddy[14233]: cp: cannot stat '/new': No such file or directory
Jan 01 15:58:24 caddy[14233]: 2019/01/01 15:58:24 /new: 500 Internal Server Error
Jan 01 15:58:24 caddy[14233]: 2019/01/01 15:58:24 exit status 1
```

I noticed the second line, said, "cannot stat '/new'", so for testing, I modified the script to
```shell
bash -c "echo file=$FILE"
bash -c "echo $PWD"
bash -c "cp $FILE /mnt/CloudDisk/.Trash"
```

the log now is
```
Jan 01 16:02:59 caddy[14233]: 2019/01/01 16:02:59 [INFO] Blocking Command:"bash -c echo file=$FILE"
Jan 01 16:02:59 caddy[14233]: file=/new
Jan 01 16:02:59 caddy[14233]: 2019/01/01 16:02:59 [INFO] Blocking Command:"bash -c echo $PWD"
Jan 01 16:02:59 caddy[14233]: /
Jan 01 16:02:59 caddy[14233]: 2019/01/01 16:02:59 [INFO] Blocking Command:"bash -c cp $FILE /mnt/CloudDisk/.Trash"
Jan 01 16:02:59 caddy[14233]: cp: cannot stat '/new': No such file or directory
Jan 01 16:02:59 caddy[14233]: 2019/01/01 16:02:59 /new: 500 Internal Server Error
Jan 01 16:02:59 caddy[14233]: 2019/01/01 16:02:59 exit status 1
```

**Additional context**
Add any other context about the problem here. If applicable, add screenshots to help explain your problem.

**How to reproduce?**
Tell us how to reproduce this issue. How can someone who is starting from scratch reproduce this behaviour as minimally as possible?

**Files**
A list of relevant files for this issue. Large files can be uploaded one-by-one or in a tarball/zipfile.

### LINKED ISSUE #594: Copy text via shortcut don't work
Hi

I cannot copy text with shortcut ctrl+c inside of the filebrowser app. it only works if I copy the text via context menu. Issue is not dependent on any OS, I could reproduce the issue on MacOS Chrome, Linux Chromium, Windows Chrome and Mac Firefox.

STEPS TO REPRODUCE
Login to Filebrowser app:
host:port/filebrowser/files/
Mark the small text "File Browser v1.8.0" with mouse
Press ctrl+c
Press ctrl+v somewhere

ACTUAL BEHAVIOR
Some other text you copied before is pasted, not the "File Browser v1.8.0"

EXPECTED BEHAVIOR
"File Browser v1.8.0" should have been pasted

Interestingly, I can copy paste the text on the login screen of the filebrowser app. Issue only happen, if you are logged in. I usually would need this function while renaming files via GUI.

hacdias:
## Breaking changes

- Previous configuration database and configuration files are not compatible.
- Old share links **ARE NOT** compatible with the new configuration hence not imported.
- **DOES NOT** have direct integration with Hugo nor Jekyll.
- **DOES NOT** have user-specific CSS. The admin can customize for the whole installation.
- Command runner `ROOT` variable is now `SCOPE`.
- Command runner `FILE` variable is now an absolute path and not relative to `SCOPE`.

## TODOS

- [ ] fix: files header not scrolling
- [x] add user/global rules via CLI
- [x] Merge  https://github.com/filebrowser/frontend/pull/46
- [x] Merge and archive https://github.com/filebrowser/community/pull/7
- [x] Command to import old DB and Config files.
- [x] Re-build Share Page
- [x] Cleanup README.
- [x] close #572: symlink should not be treated as file
- [x] fix #402: disallow rules are now respected
- [x] fix: folder downloads are empty
- [x] PERF: got slower with the new FS. Maybe don't use the userFs and create MatchRules that can take many rules as arg (variadic).
- [x] close #424: global allow/disallow rules
- [x] FIx #594
- [x] fix #593 
- [x] Add branding to settings
- [x] Re-build Command Web Socket
- [x] Make it work with just `filebrowser -s /path`. 
- [x] Re-build Search Web Socket
- [x] close #583: dirs cannot be opened by click
- [x] Pick new code editor
- [x] rebuild settings
- [x] close #326: add default sort mode
- [x] close #420: custom branding
- [x] Built-in HTTPS support
- [x] Re-add Travis
- [x] close #576: add IP on logging
- [x] close #469: go.mod
- [x] close #358: edit every text file
- [x] close #314: "
- [x] close #568: limit the size of text files to open
- [x] close #562: closed via CLI tooling
- [x] close #381: "
- [x] close #257: "
- [x] close #435: "
- [x] close #401: "
- [x] close #246: "
- [x] close #371: CLI tooling + unique DB + no config file
- [x] close #502: more permissions (not by folder though, still by user)
- [x] close #350: " (download perm)
- [x] close #567: archiver 3 implemented
- [x] addresses #301: `Auther` interface
- [x] close #577
- [x] close #77: no Hugo integration
- [x] close #523: "
- [x] close #555: "
- [x] close #507: no per-user CSS, only global
- [x] close #439: "signup" is a global setting
- [x] close #489: database rework


hacdias:
Update: decided to reuse the current Front end in Vue and update the dependencies. More here: https://github.com/filebrowser/frontend/pull/46

hacdias:
@1138-4EB @Equim-chan this only has a database and no configuration file. My suggestion for automated installations is to use a script or something like that to create the users and change the configuration. Most of the config options will now be accessible through the web interface too.

If anyone wants a config file back, we can think about it later, but this is a major release, so shouldn't mind about breaking changes because there are many.

I'll create command to import an old config into the newer database format. Any other suggestion?

hacdias:
Also, decided to keep `http.filemanager` plugin for Caddy. Only `http.hugo` and `http.jekyll` will be removed.

wizard.sh
```diff
@@ -0,0 +1,299 @@
...
+  echo "Repo:                     $REPO"
+  echo "Will lint:                $LINT"
+  echo "Will build:               $BUILD"
+  echo "Will release:             $RELEASE"
+  echo "Will push latest docker:  $PUSH_LATEST"
+  echo "Use Docker:               $USE_DOCKER"
+  echo "Is CI:                    $CI"
+}
+
+dockerLogin () {
```
eine, line 22: This is mostly meant for Travis CI. Locally, the user should handle the login before running the script. This is because a specific version of `docker-credential-pass` is distributed inside `filebrowser/dev`. I.e., on Windows, `dockerLogin` will probably fail. It will also fail in any non-amd64 machine.

So, it is ok to keep this function here. But I think we should add a comment about it.

hacdias: We can also test if is a ci using the `CI` env var. Otherwise we just do nothing on that function. WHat do you think?

wizard.sh
```diff
@@ -0,0 +1,299 @@
...
+  echo "Updating version from \"$from\" to \"$to\""
+  sed -i.bak "s|$from|$to|g" $REPO/types/version.go
+}
+
+buildBinary () {
+  cd $REPO
+  go get -v ./...
+  updateVersion $untracked "($COMMIT_SHA)"
+  echo "Build CLI"
+  go build -a -o filebrowser
```
eine: In `filebrowser/dev` envvar `CGO_ENABLED=0` is set. See https://github.com/filebrowser/docker-dev/blob/master/Dockerfile#L12. Thus, this script will produce a portable binary when executed inside `filebrowser/dev`. Otherwise, the binary will be static but it will depend on the standard library available on the system.

Therefore, it might be desirable to prefix line 88 with `CGO_ENABLED=0`.

wizard.sh
eine: Agree.

hacdias: CHanged that, please take a look @1138-4EB

wizard.sh
hacdias: Isn't CGO disabled by default though? We can absolutely add it anyway eheh

hacdias: Added!

wizard.sh
hacdias: @1138-4EB maybe if it's not CI, we shouldn't do `docker login` and `docker logout` and just bypass those steps. Probably the user is logged in already. Agree?

wizard.sh
eine: It is enabled by default for native builds, and disabled for cross-compilation. See: https://golang.org/cmd/cgo/

> The cgo tool is enabled by default for native builds on systems where it is expected to work. It is disabled by default when cross-compiling. You can control this by setting the CGO_ENABLED environment variable when running the go tool: set it to 1 to enable the use of cgo, and to 0 to disable it.

hacdias:
@1138-4EB hope you liked the merge of the scripts into one! Although there seems that for some reason it can't find [wizard.sh](https://travis-ci.com/filebrowser/filebrowser/jobs/167632682#L463-L465) but dos2unix works...

eine:
> @1138-4EB hope you liked the merge of the scripts into one! Although there seems that for some reason it can't find wizard.sh but dos2unix works...

It looks good! I am having a look at that error. Unfortunately, I need to leave now. So I'll do it next year ;). Have a nice new years eve!

hacdias:
~@1138-4EB it was because I changed from `sh` to `bash` to make sure `getopts` existed and didn't differ in behaviour. I fixed it by adding `bash` to filebrowser/dev.~

Actually `getopts` is POSIX-compliant so it must be present on all POSIX-compliant systems. Reverted the changes to filebrowser/dev and replaces all `==` by `=`. Now there seems to be a permissions issue (?).

Thanks! Have a nice New Year's Eve!

hacdias:
@1138-4EB I believe the error might be along these lines:

https://github.com/filebrowser/filebrowser/blob/8d334046a98a59b60ba3de401ffcb7c620a6b681/wizard.sh#L202-L204

I just don't understand why we didn't have the errors before. Another question: why are we copying the files? Since we mounted the volume, wouldn't the files persist?

**Edit:** for what I just analyzed we just need the `filebrowser` and `rice-box.go` files. I see you did this `cp` thingy to avoid the permissions issue caused by the commands inside of Docker running as root. But there must be another way of doing this.

eine:
> Actually getopts is POSIX-compliant so it must be present on all POSIX-compliant systems. Reverted the changes to filebrowser/dev and replaces all == by =. Now there seems to be a permissions issue (?).

Indeed. The original issue was that `bash` is not the default shell in Alpine Linux. So it failed because of the shebang. Moreover, `bash` is not always available. E.g., the Ubuntu version used in travis includes `dash`. Therefore, I think it is better to make the script the most general, instead of requiring `bash`.

> I just don't understand why we didn't have the errors before. Another question: why are we copying the files? Since we mounted the volume, wouldn't the files persist?
> 
> Edit: for what I just analyzed we just need the filebrowser and rice-box.go files. I see you did this cp thingy to avoid the permissions issue caused by the commands inside of Docker running as root. But there must be another way of doing this.

You are ok:

- The error is produced in those lines.
- Yes, files persist because we mounted/bind the volume.
- Yes, we only need `filebrowser` and `rice-box.go`. We copy `dist` and `node_modules` too because:
    - `dist`: in case the user wants to distribute the frontend.
    - `node_modules`: to avoid downloading all the dependencies each time.

Certainly the cp are redundant ATM. A few days ago the content was mounted/bind at `/src` inside the cointainer, but it was then copied to `go/src/github.com/filebrowser/filebrowser`. Therefore, it made sense to cp the artifacts. However, right now, we are trying to copy a directory on itself. That's probably the source of the error.

**EDIT**

I think it is fixed now. See branch `fix-wizard`.

eine:
> @1138-4EB @Equim-chan this only has a database and no configuration file. My suggestion for automated installations is to use a script or something like that to create the users and change the configuration. Most of the config options will now be accessible through the web interface too.

I find it interesting to have access to all the config options through the interface.

But, as commented in #574, I think that supporting configuration files is desirable too. It is quite easy to get info from a form and generate a JSON/YAML with it. Requiring a custom script involves an intermediate step to translate the config data to a suitable set of instructions.

> If anyone wants a config file back, we can think about it later, but this is a major release, so shouldn't mind about breaking changes because there are many.

As you say, we can add it later. For now, I think it is better that you work on your own until you feel the code to be ready for review. At that point we will evaluate if we should add config files before or after publishing v2.

hacdias:
@1138-4EB I merged your fix here but I wonder if this will have any repercussions over `dockerPushLatest` since the files were created as `root` user. Probably there won't be an issue because it just copies the files and they have read permission and the user inside the Docker image is root.

eine:
It should not have any effect on `dockerPushLatest`. The single annoyance now is on GNU/Linux hosts: when the user wants to remove the artifacts (`filebrowser`, `frontend/dist`, `rice-box.go`...) root permissions are required. This is annorying because `sudo` should not be required to remove a repo you cloned just to build a binary. However, sudo privileges are required in order to run docker, so we can expect the user to be sudoer if he/she ran docker.

hacdias:
@1138-4EB we can work out the filebrowser/dev image so the user **inside** is not sudo. I just [read this](https://medium.com/@mccode/processes-in-containers-should-not-run-as-root-2feae3f0df3b).

It's 9:30am here and I'll only work on File Browser after midday. If you want to, you could try creating an user on filebrowser/dev with the least required permissions to run everything. If you don't have time, I can try doing so after lunch. What do you think? 

**Edit:** change of plans: I'm working a bit on this now. I'll solve some other issues not related to this comment. Waiting for your feedback.

eine:
Excuse me @hacdias. I went to sleep and skipped your previous comment. I am going to work on `filebrowser/dev` now. On the one hand, I want to support running gometalinter, yarn and go without root privileges. On the other hand, I will add golangci-lint. I will keep you posted.

eine:
@hacdias I added `-u $(id -u)` to `docker run`. Now, all the commands are executed with the same user id as the one on the host. This ensures that the user will not require any privilege to remove the artifacts. Expectedly, both yarn and go complained about it:

- Go tried to create `/.cache`. I fixed it setting `XDG_CACHE_HOME=//tmp/.cache`. See [docker-library/golang#225](https://github.com/docker-library/golang/issues/225#issuecomment-403170792). However, it seems that we might need to use `GOCACHE` in future versions: https://github.com/golang/go/issues/26280#issuecomment-445294378
- Yarn complains about `Error: EACCES: permission denied, open '/.yarnrc'`. Depite it, the build works. I did not find a solution yet, so I just used `set +e`.

You can see these changes in https://github.com/filebrowser/filebrowser/commit/895ecb1251656ec5c5f566fff8079dc54e4b4484

hacdias:
@1138-4EB hmm... wouldn't docker fail if the host user id doesn't exist inside the docker image? it might happen

eine:
Docker only fails if you need user specific data/configs, which is not the case. In the end, the yarn error was similar to golang's: `HOME` is unset when the user is not defined. So I just set `HOME=/tmp` as a container envvar.

hacdias:
@1138-4EB there are already a few things left. I think it would be nice to merge this as soon the share page is rebuilt. That way we can work on master. This PR is getting too big for a PR.

We won't release 2.0 soon. Also, if people use the latest docker image and complain for breaking changes we just need to remind them that latest means master and master might be incompatible with previous versions. We should always recommend to use the latest version and not the latest tag in production.

Merging this will also help me re-build the Caddy plugin. It will be really small I think 😄 so I'm also happy for it.

We could also try out the docker image and the 🧙‍♂️!

What do you think?

eine:
@hacdias, sure. I am going to publish v1.11.0 now, so that we have a stable v1 and we can move forward with v2 on master.

About the pending task named *fix: Docker image*, is there anything I can help test/fix?

hacdias:
@1138-4EB release it! Thanks! I merged the frontend already so make sure not to update it on master eheh.

It's more wait to see if the Docker image works well and if the wizard builds it correctly 😄

eine:
Well, the release stage did need some effort (because of filebrowser/docker-dev#3), but it is working ok now. Plus, the good news are that both golangci-lint and goreleaser work seamlessly when the repo is outside of GOPATH.

I created a new branch named `v1` to keep possible bugfixes there. Then, I merged `master`/`v1` to `rewrite`, and named it `v2`.

You can now:

- Forward `rewrite` to `v2`, and then merge this PR (forwarding `master` to `v2`).
- Forward `master` to `v2` and close this PR manually.

hacdias:
@1138-4EB thanks! I kept `v1` and merged v2 into master. Although I don't feel comfortable in keeping the version 1 in a branch. It makes me feel like Python where two versions keep growing separately for no reason. We can keep it this way and eventually remove it in the future. We always will have the tags and I doubt there are any reasons to keep on v1.

hacdias:
Soon will have a v2 released!

eine:
So fast! Nice!

Certainly, we could remove both branches, `v1` and `v2`, and use `master` only:

- `v2` was just temporal, indeed.
- With `v1`, we need to remember that any bugfix should start from 013ddf45c71248c6f1e451dc31f09744e43ae6aa, which is the commit next to the tagged one (v1.11.0). That's why I think it can be helpful to keep it. If we see that no critical fixes need to be applied, we can remove it in the future.

eine:
BTW, I tried the windows-amd64 binary and the docker image corresponding to v1.11.0. Both work as expected. However, caddy is failing because of some of the changes we did to filebrowser/dev (such as changing the linter tool). I'm going to fix it.

hacdias:
Thanks. Don't worry about Caddy for now. Our plugins can't be downloaded right now. I set them as hidden on Caddy website until we have a workable version with 2.0. Caddy doesn't let us use dependency management with go mod so I was thinking about vendoring or something like that in the plugin. It will need a bit more thought though.

eine:
Ok. So, I will review the Dockerfile thing. I downloaded the latest filebrowser image (v2) and it refuses to start due to some problem with the entrypoint.

hacdias:
Yes, please do that! Also check if import command works as expected (ref https://docs.filebrowser.xyz/upgrade-from-1.x)

eine:
After having a quick look at it, I think that we should avoid `docker-entrypoint.sh`. If that feature is really required, it should be integrated in the golang codebase.

hacdias:
@1138-4EB let's continue on https://github.com/filebrowser/filebrowser/issues/600

PR #582, opened 2018-12-12
Manage default user with env variables

Airmanbzh:
Description
Manage default user/password with env var **using viper**. If there is no var found, the default will be admin/admin.
This feature will help to deploy a filebrowser instance with docker (no need to connect to the instance to update the main account).

PR already done [here](https://github.com/filebrowser/filebrowser/pull/562) but I accidentally deleted (and pushed) my original commit.


eine:
Thanks @Airmanbzh!

PR #586, opened 2018-12-25
Add go mod support

### LINKED ISSUE #554: Build failed after hugo updated: undefined: parser.ReadFrom

### LINKED ISSUE #579: Dependency management is in urgent need
It took me more than half hour to solve the dependency problem.  It needs more work than just doing `go get`. Some packages changed to a total different API in master branch, so I have to check the git log and try to choose some history commit  by luck. 

For examle, `github.com/mholt/archiver` changed to new API after 5105dc13ef4742e45ae9fcc7e98fa413cb7ca876. The new API cannot be used in the project.

**Suggestion**: use vendor dir to include all dependencies to this repo.

eine:
Close #554, close #496 and close #579.

This PR adds `go.mod` and `go.sum` files. At the same time, scripts are simplified according to docker-dev#3.

However, I don't think this is ready to merge yet. ATM filebrowser can be succesfully built, because hugo and archiver versions are properly fixed. However, gometalinter fails. I switched the order of the stages in travis to show it. See: https://travis-ci.com/filebrowser/filebrowser/builds/95695921

Any thoughts about how to fix stage `lint`?


hacdias:
It seems that some packages are not imported correctly. Can you `go build` this PR on your OS? Does it work? If so, I'd say to ignore the error for now or even remove gometalinter for now until we finish #574

eine:
> It seems that some packages are not imported correctly.

You mean in the `lint` stage, don't you?

> Can you go build this PR on your OS? Does it work?

Yes, I can. And, yes, it works.

> If so, I'd say to ignore the error for now or even remove gometalinter for now until we finish #574.

`gometalinter` might be a *problem* even after we finish #574, as it seems that go modules are not supported: alecthomas/gometalinter#521. Are you using gometalinter in the rewrite?

eine:
In the end, I could make gometalinter work. After running `go get` and `go mod vendor` with `GO111MODULE=on`, gometalinter can be executed with `GO111MODULE=off` without dependency issues.

I am merging this. But I think we might consider changing to golangci/golangci-lint. See https://github.com/golangci/golangci-lint#comparison

PR #588, opened 2018-12-28
fix: move everything to LF

hacdias:
License: MIT
Signed-off-by: Henrique Dias <hacdias@gmail.com>


build/build_all.sh
```diff
@@ -25,8 +25,6 @@ if [ "$USE_DOCKER" != "" ]; then
     filebrowser/dev \
     sh -c "\
       cd build && \
-      dos2unix build_assets.sh && \
```
eine, line 28: On windows, if `autocrlf` is set to `true`, git will replace LF with CRLF. So, when docker is used on windows, executing these scripts inside linux containers might fail. That's the rationale to use `dos2unix`.

hacdias: @1138-4EB you're right! Will revert those two lines 😄

Issue #600, opened 2019-01-05
New Dockerfile


hacdias:
Continuation of https://github.com/filebrowser/filebrowser/pull/575#issuecomment-451699904

@1138-4EB I added the entry point so the user wouldn't need to set the `--port 80` flag if they wan't to add more flags to the command. I saw that in postgres Docker image.

We could remove it, but then the user would put the `--port 80` if they add anything else to the command.

```Dockerfile
FROM scratch

COPY --from=filebrowser/dev /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/ca-certificates.crt

VOLUME /tmp
VOLUME /srv
EXPOSE 80

COPY filebrowser /filebrowser

ENTRYPOINT [ "/filebrowser" ]
CMD [ "--port=80" ]
```

If you don't add anything after the image name, it will run `filebrowser --port 80`. If you add `x`, it will run `filebrowser x` and remove the `--port 80`. This is because we can't have the flag always there. Some commands would fail.

But yes, we can remove the entrypoint. There is a quick setup when the DB doesn't exist and the user doesn't call `filebrowser init`.

eine:
I think that we can handle this situation as we did in `v1`: if no `--port` flag, `FB_PORT` envvar or configuration option is provided, the port defaults to `80`. So the user only needs to set it if he/she wants a different port.

I am more concerned about the location of the database. In `docker-entrypoint.sh` it deaults to `/database.db`, which I find to be *ugly* outside a container. It should probably default to `./database.db`.

hacdias:
But that value is **inside** the container (/database.db). THat's the path you need to mount something to.

Also, as part of the refactoring the ENV vars were removed. We're trying to leave the configuration as simple as possible. But I'm interested in hearing use cases.

eine:
Despite the configuration being simplified, we should not constrain the deployment too much. For example, begin able to set runtime parameters such as the port, the address, the location of the DB, etc. should be easy. I.e., the default location of the database can be `/database.db`, but the user should be able to set it in any arbitrary path (either inside or outside of the container).

hacdias:
Yes and they can do all of that via flags. In which situations wouldn't flags be possible but env vars would? That's just what I don't understand

XavierBouvard:
Hi,

If I may, I am also more concerned about the database file than the port.

I agree with @1138-4EB: it is a good idea to naturally fallback to 80 if none specified in any kind of configuration.

[...]

PR #601, opened 2019-01-05
feat: dockerfile and viper

hacdias:
Addresses #600.

@1138-4EB can you take a look please?


hacdias:
The thing with this is:

- `docker run filebrowser/filebrowser` == `filebrowser -d=/database.db -p=80 -s=/srv`
- `docker run filebrowser/filebrowser --port 80` == `filebrowser -d=/database.db -s=/srv`

Also, note that **scope** flag is only required to create a database. After that it will be simply ignored. This is why I recommend using `config init` so everything is clear to the user and there are no unnecessary flags.

eine:
I executed the following commands in this branch:

``` bash
USE_DOCKER="true" ./wizard.sh -b
docker build -t filebrowser/filebrowser .
winpty docker run --rm -it filebrowser/filebrowser version
```

I get this output:

``` bash
Error: unknown flag: --database /database.db
Usage:
  filebrowser [flags]
  filebrowser [command]

Available Commands:
  cmds        Command runner management utility
  config      Configuration management utility
  help        Help about any command
  import      Imports an old configuration
  init        Initialize a new database
  rules       Rules management utility
  users       Users management utility
  version     Print the version number

Flags:
  -a, --address string    address to listen on (default comes from database)
  -c, --cert string       tls certificate (default comes from database)
  -d, --database string   path to the database (default "./filebrowser.db")
  -h, --help              help for filebrowser
  -k, --key string        tls key (default comes from database)
  -l, --log string        log output (default comes from database)
  -p, --port int          port to listen on (default comes from database)
  -s, --scope string      scope for users

Use "filebrowser [command] --help" for more information about a command.

2019/01/06 01:43:28 unknown flag: --database /database.db
```

That's because `ENTRYPOINT [ "/filebrowser", "--database /database.db" ]` treats `--database /database.db` as a single argument. It should be `ENTRYPOINT [ "/filebrowser", "--database", "/database.db" ]`.

---

> The thing with this is:
> 
>     * `docker run filebrowser/filebrowser` == `filebrowser -d=/database.db -p=80 -s=/srv`
>     * `docker run filebrowser/filebrowser --port 80` == `filebrowser -d=/database.db -s=/srv`

I am not quite sure. I think the second one would be: `docker run filebrowser/filebrowser --port 80` == `filebrowser -d=/database.db -p=80`. Unless you mean that `-s=/srv` is already set, because you assume that the first command has been executed earlier. That's not intuitive to me. IMHO, it would be easier if the initialization was done automatically when the provided database is empty (or it does not exist).

---

Moreover, related to this, it seems that once the database is created the root of the scope is fixed (`/srv` in the example above). If so, can it be reworked so that it can be set at runtime. I.e., make all the refs in the DB relative to the root scope.

eine:
@hacdias, please have a look at https://github.com/filebrowser/filebrowser/compare/dockerfile...feat-viper. It is some WIP to set defaults with spf13/viper. At the same time, it allows to set 'quickSetup' params with either a config file, CLI flags or envvars. From there on, all of those that conflict with the DB are ignored, unless flag `--force` is used.

princemaple:
+1 having the defaults defined in the code and merging all possible configs into it.

hacdias:
@1138-4EB there was never a root scope, not in v1.11.0, neither here. The `scope` flag only exists for quick setup. After that it is considered the default scope for new users. You can set it on their settings.

We can, though, add a root scope to the global settings that will be prepended to users' scope on runtime. And that can be overwritten by the flags.

---

About the command, yes, you're right eheh. My bad.

---

I feel this: https://github.com/filebrowser/filebrowser/compare/dockerfile...feat-viper is overcomplicating. I agree we can use Viper to get the base command vars either from flags, config file and env vars, but isn't there a simpler way?

And I see Viper has that bug that can't know if some setting was really set or not.

hacdias:
@1138-4EB I merged your changes into this branch. Just to make sure: we're only adding Viper to root command right?

hacdias:
Also @1138-4EB:

> database has highest priority, if undefined in DB use config params

How can you tell that a number is undefined for example? We have the PORT to run on. If the user sets it to zero themselves we cannot assume that it is undefined. Unless we could set it to a negative number.

---

I'll let you work on this PR. I'll make a PR for scopes being relative if a global scope is set and document it well.

eine:
> @1138-4EB there was never a root scope, not in v1.11.0, neither here. The `scope` flag only exists for quick setup. After that it is considered the default scope for new users. You can set it on their settings.
> 
> We can, though, add a root scope to the global settings that will be prepended to users' scope on runtime. And that can be overwritten by the flags.

Certainly, I did not try it with v1 neither. However, I feel it to be a cool feature to be able to rellocate either the database or the content/data. Having a root scope set on runtime, allows to move that folder without any modification to the DB.
 
> I feel this: [dockerfile...feat-viper](https://github.com/filebrowser/filebrowser/compare/dockerfile...feat-viper) is overcomplicating. I agree we can use Viper to get the base command vars either from flags, config file and env vars, but isn't there a simpler way?

I wish there was 😕 , but that's the simpler I could achieve. We need to define the flags, the defaults and bind the flags to viper. Nevertheless, I think that the tricky part is still to be written. I commented out `serverVisitAndReplace`, but it should be rewritten to take `--force` into account.
 
> And I see Viper has that bug that can't know if some setting was really set or not.

Unfortunately... It works for cobra, so we can tell if a CLI flag was used. But it does not work for envvars. Anyway, we can work around it by checking against the defaults by ourselves and displaying a warning if they match. I believe there is no practical difference between a flag being unset (thus taking the default value) and the user setting a flag to a default value.

> @1138-4EB I merged your changes into this branch. Just to make sure: we're only adding Viper to root command right?

For now, yes. But should probably add it to `config.go`, `import.go` and `users.go` too.

> How can you tell that a number is undefined for example? We have the PORT to run on. If the user sets it to zero themselves we cannot assume that it is undefined. Unless we could set it to a negative number.

With 'undefined' I mean 'Field not found in the DB'. For example, 'baseaddr', 'address' and 'port' might not exist in the DB. In that case, the runtime params can be used, if set. Otherwise, defaults are used. Overall, these runtime parameters could be written to the DB only when the values differ from defaults.

I know that with your initialization procedures all the fields will always exist. So let's consider the paragraph above an error callback procedure, just to avoid panicking with incomplete DBs.

> I'll let you work on this PR. I'll make a PR for scopes being relative if a global scope is set and document it well.

Great.

hacdias:
@1138-4EB I don't agree we should add Viper to other commands than root. This is the only one that has stuff that can change and that needs to be persisted across multiple runs.

Config and user commands are meant as one time runs to do a specific task.

Import is to import config from 1.x only.

eine:
I just listed the files which had flags. After checking the content, I agree that import flags are not worth. However, for config and users a configuration file might be useful. I am thinking of having an initialization config file. If I want to initialize an empty DB and set some branding, authorization and default permission, I'd prefer to have all that content in a JSON/YAML/TOML rather than a shell script with CLI args.

hacdias:
Yes, for `config init` I agree. If you could work on that, it would be awesome. But only for `config init` eheh.

PR #606, opened 2019-01-06
feat: dockerfile & viper

hacdias:
Continuation of #601


hacdias:
@1138-4EB made some changes and merged master into this branch. I like the idea and how it works but there is something that annoys me:

```
PS D:\root\Code\filebrowser\fb> go run .\main.go  users ls
2019/01/06 13:45:19 No config file provided
ID  Username  Scope  Locale  V. Mode  Admin  Execute  Create  Rename  Modify  Delete  Share  Download  Pwd Lock
1   admin     .      en      mosaic   true   true     true    true    true    true    true   true      false
```

I find it annoying to print that info on other commands.

Btw, if you think it's a good idea, we can remove the `init` command and do a quick setup on `initConfig` if the database doesn't exist. Although I think it would be better to ask for user input if the command is not root.

Imagine someone is trying to list the users but has a mistake on their DB file: they don't want to generate a new database file. It was just a mistake.

hacdias:
@1138-4EB updated the help entry of the main command. I think you'll like it.

hacdias:
@1138-4EB I've commented the info about the current config file right now and will merge this. Although I hope to remove those TODOs and find a prettier solution. Will wait for your feedback!

eine:
Wow mate! Hard to follow your rythm ;) Just an overall comment: I'm quite happy about how easily we are discussing and applying all these changes. Kudos to you!

> I find it annoying to print that info on other commands.

That's because `initConfig` is executed always, before any command. But we can move it or add some condition, so it is only shown when using a config file makes sense.

> Btw, if you think it's a good idea, we can remove the init command and do a quick setup on initConfig if the database doesn't exist.

I think there is no need to remove `init`. I'd like the functionality to be available either in initConfig or in the root command (as fallback for an empty/non-existent DB). But I also think that some users will feel more comfortable executing `init` explicitly.


> Although I think it would be better to ask for user input if the command is not root.
>
> Imagine someone is trying to list the users but has a mistake on their DB file: they don't want to generate a new database file. It was just a mistake.

Sure. I added it to `initConfig` because I was only testing the root command. But *automatically* setting the DB should only be done when rootCmd is executed.

> @1138-4EB I've commented the info about the current config file right now and will merge this. Although I hope to remove those TODOs and find a prettier solution. Will wait for your feedback!

Will have a look at it.

hacdias:
Agreed with everything! Will keep waiting for more feedback on the TODOs and printing.

Actually, the config file makes **always** sense because it can contain the database path. I think we have two options:

1. Only use the config file on root (by calling `initConfig` on `rootCmd` handler)
2. Always use the config file and have a nicer way to print if we're using a config file. Or have yet another flag to print the **database** path. Not the config path. If the database path is wrong and the user is using a config file, it is clear that the config path is wrong.

eine:
About setting the default username and password, in the current `master` the password is expected to be given plain as a cli arg. However, in v1 the password hash is expected. See #591.

hacdias:
@1138-4EB oh, I can change that then. Although how's the user supposed to hash the password previously? I know it's possible but we should also give the resources to hash the password. I'll create a simple subcommand to hash a password.

eine:
@hacdias, what's the difference between `init`, `config init` and quickSetup?

hacdias:
init is an alias for config init. It uses the flags to get all possible options. Quick setup is just used when you try to start File Browser without having a db first.

eine:
Then, we might remove `init`. I find `config init` more explicit, as `init` might mean to initialize a service.

hacdias:
Sure yes! 👍

PR #611, opened 2019-01-06
[WIP] Viper config

eine:


cmd/root.go
```diff
@@ -161,11 +166,13 @@ func serveAndListen(cmd *cobra.Command, args []string) {
 	}
 }
 
-func quickSetup(cmd *cobra.Command) {
+func quickSetup() {
 	db, err := storm.Open(v.GetString("database"))
 	checkErr(err)
 	defer db.Close()
 
+	// TODO: save also port, address, cert, key, scope; if their values differ from defaults
```
hacdias: That's what we did like 3/4 commits before. I removed it to avoid scattering a lot of information. If we have a config file, why would we need to access those fields? `scope` and `baseurl` are special cases since we need them inside the http handler. What do you think?

cmd/users_update.go
```diff
@@ -21,23 +21,19 @@ var usersUpdateCmd = &cobra.Command{
 options you want to change.`,
 	Args: cobra.ExactArgs(1),
 	Run: func(cmd *cobra.Command, args []string) {
-		db := getDB()
-		defer db.Close()
```
hacdias: Aren't there issues by not closing the DB?

cmd/utils.go
```diff
@@ -61,21 +62,26 @@ func mustGetUint(cmd *cobra.Command, flag string) uint {
...
+		log.Fatal(databasePath + " does not exist. Please run 'filebrowser init' first.")
 	}
 
 	db, err := storm.Open(databasePath)
 	checkErr(err)
-	return db
-}
 
-func getStorage(db *storm.DB) *storage.Storage {
+	defer db.Close()
```
hacdias: If you defer it here, it will execute before returning. Hence, the storage returned by this function won't have access to the DB. That's the reason I was copying the code over.

cmd/users_update.go
eine: Yes, they were. It should be fixed now.

cmd/utils.go
eine: I hoped that closing would be deferred until either `st` or `s` are not used anymore; but it applies to the scope of the function, as you say. So I fixed it by deferring outside.

cmd/root.go
eine: I'd like to respect the idea that a DB is enough. So, if anyone configures an empty DB with `config init` and he/she does not need to change any settings, copying the DB should be enough to backup/reproduce the setup.

On top of that, config files are an optional complement to make setup and/or rellocation easier. However, if you are OK with making the config file necessary, we can keep it as it is now.

hacdias: Like, I don't mind of readding it. But I personally don't like the idea of the `--force` flag very much. Why would someone want to provide values **and** forcefully use the ones from the database?

eine: I think that someone might want to have the baseurl, cert and key fixed in the DB, but set port and address from a config file, while other fields are ignored. Then, for some one-time test the cert and key can be overset. This allows the user to try new configs without modifying the DB until the changes are proven to work.

Nevertheless, why do `scope` and `baseurl` need to be in the DB? The http handler could equally get those fields from viper, isn't it?

hacdias: @1138-4EB it could, but then we wouldn't be able to use File Browser as a middleware since the installation would be bound to viper.

Maybe I got it wrong: is the force flag to use the db or the flags? I'd prefer to have the following priority without needing to use a force flag:

- Flags/env/config via viper
- If not set fallback to database

hacdias: We could actually send the values to the HTTP handler without being in the Settings struct though.

hacdias:
@1138-4EB had an idea that might work for the following precendence: Viper then DB without needing the force flag and the function ´IsSet´. Do you mind if I make some updates to the code if it works?

eine:
No problem. But I don't get it. I mean, `IsSet` is provided by Viper. Furthermore, it does not work as expected, so we cannot rely on it.

hacdias:
@1138-4EB exactly! And we won't need it! If my idea works... of course. Otherwise, we'll need force.

cmd/config_init.go
```diff
@@ -28,19 +26,18 @@ to the defaults when creating new users and you don't
...
+			log.Fatal(databasePath + " already exists")
 		}
 
 		defaults := settings.UserDefaults{}
 		getUserDefaults(cmd, &defaults, true)
 		authMethod, auther := getAuthentication(cmd)
 
-		db, err := storm.Open(databasePath)
-		checkErr(err)
+		db, st := getStorage()
```
hacdias: I'll revert this. Take a look at `getStorage` code and you'll see why it can't be used on `config init` specifically. Wait for a commit here soon 🛩

hacdias:
@1138-4EB please don't make any changed yet. :/ This will conflict with the changes I'm doing

eine:
> @1138-4EB please don't make any changed yet. :/ This will conflict with the changes I'm doing

Ok. Just remove the commits that don't fit and I'll test in a different branch meanwhile.

cmd/root.go
eine: I'm sorry, I skipped these two comments.

I am interested in that possible conflict to use File Browser as a middleware if we rely on viper. We can explicitly set values for keys in viper, even if no CLI args, no envvars and no config files are provided. However, if that is still a limiting factor, we can keep those runtime parameters in a global variable.

> We could actually send the values to the HTTP handler without being in the Settings struct though.

This is my main concern. If you see my last commits, I'm trying to replace fields BaseURL, Key, Scope/Root in the DB with a Runtime field of type `map[string]string`. This change allows to save none or many runtime parameters in the DB (only those that differ from defaults). However, the implementation I was working on does not read those values from the DB each time. Instead, an initial check is performed in `serveAndListen` to explicitly set viper keys with DB content, if suitable.

hacdias:
@1138-4EB the idea I had doesn't work unfortunately. Although I still have some improvements to introduce to this branch...

One more thing: I think that adding Viper to `config` will overcomplicate things. We could have:

1. `config import <file>`
2. `users import <file>`

And just read the files. And **only** use Viper with the root command.

Also, I don't believe using `force` is a good practice so I think we should find another way to do so. In the meanwhile, I think you should "dump" this PR and create the command I mentioned previously. Maybe that would be the best. THen we can think about how to improve root flags.

Awaiting for feedback.

hacdias:
@1138-4EB the idea I had doesn't work unfortunately. You know what you feel after making a lot of changes and in the end it doesn't work as expected 😭 

One last thing about this PR: I think we should only use viper for the root command. The others don't make sense to make use of it.

Yes, we should allow users to import config and users. So I believe we should create the following two commands:

- `config import <file>`
- `users import <file>`

Same applies for export.

And just read the files. Maybe see if viper exports their parser (or see which parser they use) because it's canse insensitive and it would be helpful for users. And only use Viper with the root command.

For root command: find a better alternative than `force`. Can you create an issue about Viper and `force` before getting into it?

Awaiting for feedback.

eine:
@hacdias I've been rebasing on top of your work. Can I force-push this branch before I focus on your last comment? Or are you actually working on top of this?

hacdias:
Yes, sure. Not working here anymore

hacdias:
But I'd recommend you to read my comment

hacdias:
Please check #613

hacdias:
@1138-4EB please read my last comment and check #613

eine:
> @1138-4EB the idea I had doesn't work unfortunately. You know what you feel after making a lot of changes and in the end it doesn't work as expected 😭

I know...
 
> One last thing about this PR: I think we should only use viper for the root command. The others don't make sense to make use of it.

I limited it to `root` and `config`. All other sources rely on `settings.RuntimeCfg`. The procedure is as follows:

- In `initConfig`, `settings.RuntimeCfg` is filled with `v.GetString`.
- If `quickSetup` is executed, `settings.RuntimeCfg` is compared with `settings.RuntimeDefaults` to get a map of non-default parameters.
  - The resulting map is saved in the DB, in a new field named `Runtime` that replaces previous BaseURL, Key and Scope/Root.
- When the server is started, if `--force` is unset, some of the fields in `settings.RuntimeCfg` are update with `Runtime` from the DB.

Note that `force=true` is not implemented ATM. So, we can remove the flag.

> Yes, we should allow users to import config and users. So I believe we should create the following two commands:
>
> Same applies for export.

I'll check the corresponding PR.

> And just read the files. Maybe see if viper exports their parser (or see which parser they use) because it's canse insensitive and it would be helpful for users. And only use Viper with the root command.

I think we can use parsers directly for import/export functions. Viper is meant to handle configuration parameters from multiple sources, not for complex structures. I'll comment further in the PR.

> For root command: find a better alternative than `force`. Can you create an issue about Viper and `force` before getting into it?

Please, check the comment above.

eine:
Also, note that this PR is not ready to merge yet. ` settings/storage.go` and `storage/bolt/importer/conf.go` need a review.

hacdias:
@1138-4EB still I don't think we need Viper for config. If we use `config import` there's no need for viper on config set/init. `config import` is capable of initing or replacing a database depending on its existence.

eine:
> @1138-4EB still I don't think we need Viper for config. If we use `config import` there's no need for viper on config set/init. `config import` is capable of initing or replacing a database depending on its existence.

It is removed now.

cmd/config_init.go
```diff
@@ -28,7 +28,7 @@ override the options.`,
 		authMethod, auther := getAuthentication(cmd)
 
 		s := &settings.Settings{
-			Key:        generateRandomBytes(64), // 256 bit
```
hacdias, line 31: Well... This mustn't be removed because Settings.Key **is not** TLS Key. Settings.Key is the secret for JWT tokens. I'm making an update or two on master to help with this PR.

hacdias: Please see #615. It will help this PR and a struct is better than an arbitrary map.

hacdias: Found a way not to use force and still use the server settings from DB merged with provided

hacdias:
So I came with an alternative that doesn't need `force`: #616

PR #612, opened 2019-01-07
feat: wrap commands to send info

hacdias:
Wrap commands in a better way to pass storage

### DIFF
--- a/cmd/cmds_add.go
+++ b/cmd/cmds_add.go
@@ -3,7 +3,6 @@ package cmd
 import (
 	"strings"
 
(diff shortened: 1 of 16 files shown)


eine:
Just because of curiosity, does this python thing have anything to do with the programming language?

hacdias:
No, it was more of a Cobra, Viper and... a Python thingy 😆 The 🐍 emoji didn't work.

eine:
It was driving me mad, trying to find any reference about golang and python on the net 😆.

PR #613, opened 2019-01-07
[wip] feat: config/users import/export

hacdias:
~`users import` is not yet finished~

### DIFF
--- a/auth/json.go
+++ b/auth/json.go
@@ -22,7 +22,7 @@ type jsonCred struct {
 
 // JSONAuth is a json implementaion of an Auther.
 type JSONAuth struct {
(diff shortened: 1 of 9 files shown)


hacdias:
@1138-4EB I think this one is ready for review!

eine (approved):
Overall, it looks good to me.

As an enhancement, I think we should consider accepting YAML for `users import`, because anchors can really help introduce long lists. See an example in go-yaml/yaml#394. Note that the issue is about retrieving the list of anchors, which would be useful to define user groups. But we can ignore that, and just use the expanded list.

Anyway, you can merge this request and create an issue to consider YAML in the future.

PR #616, opened 2019-01-08
fix: use server options from DB too

hacdias:
This is an alternative for #611 and closes #611. With this PR, the server options (tls key, tls cert, root, address, port, log and baseurl) come from the following precedence:

- Flags
- Env
- Config File
- Database

We manually check for the flags value as Viper `IsSet` has a bug, returning always true if binded with a flag so I decided to use Viper **without** default, only for **env** and **config file**.

If a flag is set, use it. Otherwise check if set with Viper (env/config), if so use it. Otherwise fallback to the values on DB. Better explained on code.

Also, `config init/export/import/set` now accept this server options too.


hacdias:
Also, I think this should be the last feature PR before 2.0. The PRs from now on should be cleaning up code and fixing bugs only if we found any.

hacdias:
@1138-4EB I'm merging this now but I still ask you to take a look at the code afterwards to see if you spot anything wrong. In the meanwhile, this will close #611. If you had something in there that isn't covered by this, please make a new PR with the changes.

PR #645, opened 2019-01-26
fixes: correct users importing id

### LINKED ISSUE #639: DB insertion starts from ID 1 after migration
**Description**
After I migrated to v2, I tried creating a new user. This created a new user with id 1, which overwrote my admin user. 😢 

**Expected behaviour**
ID continues from where it was up to.

ID starts from 1

**Additional context**
I used to use \*cough\* filemanager \*cough\* within Caddy. After migration I now use filebrowser as a standalone docker container and proxy requests to it via caddy.

**How to reproduce?**
Should be clear with the above

**Files**
N/A

hacdias:
@princemaple could you try out this branch to see if it works on your end?

### DIFF
--- a/cmd/users_import.go
+++ b/cmd/users_import.go
@@ -70,6 +70,10 @@ list or set it to 0.`,
 						checkErr(usernameConflictError(user.Username, conflictuous.ID, user.ID))
 					}
 				}
+			} else {
+				// If it doesn't exist, set the ID to 0 to automatically get a new
+				// one that make sense in this DB.
+				user.ID = 0
 			}
 
 			err = d.store.Users.Save(user)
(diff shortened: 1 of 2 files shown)


princemaple:
Hey @hacdias , unfortunately I'm not a Go developer. I don't have a Go setup to test this. I thought I'd be able to compile a new one with the Dockerfile, then the Dockerfile doesn't really contain how to build it either. The code looks good though. (Go is quite easy to read, :) would be great if there was a dockerfile that actually builds FB binary.

hacdias:
@princemaple oh, okay. That's a good idea: to have a Docker image to build a specific commit. I'm merging this and as soon as it finished building, it will be available as the latest on Docker Hub so you can try.

If it works well for you, please close #639. Otherwise let me know!

princemaple:
sure thing

PR #650, opened 2019-01-30
Create directory of database if it does not exist

eine:
If the directory where the database is to be created does not exist, `storm.Open(path)` fails:

``` bash
# ./filebrowser -d /nonexist/db.db
2019/01/30 21:07:57 open /nonexist/db.db: no such file or directory
```

This PR adds the directory with `MkdirAll`, which is functionally equivalent to `mkdir -p`.


hacdias:
Seems good!

Issue #663, opened 2019-02-02
Cannot build from scratch


sibprogrammer:
**Description**
Attempt to build from scratch has failed due to problems with dependencies.

**Expected behaviour**
Successful generation of the binary.

**What is happening instead?**
Bunch of errors:

```
 ❯ go build                                                                                                                                                               [12:17:54]
# github.com/mholt/archiver
../../go/pkg/mod/github.com/mholt/archiver@v3.1.0+incompatible/lz4.go:18:7: undefined: lz4.NewWriter
../../go/pkg/mod/github.com/mholt/archiver@v3.1.0+incompatible/tarlz4.go:84:12: undefined: lz4.Writer
../../go/pkg/mod/github.com/mholt/archiver@v3.1.0+incompatible/tarlz4.go:86:10: undefined: lz4.NewWriter
# golang.org/x/sys/unix
../../go/pkg/mod/golang.org/x/sys@v0.0.0-20181205085412-a5c9d58dba9a/unix/fcntl.go:26:42: undefined: Flock_t
../../go/pkg/mod/golang.org/x/sys@v0.0.0-20181205085412-a5c9d58dba9a/unix/ioctl.go:14:47: undefined: Winsize
../../go/pkg/mod/golang.org/x/sys@v0.0.0-20181205085412-a5c9d58dba9a/unix/ioctl.go:25:47: undefined: Termios
../../go/pkg/mod/golang.org/x/sys@v0.0.0-20181205085412-a5c9d58dba9a/unix/sockcmsg_unix.go:37:18: undefined: Cmsghdr
../../go/pkg/mod/golang.org/x/sys@v0.0.0-20181205085412-a5c9d58dba9a/unix/sockcmsg_unix.go:43:9: undefined: Cmsghdr
```

**How to reproduce?**
* checkout master branch (checked on 4a2573830ea76246c13d16fda1ad75fcd133996c)
* prepare all dependencies (`go mod download`, frontend preparation)
* run `go build`

eine:
Did you try executing `./wizard.sh -b` from the root of this repo? Optionally, `USE_DOCKER="true" ./wizard.sh -b` will execute everything in a docker container.

- https://github.com/filebrowser/filebrowser/blob/master/wizard.sh#L176-L212
- https://github.com/filebrowser/filebrowser/blob/master/wizard.sh#L79-L93
- https://github.com/filebrowser/filebrowser/blob/master/wizard.sh#L103-L110

[...]

PR #668, opened 2019-02-06
Fixed dependencies

kolaente:
**Description**
This PR fixes the dependencies. I tried to build the project, which failed. After some digging, I found out `github.com/mholt/archiver` had a broken dependency, `github.com/pierrec/lz4`. The problem is `github.com/mholt/archiver` does not use go modules, so we need to keep track of its dependencies in our `go.mod`. I've updated the version of both `github.com/mholt/archiver` and `github.com/pierrec/lz4` to fix this.

I've also added `.idea` to the gitignore.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@
 *.bak
 _old
 rice-box.go
+.idea/
--- a/go.mod
+++ b/go.mod
@@ -21,12 +21,12 @@ require (
 	github.com/kardianos/osext v0.0.0-20170510131534-ae77be60afb1 // indirect
 	github.com/kr/pretty v0.1.0 // indirect
 	github.com/maruel/natural v0.0.0-20180416170133-dbcb3e2e8cf1
-	github.com/mholt/archiver v3.1.0+incompatible
+	github.com/mholt/archiver v3.1.1+incompatible
 	github.com/mholt/caddy v0.11.1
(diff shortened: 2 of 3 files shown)


eine:
The changes in the dependencies are quite similar to https://github.com/filebrowser/filebrowser/pull/664/files. In [#663](https://github.com/filebrowser/filebrowser/issues/663#issuecomment-459952410), we saw that there is no problem in a empty/clean go environment (that's why Travis CI does not fail). However, `go mod` does not behave as expected when some of the packages are already available on the host.

> so we need to keep track of its dependencies in our `go.mod`. I've updated the version of both `github.com/mholt/archiver` and `github.com/pierrec/lz4` to fix this.

Could you, please, elaborate on this? How did you update the version? Manually or with some go command? Shouldn't it be updated automatically by `go get`?

I'm ok with merging this. I just want to understand the procedure and the motivation behind it, so that we can fix it when it happens again.

kolaente:
@1138-4EB I did `go get -u github.com/mholt/archiver` and `go get github.com/pierrec/lz4@master`.

AFAIK `go get` doesn't update anything unless you tell it to do so. Otherwise it just downloads the dependencies specified in `go.mod` and `go.sum`.

eine:
Thanks @kolaente!

Issue #657, opened 2019-02-01
Per-user auto directory creation


Casxt:
**Is your feature request related to a problem? Please describe.**
Auto create folder for new user are convenient for a small team.

**Describe the solution you'd like**
for example I can setting user default scope like `./User/{$USERNAME}`,  and I notice filebrowser will create folder if user scope are not exists, so when a new user 'A' sign up, his own floder will be create automaticlly.

**Describe alternatives you've considered**
support a few environment variable are enough make scope system more powerful.
add a Command Runner of user sign up event is also a good solution.

ttys3:
@Casxt  maybe this is what you need: http://files.80x86.io/router/rom/opt/fb/
I've added an "Auto create user home dir"  option in global settings.

hacdias:
@442hz it would be nice if you could create a PR into this repository with that feature 😄

ttys3:
@hacdias  OK,I've made some PR to this feature:
https://github.com/filebrowser/filebrowser/pull/676
https://github.com/filebrowser/frontend/pull/102

hacdias:
Both PRs were merged! Closing this issue and releasing a new version (2.0.4). Sorry for taking this long and really thanks @ttys3

laker-93:
how can this feature be used to create private folders when new users sign up?
If I have this feature enabled then user data is not private to each user.

laker-93:
I've figured out how to do this. In the admin panel, must have the following option enabled

Auto create user home dir while adding new user 

Then in the "User default settings" section must leave the 'scope' as empty.

PR #676, opened 2019-02-15
feat: per-user auto directory creation

ttys3:
Per-user auto directory creation 

**Description**
as #657  mentioned  
auto create user directory while creating new user.

**Limitation**
1. this logic **DOES NOT** work when edit an exists user.
2. this logic **DOES NOT** work when public user register on frontend page.
3. this logic **ONLY WORK** when admin is creating a new user on the user management page.

there are 2 conditions we need to consider
 when creating a new user (with auto dir creation enabled):
1.  user scope == global default user scope
2. user scope != global default user scope

### condition 1.1
for example: 
add new user `foo` and leave the scope dir to the default  user scope  `.` ,  
and `auto directory creation ` option was enabled, 

suggest if we have the following startup command:
```
/path-to/filebrowser_amd64 -r /media
```

then the user's default scope dir will be:
```
./users/foo
```

and the real path on the system will be:
```
/media/users/foo
```

### condition 1.2
another condition is the default user scope != server root path, but user scope = default user scope
for example:
we set the `default user scope` to `./bar` under `Global settings`
when we add a user `foo` and `auto directory creation ` option was enabled,
the user's default scope dir will be:
```
./bar/users/foo
```

and the real path on the system will be:
```
/media/bar/users/foo
```

### condition 2
for example: 
add new user `foo` and set  the scope dir  to `./special` ,  and leave global settings  `default  user scope` still   `.`
we do create the dir this time , but do not follow the default path creation logic.
the user's default scope dir will be:
```
./special
```

and the real path on the system will be:
```
/media/special
```

image

frontend assets PR is at  https://github.com/filebrowser/frontend/pull/102

### DIFF
--- a/cmd/config.go
+++ b/cmd/config.go
@@ -89,6 +89,7 @@ func printSettings(ser *settings.Server, set *settings.Settings, auther auth.Aut
 	w := tabwriter.NewWriter(os.Stdout, 0, 0, 2, ' ', 0)
 
 	fmt.Fprintf(w, "Sign up:\t%t\n", set.Signup)
+	fmt.Fprintf(w, "Create User Dir:\t%t\n", set.CreateUserDir)
 	fmt.Fprintf(w, "Auth method:\t%s\n", set.AuthMethod)
 	fmt.Fprintf(w, "Shell:\t%s\t\n", strings.Join(set.Shell, " "))
 	fmt.Fprintln(w, "\nBranding:")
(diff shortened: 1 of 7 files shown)


hacdias:
Please also add support for this option on `config` related commands and `users new`.

ttys3:
`config` related commands and `users new` support has been added.

hacdias (approved):
LGTM! Thanks!

Issue #681, opened 2019-02-20
Some txt file can be preview and edit, but some can not.


chinanoahli:
**Description**
Some txt file can be preview and edit, but some can not.

**Expected behaviour**
All txt file can be preview? Otherwise why preview and edit function?

**What is happening instead?**
Some txt file can be preview and edit, but some can not.
Like I have a file.txt in ./files, it just can be download, it can not being preview and edit.
But test.txt in ./files/test can being preview and edit, and of course download too.

**Additional context**
I am using filebrowser in ubuntu 18.04 lts x86_64, launch it by a custom server:
```
#/etc/systemd/system/filebrowser.service

[Unit]
Description=File Browser v2 Service
After=network.target
Wants=network.target

[Service]
Type=simple
PIDFile=/var/run/filebrowser.pid
ExecStart=/usr/local/bin/filebrowser -c=/home/filebrowser/config.json -d=/home/filebrowser/filebrowser.db -r=/home/filebrowser
RemainAfterExit=no
Restart=on-failure
RestartPreventExitStatus=23

[Install]
WantedBy=multi-user.target
```
file /home/filebrowser/config.json (seem does not have effect at all):
```
{
  "settings": {
    "signup": false,
    "defaults": {
      "scope": "./files",
      "locale": "zh-tw",
      "viewMode": "list",
      "sorting": {
        "by": "name",
        "asc": false
      },
      "perm": {
        "admin": false,
        "execute": false,
        "create": false,
        "rename": false,
        "modify": true,
        "delete": false,
        "share": false,
        "download": true
      },
      "commands": []
    },
    "authMethod": "json",
    "branding": {
      "name": "Magical Space J1",
      "disableExternal": false,
      "files": "/home/filebrowser"
    },
    "commands": {
      "after_copy": [],
      "after_delete": [],
      "after_rename": [],
      "after_save": [],
      "after_upload": [],
      "before_copy": [],
      "before_delete": [],
      "before_rename": [],
      "before_save": [],
      "before_upload": []
    },
    "shell": [],
    "rules": []
  },
  "server": {
    "root": ".",
    "baseURL": "",
    "tlsKey": "",
    "tlsCert": "",
    "port": "8080",
    "address": "127.0.0.1",
    "log": "stdout"
  },
  "auther": {
    "recaptcha": null
  }
}
```

**How to reproduce?**
I don't know yet. There's no specific pattern.

eine:
[...]

PR #721, opened 2019-05-07
fixup func isBinary to handle CJK runes correctly

ttys3:
**Description**
related issue: https://github.com/filebrowser/filebrowser/issues/681


hacdias:
Thanks for this!

Issue #715, opened 2019-04-20
Need to set --auth.method to update auth settings


hacdias:
**Description**

Trying to change the authentication settings.

**Expected behaviour**

I'd run:

```
filebrowser config set --recaptcha.key=key --recaptcha.secret=secret
```

And the configurations (if using the JSON Auth Method) would update correctly.

**What is happening instead?**

The configurations aren't changing.

**Additional context**

We need to set `--auth.method` every time we want to change the authentication settings. This applies to all types of authentication and not only `json`. Also, by setting this flag, we override the current auth configurations which might not be useful in some cases.

This will take some rework on the `config set` code to make this work. I'll update the documentation in the meanwhile so people know. If someone wants to take a stab at this bug, we'll all be greateful ❤️ 

**How to reproduce?**

Run the command above.

**Files**

N/A

chummumm:
can I set recaptcha in config.json?

hacdias:
@chummumm not right now

Soneji:
@hacdias is there any update on whether we can set recaptcha in the config.json?
Thanks

PR #732, opened 2019-05-19
Make --auth.method parameter optional when changing auth parameters Fixes: #715

unicod3:
**Description**
If user does not specify auth.method but gives authentication specific parameters, it will read the default auth.method then will update the related authentication parameters if that parameters belong to default auth.method.

Forexample:
if user gives 
`filebrowser config set --recaptcha.key=key --recaptcha.secret=secret`
and the default auth.method is 'json' it will only update the values but if the default value is 'proxy' it will panic and wants you to pass proxy related parameters. 

This logic is also same for proxy auth method.

### DIFF
--- a/cmd/config.go
+++ b/cmd/config.go
@@ -72,7 +72,7 @@ func getAuthentication(flags *pflag.FlagSet, defaults ...interface{}) (settings.
 		}
 
 		if header == "" {
-			panic(nerrors.New("you must set the flag 'auth.header' for method 'proxy'"))
+			checkErr(nerrors.New("you must set the flag 'auth.header' for method 'proxy'"))
 		}
 
 		auther = &auth.ProxyAuth{Header: header}
@@ -99,7 +99,7 @@ func getAuthentication(flags *pflag.FlagSet, defaults ...interface{}) (settings.
 		}
 
 		if key == "" || secret == "" {
-			panic(nerrors.New("you must set the flag 'recaptcha.key' and 'recaptcha.secret' for method 'json'"))
+			checkErr(nerrors.New("you must set the flag 'recaptcha.key' and 'recaptcha.secret' for method 'json'"))
 		}
 
 		jsonAuth.ReCaptcha = &auth.ReCaptcha{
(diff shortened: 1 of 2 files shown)


unicod3:
@hacdias  Should I need to fix code climate issue?

cmd/config.go
```diff
@@ -62,17 +78,28 @@ func getAuthentication(flags *pflag.FlagSet) (settings.AuthMethod, auth.Auther)
...
+			key = kmap["key"].(string)
+		}
+
+		if secret == "" {
+			smap := defaultAuther["recaptcha"].(map[string]interface{})
+			secret = smap["secret"].(string)
+		}
+
+		if key == "" || secret == "" {
+			panic(nerrors.New("you must set the flag 'recaptcha.key' and 'recaptcha.secret' for method 'json'"))
```
hacdias: Please use `checkErr`

cmd/config_set.go
```diff
@@ -71,6 +71,10 @@ you want to change. Other options will remain unchanged.`,
 		} else {
 			auther, err = d.store.Auth.Get(set.AuthMethod)
 			checkErr(err)
+			// check if there are new flags for existing auth method
+			set.AuthMethod, auther = getAuthentication(flags, set, auther)
```
hacdias: Perhaps we could simplify this and merge with line 68, passing `nil` if `hasAuth == false`.

hacdias (changes requested):
Thanks for this!!! It's an awesome PR. Could you just take a look at my comments? And no, don't worry about CodeClimate failure (it's just 2 lines).

cmd/config_set.go
unicod3: this is done

cmd/config.go
unicod3: I did not get how to use checkErr in this block though.

hacdias: Just replace `panic` by `checkErr`

unicod3: done

hacdias:
Thanks, merging!

PR #771, opened 2019-06-06
use ParseFromRequest instead of ParseFromRequestWithClaims

dom3k:
Use `ParseFromRequest` instead of `ParseFromRequestWithClaims` because `ParseFromRequestWithClaims` function is `DEPRECATED`.

Godoc: [https://godoc.org/github.com/dgrijalva/jwt-go/request#ParseFromRequestWithClaims](https://godoc.org/github.com/dgrijalva/jwt-go/request#ParseFromRequestWithClaims)
Source code [https://github.com/dgrijalva/jwt-go/blob/master/request/request.go#L43](https://github.com/dgrijalva/jwt-go/blob/master/request/request.go#L43)

### DIFF
--- a/http/auth.go
+++ b/http/auth.go
@@ -54,7 +54,7 @@ func withUser(fn handleFunc) handleFunc {
 		}
 
 		var tk authToken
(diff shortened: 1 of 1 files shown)


hacdias:
Thanks!

PR #791, opened 2019-06-23
prevent maliciously constructed parameters caused panic

ttys3:
prevent maliciously constructed parameters like `/api/public/dl/XZzCDnK2_not_exists_hash_name` cause panic

 this bug was introduced in https://github.com/filebrowser/filebrowser/pull/727


hacdias:
Thanks for the good work, once again @ttys3!

Issue #787, opened 2019-06-19
symbol '#' will truncate char.


AcidGo:
When uploading a file, the symbol of the file name is #, and subsequent characters will be truncated.
eg.
MyTest#File --> MyTest

hacdias:
Closed by #792

Issue #761, opened 2019-05-25
Cannot upload files with '#' on their name


ntgeralt:
hi，maybe you can test it.
file name with # ahead.

hacdias:
Closed by #792

PR #792, opened 2019-06-23
file upload: safe encode filename with encodeRFC5987ValueChars()

ttys3:
this should fixup 
https://github.com/filebrowser/filebrowser/issues/787

https://github.com/filebrowser/filebrowser/issues/761


hacdias:
hey @ttys3! It looks good! thanks for this

Issue #532, opened 2018-09-08
🆘 Need Maintainers


hacdias:
This project is not under active development, i.e., new features won't be created and deep bugs might be hard to be fixed. A small set of developers is taking care of keeping it building and fixing small issues.

If you're interested in becoming a maintainer of this project, please let us know through this issue and start contributing. It would be awesome for the more than 3000 people that starred this project!

yurividal:
@hacdias that is SO SAD... =(

I wish i had any knowledge in this programing field, to help... 

Anyways, thank you for all the hard-work

hacdias:
@yurividal 😞

rhjmoore:
Do you need help maintaining the Go side or the web client side more?

hacdias:
@rhjmoore both. All help is welcome 😄

jazware:
Hey, I'll see if I can carve up some time to help out. I had a few commits way back when this was starting out but haven't touched the source in a while.

jazware:
If you want to make me a maintainer I can dig through the issues and get rid of the noise of people who don't understand how web or golang or whatever works...

hacdias:
@ericvolp12 awesome that you want to contribute! You can start by looking at the issues and creating PRs. We'll only give push access to someone who has a history on this repository. You can start building one right now! 😃

jazware:
Hey @hacdias it'd be nice if I had the permissions required to close or label issues as I dig through the backlog. I do have a history of commits to the project for the record.

eine:
> it'd be nice if I had the permissions required to close or label issues as I dig through the backlog

You can comment on those you are interested to fix, and we will close or relabel them correspondingly. Indeed, we rarely take actions without double approval. That is, even if an issue is to be closed, we comment there to ensure that it is solved. Then, if no reply is received in around a week, it is closed. However, if the issue is old, and no maintainer replied to it previously, we wait longer as the OP might have forgotten about it and some more time might be needed to come back.

Nonetheless, ATM we can read and respond to issues/questions, but *new features won't be created and deep bugs might be hard to be fixed*. So, it'd be great if you contributed to these topics.

dtwilliams10:
I would like to help, I haven't developed with Go or Vue but I do have full stack experience that I can try to help with. After looking through the issue log is there any priority or order that you would want things worked on?

hacdias:
@dtwilliams10 hello! Thanks 😄 hope you contribute. There's no special order. Just bugs >>>> features.

[...]

PR #798, opened 2019-07-01
Proposal for public folders

alexandrestein:
**Description**
I try to make it easy to share folders.
I think folder should be taken as is and not as a big **zip**.
My goal is to use the **Listing** component which I feed with a **GET** request to `/api/public/resources/`.

I don't see how you fill-up the **Listing** component.
So at this point this is not a working solution.
I have the folder content from the server but I'm not able to show it. 

I would be happy to have some help from you on the **Vue** side.
To fill-up the component and be able to explore it from the public area using the hash.

What do you think?

### DIFF
--- a/frontend/src/views/Share.vue
+++ b/frontend/src/views/Share.vue
@@ -1,6 +1,13 @@
 <template>
-  <div class="share" v-if="loaded">
-    <a target="_blank" :href="link" v-if="!file.isDir">
+  <div
+    class="share"
+    v-if="loaded"
+  >
+    <a
+      target="_blank"
+      :href="link"
+      v-if="!file.isDir"
+    >
       <div class="share__box">
         <div class="share__box__download">{{ $t('download.downloadFile') }}</div>
         <div class="share__box__info">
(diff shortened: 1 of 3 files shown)


ttys3:
good job!
by the way, I think we'd better use Preview when double click on a single file.

alexandrestein:
You would like to use the Preview component instead of the actual link `<a>` in `Share.vue`, I'm right?

I think this would be great, but in an other PR or after this one works.
For now I just want to give access to directories without the need of downloading the all tree of the folder.

Does some one can tell me how to fill-up `listing` component?
Or you can update the code to make it works. :smile:

ttys3:
@alexandrestein  yes.
we need add some `computed` method like:
`isDir` 
`isFile`

when it is a dir, we let the `Listing` component to render
and if it is a single file, , we let the `Preview` component to render

```vue
<template>
  <div class="share" v-if="loaded">
        <listing v-if="isDir" ></listing>
	<preview v-if="isFile" ></preview>
  </div
</template>
```

since the original `Preview.vue` depends on some `state` ,  we probably can not use the  `Preview` directly, this may need some tricks

alexandrestein:
Thank you @ttys3 but it's not my actual concern.
The preview can be great but display folder is what I want right now.

I was hoping that maintainers or some one who knows the package well enough could finish this "_partly_" done stuff.

I'm sure there is not too much to do to make it works.

hacdias:
I'm closing this since this repository will be archived. Please read #532 for more information. Thanks for all your contributions!

PR #879, opened 2019-10-11
[Frontend] Fix invalid start_url in manifest file.

DanielVladu:
As far as I know the invalid start_url format breaks only the "Add to Home Screen" functionality, so it's a really minor change, fixing a nice functionality.


hacdias:
Thanks!

Issue #498, opened 2018-08-07
filebrowser/dev docker image


hacdias:
@1138-4EB I see you created a [chore/devenv](https://github.com/filebrowser/filebrowser/tree/chore/devenv) branch to automatically build the filebrowser/dev image. I suggest using a separate repository. Not sure if you have permission to create repositories, but I'll create one and then you can move the stuff there.

hacdias:
Here it is: https://github.com/filebrowser/docker-dev

PR #885, opened 2019-10-17
Use Docker to build app & docker-compose

josepaiva94:
**Description**
I did not want to install Go and all dependencies necessary to build the app on my machine, so I've made some tweaks to the Dockerfile to build it. I have also added docker-compose as I will use in my next developments.

If you think this is helpful, feel free to use.


hacdias:
@1138-4EB since you are more acquaintanced with Docker, could you take a look at this one please?

eine:
@hacdias, resources for the use case that @josepaiva94 is trying to support were already pushed to this repo in August 2018. However, the content was removed without prior notice or discussion in May 2019:

- https://github.com/filebrowser/filebrowser/issues/498
- https://github.com/filebrowser/docker-dev
- https://hub.docker.com/r/filebrowser/dev/
- https://github.com/filebrowser/filebrowser/blob/d47774c828394dd234fdf57ab2d38e103e51cdfd/wizard.sh#L198-L206
- https://github.com/filebrowser/filebrowser/commit/cc6689ac3a112227689fd97f2385854a0d39b12d#diff-c936abc8f548b944159df70e8a3f0fff

Moreover, I think that the implementation in this PR is not desirable:

- Snippets that are already defined in `wizard.sh` are redefined in the Dockerfile. This increases the maintenance burden, and it might lead to inconsistencies in the procedure between users building the tool on the host and those using the dockerfile.
- No default support is provided for users that want to build (and test) the tool on the host, before adding it to an alpine docker image. They are forced to build the binary again during the docker image build process. This conflicts with how goreleaser is expected to be used.
- The definition of the steps in the dockerfile is suboptimal. Many `RUN` instructions can be merged, in order to reduce the number of layers in the image.
  - `node_modules` from the host is copied in the build image.
- I don't know if it's worth using `node:latest`, instead of building both the frontend and the backend in a `golang:alpine` image.
  - Independently of using one or two *build* images, I think that these should be cached (pushed to hub.docker.com), to avoid installing build dependencies each time.
- The compose file is an interesting/useful contribution. However, I would expect users to run a compose file based on an image and not a Dockerfile. Furthermore, I'd expect the compose file to be used independently of the sources of the repo. Therefore, I think that the addition of `.docker/sysroot/*` to `.gitignore` does not fit. Should it be kept, I'd use `.docker` instead.

Summarizing, IMHO @josepaiva94's contributions should focus on either using filebrowser/dev again, or on updating it; but these additions do not fit in `./Dockerfile` because it introduces a non-backwards workflow.

josepaiva94:
@1138-4EB and @hacdias Agreed with your input, except that I would keep `node:latest` instead of adding node to `golang:alpine`. I can fix the mentioned points, if you want.

> No default support is provided for users that want to build (and test) the tool on the host, before adding it to an alpine docker image. They are forced to build the binary again during the docker image build process. This conflicts with how goreleaser is expected to be used.

To solve this, we may keep 2 Dockerfiles. What do you think?

Anyway, this was just the first step in my roadmap. Will try to contribute with features that I need and are compatible with yours.

hacdias:
Hello @josepaiva94! Sorry for only getting back to you now. Could you revert the updates to package(-lock).json since those do not fit into the scope of this PR? Also, could you try out with two Dockerfiles to show me better your idea?

josepaiva94:
Hello @hacdias , I have made the two Dockerfiles and the changes according to @1138-4EB comments. I've also replaced my package(-lock).json with yours.

PR #889, opened 2019-10-17
Generate manifest.json dynamically

DanielVladu:
Generate manifest.json dynamically in index.html so that we can take advantage of the program variables like custom name, base url and static url.

Using this the Add to Home screen functionality should work in standalone display mode in Chrome and Android phones even if a custom base url is set and a reverse proxy server is used.


hacdias (approved):
LGTM! Thanks for this!

hacdias:
Why did you close this?

DanielVladu:
The first commit had some issues, closed the pull request to make sure it does not get merged until I fix them. Now everything should be ok :)

hacdias:
@kriminalrat cool! Thanks. I see the fix you did. Makes sense :)

PR #896, opened 2019-10-22
delete keyup only if no active prompt #895

### LINKED ISSUE #895: Renaming prompt DELETE key wrong event
**Description**
In file renaming prompt, when I press DELETE key, unexpected things happen.

**Expected behaviour**
When I press `DELETE` key in renaming, not file will be deleted but only one character.

When I press `DELETE`, it cancels my renaming dialog and asks me whether I want to remove that file.

**How to reproduce?**
Try to rename file and delete one character using `DELETE` key.

m1k1o:
**Description**
Fixes #895. When there are active prompts, and someone pushes `DELETE` key it should not remove file (closing existing prompt and opening new one). Especially when people are renaming file and they press delete.

### DIFF
--- a/frontend/src/views/Files.vue
+++ b/frontend/src/views/Files.vue
@@ -176,7 +176,8 @@ export default {
           !this.isFiles ||
           this.loading ||
           !this.user.perm.delete ||
-          (this.isListing && this.selectedCount === 0)) return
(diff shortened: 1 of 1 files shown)


hacdias:
Thanks for this fix!

PR #906, opened 2019-10-30
Add (soft) wrap to the editor

ghost:
**Description**
It wraps lines which are longer than the viewport. It's just `wrap: true` added to the editor in the frontend. Any reasons this is not the default?

**Before**
20191030-125959-1520x471+390+196

**After**
20191030-130150-1521x466+389+200


ghost:
Force pushed the same but signed with my pgp key.

PR #918, opened 2019-11-17
Determine the real IP address of the client for logging

spacebat:
When running behind a reverse proxy such as nginx, the remote IP as
logged is always that of the proxy. Figuring out the correct address
in this context is a little tricky, hence the following module is
used:

https://github.com/tomasen/realip

Because of the use of a simple module with no further dependencies, only a minor change is needed in `http/data.go`. I'm new to Go, so maybe I'm not doing the canonical thing to handle errors from the new function.

I can't find any issues that this closes, I didn't create an issue because it was easy to fix. I'm currently running this version and it works fine for me.

One side-effect is that the port of the client is not logged now, even when a reverse proxy is not in use. I don't think that information is very useful though - most network services I've seen don't log it.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -25,6 +25,7 @@ require (
 	github.com/spf13/jwalterweatherman v1.1.0 // indirect
 	github.com/spf13/pflag v1.0.5
 	github.com/spf13/viper v1.5.0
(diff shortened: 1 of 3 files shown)


spacebat:
I ran into that same build error as circleci the first time I built. It seems unrelated to my changes. Building a second time it did not occur so I guess it's random.

PR #920, opened 2019-11-19
Refactor prompts

freedomlang:
Use `component` instead of `v-if` and `v-else-if`


hacdias:
Thanks for this simplification!

PR #979, opened 2020-06-06
Picture Thumbnails support

monkeyWie:
**Description**
Add picture thumbnails support,use [https://github.com/nfnt/resize](https://github.com/nfnt/resize) lib.
effect:
image
image

:heart: Thank you!


nambrosch:
@monkeyWie I noticed that nfnt/resize isn't updated anymore, there are some suggestions about alternatives here, have you checked them out?  https://github.com/nfnt/resize/issues/63

monkeyWie:
@nambrosch oh,I'm not checked,I just saw [nfnt/resize#63](https://github.com/nfnt/resize/issues/63),maybe I can use `golang.org/x/image/draw` lib.

o1egl:
A few years ago, I was working on an image processing service that was on the hot path. The winner among all tested libs was opencv compiled with libjpeg-turbo library. I don't think that performance will be an issue in that type of project. Therefore, I propose to use one of these libraries:

- https://github.com/muesli/smartcrop
- https://github.com/disintegration/imaging

http/preview.go
```diff
@@ -0,0 +1,126 @@
...
+		width = maxSize
+		height = 0
+	} else if height > maxSize && height > width {
+		width = 0
+		height = maxSize
+	} else {
+		width = 0
+		height = 0
+	}
+	m := resize.Resize(uint(width), uint(height), img, resize.Lanczos3)
```
o1egl, line 124: Please use NearestNeighbor for faster processing.

http/preview.go
```diff
@@ -0,0 +1,126 @@
...
+	default:
+		return nil, errors.New("extension is not supported")
+	}
+	if err != nil {
+		return nil, err
+	}
+	return buf, nil
+}
+
+const maxSize = 1080
```
o1egl, line 104: I don't think that 1080px is a good size for thumbnail

http/preview.go
```diff
@@ -0,0 +1,126 @@
...
+
+func compressImage(decode func(r io.Reader) (image.Image, error), fd io.Reader) (*bytes.Buffer, image.Image, error) {
+	img, err := decode(fd)
+	if err != nil {
+		return nil, nil, err
+	}
+	buf := bytes.NewBuffer([]byte{})
```
o1egl, line 111:
```suggestion
	buf := &bytes.Buffer{}
```

http/preview.go
```diff
@@ -0,0 +1,126 @@
...
+func compressImageHandler(file *files.FileInfo, fd io.Reader) (*bytes.Buffer, error) {
+	var (
+		buf *bytes.Buffer
+		m   image.Image
+		err error
+	)
+
+	switch file.Extension {
+	case ".jpg", ".jpeg":
+		buf, m, err = compressImage(jpeg.Decode, fd)
```
o1egl, line 75: It's not the right way to decode images. Just import decoders and use https://golang.org/pkg/image/#Decode function

http/http.go
```diff
@@ -59,6 +59,7 @@ func NewHandler(store *storage.Storage, server *settings.Server) (http.Handler,
 	api.Handle("/settings", monkey(settingsPutHandler, "")).Methods("PUT")
 
 	api.PathPrefix("/raw").Handler(monkey(rawHandler, "/api/raw")).Methods("GET")
+	api.PathPrefix("/compress").Handler(monkey(compressHandler, "/api/compress")).Methods("GET")
```
o1egl: Let's call it /thumb or /thumbnail

o1egl (changes requested):

http/preview.go
monkeyWie: > A few years ago, I was working on an image processing service that was on the hot path. The winner among all tested libs was opencv compiled with libjpeg-turbo library. I don't think that performance will be an issue in that type of project. Therefore, I propose to use one of these libraries:
> 
> * https://github.com/muesli/smartcrop
> * https://github.com/disintegration/imaging

Thanks for your advice,I haven't done image processing before,so I only found [nfnt/resize](https://github.com/nfnt/resize) from google,next I will try use these libraries.

monkeyWie:
@o1egl Thanks for your advice,I haven't done image processing before,so I only found `nfnt/resize` from google,next I will try use these libraries.

http/preview.go
monkeyWie: ok.

http/preview.go
monkeyWie: Do you have a advice?

o1egl: Let's match the size with the actual thumbnails size on the page.

PR #980, opened 2020-06-08
Support thumbnail image view

monkeyWie:
Reopen PR,now is using `github.com/disintegration/imaging` lib to handle image,and refactor code.
Regarding the frontend code,I run `npm run lint` to fix the errors detected by eslint.

### DIFF
--- a/frontend/src/components/files/Preview.vue
+++ b/frontend/src/components/files/Preview.vue
@@ -86,15 +86,14 @@ export default {
     download () {
       return `${baseURL}/api/raw${this.req.path}?auth=${this.jwt}`
     },
-    thumbnail () {
+    previewUrl () {
       if (this.req.type === 'image') {
         return `${baseURL}/api/preview/big${this.req.path}?auth=${this.jwt}`
-      } else {
-        return `${baseURL}/api/raw${this.req.path}?auth=${this.jwt}`
       }
+      return `${baseURL}/api/raw${this.req.path}?auth=${this.jwt}`
     },
     raw () {
-      return `${this.thumbnail}&inline=true`
+      return `${this.previewUrl}&inline=true`
     }
   },
   async mounted () {
--- a/frontend/src/components/files/Preview.vue
+++ b/frontend/src/components/files/Preview.vue
@@ -86,8 +86,15 @@ export default {
     download () {
       return `${baseURL}/api/raw${this.req.path}?auth=${this.jwt}`
     },
+    thumbnail () {
+      if (this.req.type === 'image') {
+        return `${baseURL}/api/preview/big${this.req.path}?auth=${this.jwt}`
+      } else {
+        return `${baseURL}/api/raw${this.req.path}?auth=${this.jwt}`
+      }
+    },
     raw () {
-      return `${this.download}&inline=true`
+      return `${this.thumbnail}&inline=true`
     }
   },
   async mounted () {
--- a/http/http.go
+++ b/http/http.go
@@ -59,6 +59,7 @@ func NewHandler(store *storage.Storage, server *settings.Server) (http.Handler, 
 	api.Handle("/settings", monkey(settingsPutHandler, "")).Methods("PUT")
 
 	api.PathPrefix("/raw").Handler(monkey(rawHandler, "/api/raw")).Methods("GET")
+	api.PathPrefix("/preview/{size}").Handler(monkey(previewHandler, "/api/preview")).Methods("GET")
 	api.PathPrefix("/command").Handler(monkey(commandsHandler, "/api/command")).Methods("GET")
 	api.PathPrefix("/search").Handler(monkey(searchHandler, "/api/search")).Methods("GET")
 
--- a/http/http.go
+++ b/http/http.go
@@ -59,7 +59,7 @@ func NewHandler(store *storage.Storage, server *settings.Server) (http.Handler, 
 	api.Handle("/settings", monkey(settingsPutHandler, "")).Methods("PUT")
 
 	api.PathPrefix("/raw").Handler(monkey(rawHandler, "/api/raw")).Methods("GET")
-	api.PathPrefix("/preview/{size}").Handler(monkey(previewHandler, "/api/preview")).Methods("GET")
+	api.PathPrefix("/preview/{size}/{path:.*}").Handler(monkey(previewHandler, "/api/preview")).Methods("GET")
 	api.PathPrefix("/command").Handler(monkey(commandsHandler, "/api/command")).Methods("GET")
 	api.PathPrefix("/search").Handler(monkey(searchHandler, "/api/search")).Methods("GET")
 
--- a/http/preview.go
+++ b/http/preview.go
@@ -50,9 +50,13 @@ var previewHandler = withUser(func(w http.ResponseWriter, r *http.Request, d *da
 })
 
 func handleImagePreview(w http.ResponseWriter, r *http.Request, file *files.FileInfo, size string) (int, error) {
-	// Unsupported extensions directly return the raw data
-	if file.Extension == ".ico" || file.Extension == ".svg" {
-		return rawFileHandler(w, r, file)
+	format, err := imaging.FormatFromExtension(file.Extension)
+	if err != nil {
+		// Unsupported extensions directly return the raw data
+		if err == imaging.ErrUnsupportedFormat {
+			return rawFileHandler(w, r, file)
+		}
+		return errToStatus(err), err
 	}
 
 	var imgProcessor imageProcessor
@@ -74,10 +78,7 @@ func handleImagePreview(w http.ResponseWriter, r *http.Request, file *files.File
 		return errToStatus(err), err
 	}
 	defer fd.Close()
-	format, err := imaging.FormatFromExtension(file.Extension)
-	if err != nil {
-		return http.StatusNotImplemented, err
-	}
+
 	img, err := imaging.Decode(fd, imaging.AutoOrientation(true))
 	if err != nil {
 		return errToStatus(err), err
--- a/http/preview.go
+++ b/http/preview.go
@@ -4,7 +4,6 @@ import (
 	"fmt"
 	"image"
 	"net/http"
-	"net/url"
 
 	"github.com/disintegration/imaging"
 	"github.com/gorilla/mux"
@@ -40,12 +39,7 @@ var previewHandler = withUser(func(w http.ResponseWriter, r *http.Request, d *da
 		return errToStatus(err), err
 	}
 
-	if r.URL.Query().Get("inline") == "true" {
-		w.Header().Set("Content-Disposition", "inline")
-	} else {
-		// As per RFC6266 section 4.3
-		w.Header().Set("Content-Disposition", "attachment; filename*=utf-8''"+url.PathEscape(file.Name))
-	}
+	setContentDisposition(w, r, file)
 
 	switch file.Type {
 	case "image":
--- a/http/preview.go
+++ b/http/preview.go
@@ -6,7 +6,6 @@ import (
 	"io"
 	"net/http"
 	"net/url"
-	"strings"
 
 	"github.com/disintegration/imaging"
 	"github.com/gorilla/mux"
@@ -31,11 +30,9 @@ var previewHandler = withUser(func(w http.ResponseWriter, r *http.Request, d *da
 		return http.StatusNotImplemented, nil
 	}
 
-	// Resolve file path from URL
-	path := "/" + strings.Join(strings.Split(r.URL.Path, "/")[2:], "/")
 	file, err := files.NewFileInfo(files.FileOptions{
 		Fs:      d.user.Fs,
-		Path:    path,
+		Path:    "/" + vars["path"],
 		Modify:  d.user.Perm.Modify,
 		Expand:  true,
 		Checker: d,
--- a/http/preview.go
+++ b/http/preview.go
@@ -2,13 +2,14 @@ package http
 
 import (
 	"fmt"
-	"github.com/gorilla/mux"
 	"image"
+	"io"
 	"net/http"
 	"net/url"
 	"strings"
 
 	"github.com/disintegration/imaging"
+	"github.com/gorilla/mux"
 
 	"github.com/filebrowser/filebrowser/v2/files"
 )
@@ -58,7 +59,7 @@ var previewHandler = withUser(func(w http.ResponseWriter, r *http.Request, d *da
 	}
 })
 
-func handleImagePreview(w http.ResponseWriter, file *files.FileInfo, size string) (int, error) {
+func handleImagePreview(w io.Writer, file *files.FileInfo, size string) (int, error) {
 	var imgProcessor imageProcessor
 	switch size {
 	case sizeBig:
--- a/http/preview.go
+++ b/http/preview.go
@@ -3,7 +3,6 @@ package http
 import (
 	"fmt"
 	"image"
-	"io"
 	"net/http"
 	"net/url"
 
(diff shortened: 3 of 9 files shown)


http/thumbnail.go
```diff
@@ -0,0 +1,92 @@
...
+	} else {
+		// As per RFC6266 section 4.3
+		w.Header().Set("Content-Disposition", "attachment; filename*=utf-8''"+url.PathEscape(file.Name))
+	}
+
+	srcImg, err := imaging.Decode(fd, imaging.AutoOrientation(true))
+	if err != nil {
+		return errToStatus(err), err
+	}
+	dstImg := fitResizeImage(srcImg)
```
o1egl: Please use https://pkg.go.dev/github.com/disintegration/imaging?tab=doc#FormatFromExtension with a fixed size

http/thumbnail.go
```diff
@@ -0,0 +1,92 @@
...
+		case ".jpg", ".jpeg":
+			return jpeg.Encode(w, dstImg, nil)
+		case ".png":
+			return png.Encode(w, dstImg)
+		case ".gif":
+			return gif.Encode(w, dstImg, nil)
+		default:
+			return errors.ErrNotExist
+		}
+	}()
```
o1egl: Please use https://pkg.go.dev/github.com/disintegration/imaging?tab=doc#FormatFromExtension and https://pkg.go.dev/github.com/disintegration/imaging?tab=doc#Encode functions

http/thumbnail.go
```diff
@@ -0,0 +1,92 @@
...
+	if err != nil {
+		return errToStatus(err), err
+	}
+	return 0, nil
+}
+
+const maxSize = 1080
```
o1egl: It doesn't make sense to resize an image to 1080px and display it as a thumbnail

o1egl (changes requested):
I tried to run it locally and I see a lot of errors in ui.

http/thumbnail.go
monkeyWie: When resize the image resolution, can effectively reduce the size of the picture, for example reduce a 5MB picture to 500KB,so I choose 1080px as the reduce threshold.

o1egl: If the actual size of the thumbnail is 100x100 for example then why we should return 1080x1080 image for that?

monkeyWie: In fact when the width and height < 1080,will return raw image,as below:
```go
func fitResizeImage(srcImage image.Image) image.Image {
	width := srcImage.Bounds().Dx()
	height := srcImage.Bounds().Dy()
	if width > maxSize && width > height {
		width = maxSize
		height = 0
	} else if height > maxSize && height > width {
		width = 0
		height = maxSize
	} else {
		return srcImage
	}
	return imaging.Resize(srcImage, width, height, imaging.NearestNeighbor)
}
```

o1egl: I still do not understand why we should transfer additional data over the network and resize the image in the browser? Since the thumbnail size is fixed on ui, let's do the same on the backend

monkeyWie: Because the preview image on the UI also uses the thumbnail size, not just the small picture in the list, this is to improve the picture preview speed.

monkeyWie:
> I tried to run it locally and I see a lot of errors in ui.

I found this error,it was caused by merge conflict,I will fix it.

http/thumbnail.go
o1egl: In that case let's call this api preview and add `size` param with 2 possible values `thumb` and `big` to generate images with predefined sizes. like `/preview/thumb/path/to/file`. So we can use imaging.Thumbnail() and imaging.Fit() functions for that

http/thumbnail.go
```diff
@@ -0,0 +1,92 @@
...
+		Modify:  d.user.Perm.Modify,
+		Expand:  true,
+		Checker: d,
+	})
+	if err != nil {
+		return errToStatus(err), err
+	}
+
+	if file.IsDir || file.Type != "image" {
+		return http.StatusNotFound, nil
```
o1egl: How about returning StatusNotImplemented code?
```suggestion
		return http.StatusNotImplemented, nil
```

o1egl (changes requested):

http/thumbnail.go
monkeyWie: What do you think of the following changes?
1. add `/preview/thumb/file` API for list icon,fixed size 128*128,like this:
image
2. change `/thumbnail/file` API to `/preview/big/file`  for review, other unchanged.

http/thumbnail.go
monkeyWie: ok.

http/thumbnail.go
o1egl: I think we can use single handler with path param `/preview/{size}`.

monkeyWie: Of course,and the most important thing is to preview the picture max size, or keep the 1080px?

o1egl: Let's keep it 1080px. User can always download original one.

monkeyWie: Hi,I have a new commit,please review again.

http/prewiew.go
```diff
@@ -0,0 +1,94 @@
...
+	if err != nil {
+		return errToStatus(err), err
+	}
+	format, err := imaging.FormatFromExtension(file.Extension)
+	if err != nil {
+		return http.StatusNotFound, err
+	}
+	var dstImage image.Image
+	if size == sizeBig {
+		dstImage = fitResizeImage(srcImg, 1080)
```
o1egl: Please use `imaging.Fit()`

http/prewiew.go
```diff
@@ -0,0 +1,94 @@
...
+	} else {
+		dstImage = fitResizeImage(srcImg, 128)
```
o1egl: Please use `imaging .Thumbnail()`

o1egl: Also I would consider using one of those filters. We don't need super sharp image for thumbnails
```
- Linear
	Bilinear resampling filter, produces a smooth output. Faster than cubic filters.

- Box
	Simple and fast averaging filter appropriate for downscaling.
	When upscaling it's similar to NearestNeighbor.

- NearestNeighbor
	Fastest resampling filter, no antialiasing.
```

http/http.go
```diff
@@ -59,6 +59,8 @@ func NewHandler(store *storage.Storage, server *settings.Server) (http.Handler,
 	api.Handle("/settings", monkey(settingsPutHandler, "")).Methods("PUT")
 
 	api.PathPrefix("/raw").Handler(monkey(rawHandler, "/api/raw")).Methods("GET")
+	api.PathPrefix("/preview/thumb").Handler(monkey(previewThumbHandler, "/api/preview/thumb")).Methods("GET")
```
o1egl: Use path params
```suggestion
	api.PathPrefix("/preview/{size}").Handler(monkey(previewThumbHandler, "/api/preview")).Methods("GET")
```

http/prewiew.go
```diff
@@ -0,0 +1,94 @@
...
+	"net/http"
+	"net/url"
+
+	"github.com/disintegration/imaging"
+
+	"github.com/filebrowser/filebrowser/v2/files"
+)
+
+const sizeThumb = "thumb"
+const sizeBig = "big"
```
o1egl: Please group constants
```suggestion
const (
	sizeThumb = "thumb"
	sizeBig = "big"
)
```

http/prewiew.go
```diff
@@ -0,0 +1,94 @@
...
+		}
+		if size != sizeBig && size != sizeThumb {
+			return http.StatusNotFound, nil
+		}
+
+		return previewFileHandler(w, r, file, size)
+	})
+}
+
+func previewFileHandler(w http.ResponseWriter, r *http.Request, file *files.FileInfo, size string) (int, error) {
```
o1egl:
```suggestion
func imagePreviewHandler(w http.ResponseWriter, r *http.Request, file *files.FileInfo, size string) (int, error) {
```

http/prewiew.go
```diff
@@ -0,0 +1,94 @@
...
+		return errToStatus(err), err
+	}
+	defer fd.Close()
+
+	if r.URL.Query().Get("inline") == "true" {
+		w.Header().Set("Content-Disposition", "inline")
+	} else {
+		// As per RFC6266 section 4.3
+		w.Header().Set("Content-Disposition", "attachment; filename*=utf-8''"+url.PathEscape(file.Name))
+	}
```
o1egl: I think it should be done just before encoding process

o1egl (changes requested):
I would rewrite it to something like that. With that we can keep this handler flexible to support different file types:
```go
const (
	sizeThumb = "thumb"
	sizeBig = "big"
)

type imageProcessor func(img image.Image) (image.Image, error)

func previewHandler(w http.ResponseWriter, r *http.Request, d *data) (int, error) {
	if !d.user.Perm.Download {
		return http.StatusAccepted, nil
	}

	file, err := files.NewFileInfo(files.FileOptions{
		Fs:      d.user.Fs,
		Path:    r.URL.Path,
		Modify:  d.user.Perm.Modify,
		Expand:  true,
		Checker: d,
	})
	if err != nil {
		return errToStatus(err), err
	}

	size := mux.Vars(r)["size"]
	switch file.Type {
	case "image":
		return handleImagePreview(w, file, size)
	default:
		return http.StatusNotImplemented, fmt.Errorf("can't create preview for %s type", file.Type)
	}
}

func handleImagePreview(w http.ResponseWriter, file *files.FileInfo, size string) (int, error) {
	var imgProcessor imageProcessor
	switch size {
	case sizeBig:
		imgProcessor = func(img image.Image) (image.Image, error) {
			return imaging.Fit(img, 1080, 1080, imaging.Lanczos), nil
		}
	case sizeThumb:
		imgProcessor = func(img image.Image) (image.Image, error) {
			return imaging.Thumbnail(img, 128, 128, imaging.Box), nil
		}
	default:
		return http.StatusBadRequest, fmt.Errorf("unsupported preview size %s", size)
	}

	fd, err := file.Fs.Open(file.Path)
	if err != nil {
		return errToStatus(err), err
	}
	defer fd.Close()
	format, err := imaging.FormatFromExtension(file.Extension)
	if err != nil {
		return http.StatusNotImplemented, err
	}
	img, err := imaging.Decode(fd, imaging.AutoOrientation(true))
	if err != nil {
		return errToStatus(err), err
	}
	img, err = imgProcessor(img)
	if err != nil {
		return errToStatus(err), err
	}
	if imaging.Encode(w, img, format) != nil {
		return errToStatus(err), err
	}
	return 0, nil
}
```

http/prewiew.go
monkeyWie: oh,this pretty good!

http/http.go
monkeyWie: If use path params,I need trim `r.URL.Path`,like `/thumb/path/file`  trim  to `/path/file`,do you have any good suggestions for this?

o1egl: You can define filepath as another param.

monkeyWie:
@o1egl I have new submits,please review again.

http/preview.go
```diff
@@ -0,0 +1,98 @@
...
+		return http.StatusAccepted, nil
+	}
+	vars := mux.Vars(r)
+	size := vars["size"]
+	if size != sizeBig && size != sizeThumb {
+		return http.StatusNotImplemented, nil
+	}
+
+	// Resolve file path from URL
+	path := "/" + strings.Join(strings.Split(r.URL.Path, "/")[2:], "/")
```
o1egl: Why not to use the path as a variable like the size? Now it looks like a real magic :)

http/preview.go
```diff
@@ -0,0 +1,98 @@
...
+
+type imageProcessor func(src image.Image) (image.Image, error)
+
+var previewHandler = withUser(func(w http.ResponseWriter, r *http.Request, d *data) (int, error) {
+	if !d.user.Perm.Download {
+		return http.StatusAccepted, nil
+	}
+	vars := mux.Vars(r)
+	size := vars["size"]
+	if size != sizeBig && size != sizeThumb {
```
o1egl, line 27: I think it should be 
```suggestion
	if size != sizeBig || size != sizeThumb {
```

o1egl (commented):
@monkeyWie it still doesn't work. The ui is broken

http/preview.go
monkeyWie: For this condition:
```
if size != sizeBig || size != sizeThumb
```
If the size is "thumb",an error will be returned directly.

http/preview.go
monkeyWie: Do you mean like this?
```
api.PathPrefix("/preview/{size}/{path}")
```

http/preview.go
o1egl: Oh. My mistake. Newer mind

http/preview.go
o1egl: Yes

http/preview.go
monkeyWie: 🤣

http/preview.go
monkeyWie: This way doesn't work when the path like `photos/xxx.jpg`,the request URI is `/preview/big/photos/xxx.jpg`.

o1egl: Use regexp for that `/preview/{size}{path:.*}`

monkeyWie: I got it👍

monkeyWie:
> @monkeyWie it still doesn't work. The ui is broken

Can't build frontend?

o1egl:
> > @monkeyWie it still doesn't work. The ui is broken
> 
> Can't build frontend?

UI is broken. It displays blank page.

monkeyWie:
> > > @monkeyWie it still doesn't work. The ui is broken
> > 
> > 
> > Can't build frontend?
> 
> UI is broken. It displays blank page.

I test no problem,can you rebuild frontend again?

frontend/src/components/files/Preview.vue
```diff
@@ -86,8 +86,15 @@ export default {
     download () {
       return `${baseURL}/api/raw${this.req.path}?auth=${this.jwt}`
     },
+    thumbnail () {
+      if (this.req.type === 'image') {
+        return `${baseURL}/api/preview/big${this.req.path}?auth=${this.jwt}`
+      } else {
```
o1egl: else is redundant

frontend/src/components/files/Preview.vue
```diff
@@ -86,8 +86,15 @@ export default {
     download () {
       return `${baseURL}/api/raw${this.req.path}?auth=${this.jwt}`
     },
+    thumbnail () {
```
o1egl:
```suggestion
    previewUrl() {
```

http/raw.go
```diff
@@ -22,7 +22,7 @@ func parseQueryFiles(r *http.Request, f *files.FileInfo, _ *users.User) ([]strin
 		fileSlice = append(fileSlice, f.Path)
 	} else {
 		for _, name := range names {
-			name, err := url.QueryUnescape(strings.Replace(name, "+", "%2B", -1)) //nolint:shadow
+			name, err := url.QueryUnescape(strings.Replace(name, "+", "%2B", -1)) // nolint:shadow
```
o1egl: this change is not correct. there should not be a space in nolint comment

http/raw.go
```diff
@@ -35,7 +35,7 @@ func parseQueryFiles(r *http.Request, f *files.FileInfo, _ *users.User) ([]strin
 	return fileSlice, nil
 }
 
-//nolint: goconst
+// nolint: goconst
```
o1egl: same

o1egl (changes requested):
minor changes

monkeyWie:
@o1egl I think it's ok

o1egl (approved):

PR #981, opened 2020-06-11
Add folder upload

ramiresviana:
**Description**
Implements ability to upload folders via drag and drop.

**Further comments**
Uses DataTransferItem.webkitGetAsEntry() to recursively retrieve data from directories, allows multiple selection with files and directories. Works with modern browsers, fallback to old DataTransfer API when not compatible.

### DIFF
--- a/frontend/src/components/buttons/Upload.vue
+++ b/frontend/src/components/buttons/Upload.vue
@@ -10,7 +10,11 @@ export default {
   name: 'upload-button',
   methods: {
     upload: function () {
-      document.getElementById('upload-input').click()
+      if (typeof(DataTransferItem.prototype.webkitGetAsEntry) !== 'undefined') {
+        this.$store.commit('showHover', 'upload')
+      } else {
+        document.getElementById('upload-input').click();
+      }
     }
   }
 }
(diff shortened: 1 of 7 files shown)


o1egl:
@ramiresviana thank you for your PR. Could you please fix build errors?

ramiresviana:
@o1egl fixed build errors.

o1egl:
@ramiresviana great job. Can you please add support of upload button too?

ramiresviana:
@o1egl is not possible to make an input that accepts both files and folders. What can be made is add another button in the header that only allows selecting folders for upload.

o1egl:
@ramiresviana do you have an idea how to display it? Maybe we can display a popup with 2 file/filder options if browser supports it.

ramiresviana:
@o1egl i can make an prompt that opens when the user clicks on upload button, like this:
image

o1egl:
@ramiresviana that would be perfect! But let’s add browser compatibility check and display it only for supported ones.

ramiresviana:
- Added upload prompt: shows when header upload button is clicked. When browser does not support webkitGetAsEntry for folder upload, file picker window will be showed without prompt.
- Empty folder upload: allows uploading empty folder structure. Works only with drag and drop, because form file input does not return empty folders.

@o1egl the buttons in the upload prompt are responsive and can be reused in other components if needed.

o1egl:
@ramiresviana awesome job!

PR #987, opened 2020-06-17
Fix documentation links on README

haylinmoore:
**Description**
I fixed the links on the README. The current links lead to 404s and were kinda confusing so I fixed them and added a few more links

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,16 +14,20 @@ filebrowser provides a file managing interface within a specified directory and 
 
 ## Features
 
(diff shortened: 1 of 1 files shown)


o1egl:
@hamptonmoore thank you for your contribution. Could you please fix all links to use new https://filebrowser.org domain?

haylinmoore:
Yes :) I will get around to that tonight, I was not aware of that domain

o1egl:
And please use https://filebrowser.org/{{section}} instead of a link to the .md file

PR #1012, opened 2020-07-03
Added preview bypass for .gif files

Salzian:
This PR adds a bypass for the preview generation of GIF files. I noticed gif files use the default gif.Encode function which only processes the first frame, making the whole GIF purpose useless. When previewing gif files right now, they're not animated. :frowning_face: 

The bypass is only activated for big previews since I thought a bunch of small gif files in a folder view looks silly and is probably not too keen on performance. Also the previews where streched to squares and I didn't want to go down that rabbit hole.

I literally learned Go for this within 2 hours but got it working. After all it's just a simple bypass using another function for gifs. Any feedback and suggestions are greatly appreciated.

I also thought about incorporating the process in the preexisting imgProcessor but it takes multiple seconds(!) to process all frames. I think just handing it out as is, is the better solution here. When you handle gif files you know they're not optimized for size.

Also, maybe this behavior can be hidden behind a settings toggle if desirable. I think previewing gifs in big mode should be animated by default.

### DIFF
--- a/http/preview.go
+++ b/http/preview.go
@@ -81,13 +81,13 @@ func handleImagePreview(w http.ResponseWriter, r *http.Request, file *files.File
 	defer fd.Close()
 
 	if format == imaging.GIF && size == sizeBig {
-		g, err := gif.DecodeAll(fd)
-		if err != nil {
-			return errToStatus(err), err
+		g, gerr := gif.DecodeAll(fd)
+		if gerr != nil {
+			return errToStatus(gerr), err
 		}
 
 		if gif.EncodeAll(w, g) != nil {
-			return errToStatus(err), err
+			return errToStatus(gerr), err
 		}
 		return 0, nil
 	}
--- a/http/preview.go
+++ b/http/preview.go
@@ -3,7 +3,6 @@ package http
 import (
 	"fmt"
 	"image"
-	"image/gif"
 	"net/http"
 
 	"github.com/disintegration/imaging"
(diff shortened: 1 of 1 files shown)


http/preview.go
```diff
@@ -79,6 +80,18 @@ func handleImagePreview(w http.ResponseWriter, r *http.Request, file *files.File
 	}
 	defer fd.Close()
 
+	if format == imaging.GIF && size == sizeBig {
+		g, gerr := gif.DecodeAll(fd)
```
o1egl: It doesn't make sense do decode an image. Just send it as is

o1egl (changes requested):

Salzian:
I was stuck in the image encoding thinking 😅 I've replaced it with the rawFileHandler and I pass the gif now directly to the output.

I also moved the logic above the imgProcessor initialization since it's useless to initialize and then not using it.

http/preview.go
Salzian: I was stuck in the image encoding thinking 😅 I've replaced it with the rawFileHandler and I pass the gif now directly to the output.

I also moved the logic above the imgProcessor initialization since it's useless to initialize and then not using it.

http/preview.go
```diff
@@ -59,6 +59,20 @@ func handleImagePreview(w http.ResponseWriter, r *http.Request, file *files.File
...
+	if err != nil {
+		return errToStatus(err), err
+	}
+	defer fd.Close()
+
+	if format == imaging.GIF && size == sizeBig {
+		_, err := rawFileHandler(w, r, file)
+		if err != nil {
+			return errToStatus(err), err
+		}
```
o1egl:
```suggestion
		if _, err := rawFileHandler(w, r, file); err != nil { //nolint:govet
			return errToStatus(err), err
		}
```

o1egl (commented):
Please fix linter errors

http/preview.go
Salzian: I've now set up the linter on my local environment. I'm curious why it complains about shadowing when the same thing seems to happen in L62 where Fs.Open overrides err. It doesn't complain there.

Also I'm still getting a warning of `WARN [runner/nolint] Found unknown linters in //nolint directives: mnd, shadow`. Any ideas? Or is this known behaviour?

o1egl: Just skip warnings for now

o1egl (approved):

PR #1020, opened 2020-07-09
Update commands.go

oakboat:
I use it with nginx and shell is not working.
When using shell,it show
Jun 24 10:01:00 crazy filebrowser[16683]: 2020/06/24 10:01:00 http: superfluous response.WriteHeader call from github.com/filebrowser/filebrowser/v2/http.handle.func1 (
Jun 24 10:01:00 crazy filebrowser[16683]: 2020/06/24 10:01:00 /: 500 127.0.0.1 websocket: request origin not allowed by Upgrader.CheckOrigin

by this it works

:heart: Thank you!

### DIFF
--- a/http/commands.go
+++ b/http/commands.go
@@ -19,6 +19,9 @@ const (
 )
 
 var upgrader = websocket.Upgrader{
+	CheckOrigin: func(r *http.Request) bool {
+	return true
+	},
 	ReadBufferSize:  1024,
 	WriteBufferSize: 1024,
 }


http/commands.go
```diff
@@ -19,6 +19,9 @@ const (
 )
 
 var upgrader = websocket.Upgrader{
+	CheckOrigin: func(r *http.Request) bool {
+	return true
```
o1egl, line 23: We should not blindly trust any origin

o1egl (changes requested):
It is not secure to blindly trust any origin. You can add a config param with allowed origin and check it in this function

PR #1021, opened 2020-07-09
Upload queue

ramiresviana:
A more modular approach of upload handling, using vuex module and utils file for upload processing functions. Fixes the following issues:

### Non persistent upload counters

- Uploading states are being stored in the listing component, which is destroyed on route changes, making counter inconsistent for next uploads. States should be stored in the application store, which allows the component to be destroyed without losing the uploading states.

### Multiple file upload cause simultaneous requests

- The application makes a request for every upload, having multiple uploads equals simultaneous requests, causing browser resource usage issues when uploading a large amount of files. A queue is required for limiting the amount of uploads in the same time.

### DIFF
--- a/frontend/src/components/files/Listing.vue
+++ b/frontend/src/components/files/Listing.vue
@@ -103,7 +103,7 @@ export default {
     }
   },
   computed: {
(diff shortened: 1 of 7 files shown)


o1egl:
Awesome job, thanks!

Issue #1025, opened 2020-07-12
I Created a Dark Theme


yurividal:
Hi. Just wanted to share this theme for filebrowser with you, maybe you want to check it out.
Its a better dark theme.
https://github.com/yurividal/filebrowserdarktheme

mishailovic:
Beautiful! Already got my star. Much better than stock dark theme

ramiresviana:
Thank you for the contribution, looks like i made a mistake on the current dark theme colors. I will revisit the material design guidelines to make it better.

dantebarba:
+1 if it could be merged.

Issue #1022, opened 2020-07-10
Download Page is not 100% dark themed.


yurividal:
The share box is white on the dark theme.
I would suggest making it grey.

`.share__box__info {
    padding: 2em 3em;
    background-color: #2b2727;
}`

`.share__box__download {
    width: 100%;
    padding: 1em;
    cursor: pointer;
    background: #4b4b4b;
}`

Issue #976, opened 2020-06-03
Blinking previewer


minostauros:
**Description**

When we click next button on the previewer, the next() function is called as
https://github.com/filebrowser/filebrowser/blob/34bdb8fcfcc947bedda0ec35eaf5b479ea6ad614/frontend/src/components/files/Preview.vue#L123-L125

**What is happening?**

The previewer window is reloaded every time we _keep move fastly to previous or next link_.
It causes blinking effect.
https://github.com/filebrowser/filebrowser/blob/6e54dff40dd458601669078a50c8b6f65b2481ac/frontend/src/css/styles.css#L109-L118

![blinking mov](https://user-images.githubusercontent.com/6764739/83622891-8cf48280-a5cb-11ea-8970-4b446f2aa6f1.gif)

**Expected behaviour**

The #previewer div must be loaded faster.

o1egl:
Fixed in https://github.com/filebrowser/filebrowser/pull/993

Issue #1011, opened 2020-07-01
Duplicate files in same directory


PostPollux:
**Is your feature request related to a problem? Please describe.**
I did not find a way to just duplicate a file in the same directory. In my case I have some .env.example files that I just want to copy and then rename to just .env
When I click on "copy" and navigate to the path where the file sits, the "copy" button just gets greyed out.

**Describe the solution you'd like**
I think in the copy dialog instead of greying out that "copy" button it should just copy the file with a new name to that directory, just like every operating system does. Like "file.ext" -> "file (1).ext" 

Maybe even add a new button for just duplicating the file (right beside the copy button), so we do not have to navigate to that path in the copy dialog all the time if we just want to duplicate a file.

Additionally, for convenience I think it would be great to add a shortcut for this feature:
shift+d (as ctrl+d already is "add bookmark" in most browsers)

PR #1026, opened 2020-07-14
Some fixes

ramiresviana:
### Directory conflict checking

- File scan is using path attribute for uploaded directory items while the conflict checking method uses fullPath item attribute, causing conflict checking issues when uploading folders. The directory items returned by scan method should have fullPath attribute instead of path.

### Dark theme missing colors #1022 #1025

- Share box, share prompt and listing error message are missing dark theme colors. Input validation colors are missing from user profile password field. The surface colors are kinda out of the material design guidelines. Added new colors following the dark theme material guidelines, with branded dark surfaces. Added dark styles for share box, share prompt, listing error message and other small elements.

### Prompt when closing windows with active uploads

- The post frontend file api method sets the window beforeunload event for every new upload, and when the first upload is finished the event listener is removed, causing no confirmation prompt for the next pending uploads. The beforeunload listener should be set when the first upload is started and should be removed when the last upload was finished.

### Remove incomplete uploaded files

- When a upload is being made, there are a serie of issues that the frontend can go through while the file is being transferred, like losing the connection or a browser crash. The incomplete uploaded file is available on the server, without the user being aware that it is a corrupt file. Upload requests that ends prematurely or in unexpected way should have its file removed from the disk.

### Blinking previewer #976

- Navigating between files in previewer mode cause the preview component destroyed and its visibility toggled. A state is required to keep the component rendered between navigation.

### DIFF
--- a/http/resource.go
+++ b/http/resource.go
@@ -148,9 +148,28 @@ var resourcePatchHandler = withUser(func(w http.ResponseWriter, r *http.Request,
 		return http.StatusForbidden, nil
 	}
 
-	if r.URL.Query().Get("override") != "true" {
-		if _, err := d.user.Fs.Stat(dst); err == nil {
+	override := r.URL.Query().Get("override") == "true"
+	rename := r.URL.Query().Get("rename") == "true"
+
+	if !override && !rename {
+		if _, err = d.user.Fs.Stat(dst); err == nil {
 			return http.StatusConflict, nil
+		}
+	}
+
+	if rename {
+		counter := 1
+		dir, name := filepath.Split(dst)
+		ext := filepath.Ext(name)
+		base := strings.TrimSuffix(name, ext)
+
+		for {
+			if _, err = d.user.Fs.Stat(dst); err != nil {
+				break
+			}
+			new := fmt.Sprintf("%s(%d)%s", base, counter, ext)
+			dst = filepath.Join(dir, new)
+			counter++
 		}
 	}
 
--- a/http/resource.go
+++ b/http/resource.go
@@ -127,6 +127,10 @@ var resourcePostPutHandler = withUser(func(w http.ResponseWriter, r *http.Reques
 		return nil
 	}, action, r.URL.Path, "", d.user)
 
+	if err != nil {
+		_ = d.user.Fs.RemoveAll(r.URL.Path)
+	}
+
 	return errToStatus(err), err
 })
 
--- a/http/resource.go
+++ b/http/resource.go
@@ -148,6 +148,12 @@ var resourcePatchHandler = withUser(func(w http.ResponseWriter, r *http.Request,
 		return http.StatusForbidden, nil
 	}
 
+	if r.URL.Query().Get("override") != "true" {
+		if _, err := d.user.Fs.Stat(dst); err == nil {
+			return http.StatusConflict, nil
+		}
+	}
+
 	err = d.RunHook(func() error {
 		switch action {
 		// TODO: use enum
--- a/fileutils/file.go
+++ b/fileutils/file.go
@@ -2,6 +2,7 @@ package fileutils
 
 import (
 	"io"
+	"os"
 	"path/filepath"
 
 	"github.com/spf13/afero"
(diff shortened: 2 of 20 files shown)


o1egl:
Thank you for your contribution. I found that preview works better but it still blinking. 
Check out this recoding:
[IMAGE ALT TEXT HERE](https://www.youtube.com/watch?v=xdjMIaQ5Bgo)

ramiresviana:
@o1egl i've reviewed my approach to keep components rendered all time to prevent blinking.

- Added state for the file list in the previewer, instead of making a new request every time the previewer changes.
- Buttons will be rendered all times, but will remain in a disabled state during loading.
- The previewer will show images in the center of the screen while loading
- Resizing the screen will now cause the position of previewer images to be recalculated based on the center of the screen.

ramiresviana:
### Check file conflict when copying or moving

- Using copy or move buttons can cause conflict errors when the destination have files with the same name. Copying, moving and pasting actions should have conflict checking to give the user options for replacing the files.

### Option for renaming files with conflict

- Is not possible to keep both files in a conflict scenario, since the only option available is to overwrite existing files. The rename option will add a prefix to conflicting files, allowing the user to have both files. When a file with the prefix already exists, the prefix will be appended until a file with the new name dont exists.

### Duplicate files in the same directory #1011 

- With the renaming conflict files feature, the block of duplicating files in the same directory can be removed. Duplicate files in the same directory will be renamed automatically without prompt.

### Reset clipboard after pasting cutted files

- Keeping cutted files in the clipboard state after pasting then, will cause not found error in the next pasting, because the file location was changed. Clipboard should reset after pasting cutted files.

http/resource.go
```diff
@@ -144,6 +148,27 @@ var resourcePatchHandler = withUser(func(w http.ResponseWriter, r *http.Request,
...
+	rename := r.URL.Query().Get("rename") == "true"
+
+	if !override && !rename {
+		if _, err = d.user.Fs.Stat(dst); err == nil {
+			return http.StatusConflict, nil
+		}
+	}
+
+	if rename {
+		suffix := "_copy"
```
o1egl: It might be a file with a different content. Instead of a copy let's add (N) suffix

http/resource.go
```diff
@@ -144,6 +148,27 @@ var resourcePatchHandler = withUser(func(w http.ResponseWriter, r *http.Request,
...
+		_, err = d.user.Fs.Stat(dst)
+		exists := err == nil
+
+		for exists {
+			dst = dst + suffix
```
o1egl: suffix should be put before file extension.

http/resource.go
```diff
@@ -144,6 +148,27 @@ var resourcePatchHandler = withUser(func(w http.ResponseWriter, r *http.Request,
...
+			return http.StatusConflict, nil
+		}
+	}
+
+	if rename {
+		suffix := "_copy"
+		_, err = d.user.Fs.Stat(dst)
+		exists := err == nil
+
+		for exists {
```
o1egl: I think better would be to rewrite it like:
```go
for {
	// ...
	if notExist {
		break;
	}
}
```
With that you can reduce code duplications

http/resource.go
```diff
@@ -144,6 +148,27 @@ var resourcePatchHandler = withUser(func(w http.ResponseWriter, r *http.Request,
 		return http.StatusForbidden, nil
 	}
 
+	override := r.URL.Query().Get("override") == "true"
+	rename := r.URL.Query().Get("rename") == "true"
+
+	if !override && !rename {
+		if _, err = d.user.Fs.Stat(dst); err == nil {
+			return http.StatusConflict, nil
```
o1egl, line 156: Error must be triggered only if src != dst

o1egl (changes requested):

o1egl:
@ramiresviana can you please squash blinking commits into a single commit

http/resource.go
ramiresviana: This verification is required when duplicating files in the same folder, in a conflict scenario the frontend need to inform that the files should be renamed.

ramiresviana:
@o1egl rename suffix is now in the (N) incremental format.

https://github.com/filebrowser/filebrowser/blob/2636f876ab8f88eea6d9548de524ca2339eb0843/http/resource.go#L160-L174

o1egl (approved):

PR #1042, opened 2020-07-21
Some Fixes

ramiresviana:
### Renaming uses OS path separator
The renaming procedure invokes Split() and Join() methods, it uses path separator that can be different for each OS. Since filebrowser fileutils functions uses "/" for path separator, the renamed file path can have inconsistent path separator, which can cause issues. On windows systems the rename procedure causes error. The solution is to have paths with the "/" separator, achieved by using ToSlash() and using string concatenation instead of Join().

### Parent directory copy verification
Copying a directory when the source is parent of the destination one, causes a loop ended by multiple recursive copies of the source directory and a backend error. Verification is needed when copying, error should be returned when the source is parent from the destination.

### API Fetch not showing errors
Some frontend errors are showing an Object message, without any useful information. The fetch() function returns object with text(), instead of XMLHttpRequest responseText attribute. To get the error message, the text() method should be used on fetch() responses.

### Opacity not resetting on drag end
The dragEnd is not being called, which causes the opacity to not reset after a drag. The dragenter and dragleave is required for checking when the drag starts and when the drags ends. Since the event is triggered when dragging over the ListingItem, a counter is needed to know when the reset should be called.

### DIFF
--- a/http/resource.go
+++ b/http/resource.go
@@ -10,6 +10,8 @@ import (
 	"path/filepath"
 	"strings"
 
+	"github.com/spf13/afero"
+
 	"github.com/filebrowser/filebrowser/v2/errors"
 	"github.com/filebrowser/filebrowser/v2/files"
 	"github.com/filebrowser/filebrowser/v2/fileutils"
@@ -139,38 +141,25 @@ var resourcePatchHandler = withUser(func(w http.ResponseWriter, r *http.Request,
 	dst := r.URL.Query().Get("destination")
 	action := r.URL.Query().Get("action")
 	dst, err := url.QueryUnescape(dst)
-
 	if err != nil {
 		return errToStatus(err), err
 	}
-
 	if dst == "/" || src == "/" {
 		return http.StatusForbidden, nil
+	}
+	if err = checkParent(src, dst); err != nil {
+		return http.StatusBadRequest, err
 	}
 
 	override := r.URL.Query().Get("override") == "true"
 	rename := r.URL.Query().Get("rename") == "true"
-
 	if !override && !rename {
 		if _, err = d.user.Fs.Stat(dst); err == nil {
 			return http.StatusConflict, nil
 		}
 	}
-
 	if rename {
-		counter := 1
-		dir, name := filepath.Split(dst)
-		ext := filepath.Ext(name)
-		base := strings.TrimSuffix(name, ext)
-
-		for {
-			if _, err = d.user.Fs.Stat(dst); err != nil {
-				break
-			}
-			new := fmt.Sprintf("%s(%d)%s", base, counter, ext)
-			dst = filepath.ToSlash(dir) + new
-			counter++
-		}
+		dst = addVersionSuffix(dst, d.user.Fs)
 	}
 
 	err = d.RunHook(func() error {
@@ -180,11 +169,14 @@ var resourcePatchHandler = withUser(func(w http.ResponseWriter, r *http.Request,
 			if !d.user.Perm.Create {
 				return errors.ErrPermissionDenied
 			}
+
 			return fileutils.Copy(d.user.Fs, src, dst)
 		case "rename":
 			if !d.user.Perm.Rename {
 				return errors.ErrPermissionDenied
 			}
+			dst = filepath.Clean("/" + dst)
+
 			return d.user.Fs.Rename(src, dst)
 		default:
 			return fmt.Errorf("unsupported action %s: %w", action, errors.ErrInvalidRequestParams)
@@ -193,3 +185,35 @@ var resourcePatchHandler = withUser(func(w http.ResponseWriter, r *http.Request,
 
 	return errToStatus(err), err
 })
+
+func checkParent(src, dst string) error {
+	rel, err := filepath.Rel(src, dst)
+	if err != nil {
+		return err
+	}
+
+	rel = filepath.ToSlash(rel)
+	if !strings.HasPrefix(rel, "../") && rel != ".." && rel != "." {
+		return errors.ErrSourceIsParent
+	}
+
+	return nil
+}
+
+func addVersionSuffix(path string, fs afero.Fs) string {
+	counter := 1
+	dir, name := filepath.Split(path)
+	ext := filepath.Ext(name)
+	base := strings.TrimSuffix(name, ext)
+
+	for {
+		if _, err := fs.Stat(path); err != nil {
+			break
+		}
+		renamed := fmt.Sprintf("%s(%d)%s", base, counter, ext)
+		path = filepath.ToSlash(dir) + renamed
+		counter++
+	}
+
+	return path
+}
--- a/http/resource.go
+++ b/http/resource.go
@@ -168,7 +168,7 @@ var resourcePatchHandler = withUser(func(w http.ResponseWriter, r *http.Request,
 				break
 			}
 			new := fmt.Sprintf("%s(%d)%s", base, counter, ext)
-			dst = filepath.Join(dir, new)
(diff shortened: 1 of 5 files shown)


ramiresviana:
### Conflict handling on upload button #1043 
The upload button was using old method this.handleFiles() that is now available with upload.handleFiles().

Also selecting file with same name consecutive times causes the change event to not be triggered, which was resolved by reseting the file input when clicking on the upload button.

o1egl (approved):

http/resource.go
```diff
@@ -180,6 +180,17 @@ var resourcePatchHandler = withUser(func(w http.ResponseWriter, r *http.Request,
...
+
+			rel, err := filepath.Rel(src, dst)//nolint:shadow
+			if err != nil {
+				return err
+			}
+
+			rel = filepath.ToSlash(rel)
+			if !strings.HasPrefix(rel, "../") && rel != ".." {
+				return errors.ErrSourceIsParent
+			}
```
o1egl: don't we want to do the same check for `rename`?

ramiresviana: Yeah, this is really necessary. I can add a function for checking:
```go
func CheckParent(src, dst string) error {
	rel, err := filepath.Rel(src, dst)
	if err != nil {
		return err
	}

	rel = filepath.ToSlash(rel)
	if !strings.HasPrefix(rel, "../") && rel != ".." {
		return errors.ErrSourceIsParent
	}

	return nil
}
```
And call it after permission checking on copy and rename actions:
```go
err = CheckParent(src, dst)
if err != nil {
	return err
}
```

o1egl: Looks good

ramiresviana: @o1egl for consistency, i am thinking to refactor the renaming into a function too:
```go
func AddSuffix(path string, fs afero.Fs) string {
	counter := 1
	dir, name := filepath.Split(path)
	ext := filepath.Ext(name)
	base := strings.TrimSuffix(name, ext)

	for {
		if _, err := fs.Stat(path); err != nil {
			break
		}
		renamed := fmt.Sprintf("%s(%d)%s", base, counter, ext)
		path = filepath.ToSlash(dir) + renamed
		counter++
	}

	return path
}
```
So it will be called after permission checking on copy and rename:
```go
if rename {
	dst = AddSuffix(dst, d.user.Fs)
}
```
Can i amend that into the parent verification commit?

o1egl: Sure. But maybe `addVersionSuffix` or something like that will give more context of that function.

http/resource.go
```diff
@@ -157,34 +159,38 @@ var resourcePatchHandler = withUser(func(w http.ResponseWriter, r *http.Request,
...
 			}
+
+			if rename {
+				dst = AddVersionSuffix(dst, d.user.Fs)
+			}
+
+			err = CheckParent(src, dst)
+			if err != nil {
+				return err
+			}
```
o1egl: I think it could be moved up. Next to the `if dst == "/" || src == "/"` check

ramiresviana: If there is no problem calling CheckParent() and AddVersionSuffix() before d.user.Perm checking, we can have this:

```go
...
if dst == "/" || src == "/" {
}

override := r.URL.Query().Get("override") == "true"
rename := r.URL.Query().Get("rename") == "true"

if !override && !rename {
	if _, err = d.user.Fs.Stat(dst); err == nil {
		return http.StatusConflict, nil
	}
}

if rename {
	dst = AddVersionSuffix(dst, d.user.Fs)
}

err = CheckParent(src, dst)
if err != nil {
	return err
}

err = d.RunHook(func() error {
...
```
CheckParent() needs to be called after AddVersionSuffix().

o1egl: Why it should be called after AddVersionSuffix? We need to check user input not AddVersionSuffix output

ramiresviana: Filepath.Rel(src, dst) returns "." when src equals to dst, that causes CheckParent() to return error and makes duplicating files on same folder not possible. I can make it:

1. Call AddVersionSuffix() before CheckParent(), always resulting in src != dst
2. Or have a clause on CheckParent() to not return error when rel == ".".

o1egl: How about
```go
var resourcePatchHandler = withUser(func(w http.ResponseWriter, r *http.Request, d *data) (int, error) {
	src := r.URL.Path
	dst := r.URL.Query().Get("destination")
	action := r.URL.Query().Get("action")
	dst, err := url.QueryUnescape(dst)
	if err != nil {
		return errToStatus(err), err
	}
	src = cleanPath(src)
	dst = cleanPath(dst)

	if err := checkParent(d.user.Fs, src, dst); err != nil {
		return err
	}
	...
}

func checkParent(fs afero.Fs, src, dst string) error {
	isSrcDir, err := afero.IsDir(fs, src)
	if err != nil {
		return err
	}
	if !isSrcDir {
		return nil
	}
	if src != dst && strings.HasPrefix(dst, src) {
		return errors.ErrSourceIsParent
	}
	return nil
}

func cleanPath(fPath string) string {
	return filepath.Clean(filepath.Join("/", fPath))
}
```

ramiresviana: - HasPrefix() is not good, checkParent() returns error when src="/a/b/c" and dst="/a/b/c2".
- CleanPath() is not needed for the copy action, the fileutils.Copy() will clean the path.
- filepath.Join() causes inconsistency between operational systems.
- For isSrcDir verification to work on checkParent, it is necessary to verify the dst too. rel != "." resolves that.

Other http handlers do not even use filepath.Clean. Since every http handler have its own way of handling path, there will be repetitive code and missing security procedures. The backend should apply a sanitizer for every request, so the http handlers can use a safe and standard path parameter. Maybe in the future i try to review the backend structure to make it more robust.

For now, i think this should be sufficient for this PR:

```go
var resourcePatchHandler = withUser(func(w http.ResponseWriter, r *http.Request, d *data) (int, error) {
	src := r.URL.Path
	dst := r.URL.Query().Get("destination")
	action := r.URL.Query().Get("action")
	dst, err := url.QueryUnescape(dst)
	if err != nil {
		return errToStatus(err), err
	}
	if dst == "/" || src == "/" {
	}
	if err = checkParent(src, dst); err != nil {
		return http.StatusBadRequest, err
	}

	override := r.URL.Query().Get("override") == "true"
	rename := r.URL.Query().Get("rename") == "true"
	if !override && !rename {
		if _, err = d.user.Fs.Stat(dst); err == nil {
			return http.StatusConflict, nil
		}
	}
	if rename {
		dst = addVersionSuffix(dst, d.user.Fs)
	}

	err = d.RunHook(func() error {
		switch action {
		// TODO: use enum
		case "copy":
			if !d.user.Perm.Create {
				return errors.ErrPermissionDenied
			}

			return fileutils.Copy(d.user.Fs, src, dst)
		case "rename":
			if !d.user.Perm.Rename {
				return errors.ErrPermissionDenied
			}
			dst = filepath.Clean("/" + dst)

			return d.user.Fs.Rename(src, dst)
		default:
			return fmt.Errorf("unsupported action %s: %w", action, errors.ErrInvalidRequestParams)
		}
	}, action, src, dst, d.user)

	return errToStatus(err), err
})

func checkParent(src, dst string) error {
	rel, err := filepath.Rel(src, dst)
	if err != nil {
		return err
	}

	rel = filepath.ToSlash(rel)
	if !strings.HasPrefix(rel, "../") && rel != ".." && rel != "." {
		return errors.ErrSourceIsParent
	}

	return nil
}
...
```

Tested on linux and windows.

o1egl: Fair enough

o1egl (approved):

Issue #1039, opened 2020-07-19
How to turn off the image thumbnail function?


skyrim663:
I am running filebrowser 2.3.0 on my server. But whenever I open a folder with a lot of pictures, the server's computing resources will be full immediately. I had to restart the server. So besides installing filebrowser version 2.2.0, is there any other solution? how to turn off the image thumbnail function?

nambrosch:
An option to disable the thumbnail feature might not be a bad idea, but I'm wondering what type of processor and disk (ssd, 5400rpm, sd card, etc) are in your device?  What "computing resources" are being exhausted?

skyrim663:
> 禁用缩略图功能的选项可能不是一个坏主意，但我想知道设备中装有哪种类型的处理器和磁盘（ssd，5400rpm，sd卡等）？哪些“计算资源”正在用尽？

The disk type is SSD. When opening a folder with too many pictures, the CPU is used the most. Probably because it does not have very high performance.
So how should I turn off the thumbnail function?
Can it be achieved by modifying a configuration file?

skyrim663:
> An option to disable the thumbnail feature might not be a bad idea, but I'm wondering what type of processor and disk (ssd, 5400rpm, sd card, etc) are in your device? What "computing resources" are being exhausted?

The disk type is SSD. When opening a folder with too many pictures, the CPU is used the most. Probably because it does not have very high performance.
So how should I turn off the thumbnail function?
Can it be achieved by modifying a configuration file?

o1egl:
At the moment there is no such functionality. I think adding caching and limiting the number of workers will solve your issue.

skyrim663:
> At the moment there is no such functionality. I think adding caching and limiting the number of workers will solve your issue.

Looking forward to the next version update can optimize this problem.

niubility000:
Me too,  but finally I found it's not cpu's limit, it's "out of memory". My memory is only 500MB.  So add some virtual memory. Everything is fine now.

Issue #1033, opened 2020-07-15
Cache thumbnail files


lennardwalter:
**Is your feature request related to a problem? Please describe.**
I'm running filebrowser on a pretty low-end machine. When opening a folder with some photos the CPU usage immediately goes to 100 percent and loading the thumbnails takes very long.

**Describe the solution you'd like**
Caching the thumbnail data so it doesn't need to be cropped again.

**Additional context**
Maybe there is even something like this implemented... I started using filebrowser just now, so it is possible I just didn't find the feature.

Issue #651, opened 2019-01-31
The new setup procedure and `--noauth`


eine:
I came to this context while trying to start filebrowser in a docker container with `noauth` and a predefined list of enabled commands. If filebrowser is started with:

``` bash
# ./filebrowser -a 0.0.0.0 --noauth
2019/01/31 01:50:57 No config file used
2019/01/31 01:50:57 Listening on [::]:8080
```

The quick setup procedure is executed and the defaults are applied. I.e., `&settings.Settings{...}` is created, where `Scope`, `Locale` and `Perm` are initialized, but `Commands` is left empty (uninitialized). Moreover, no CLI flag is provided, and it is not read from the config file.

As a result, the user needs to browse the Settings. However, adding commands in `Global Settings` does not have any effect, because `noauth` defaults to user `admin`, which was created with an empty list of commands. It must be changed in `User Management`, by editing the options for user `admin`.

Since I need to do it without user intervention, I checked the subcommands:

- `filebrowser config set --commands 'git,ghdl,pyhton'`. The change is properly applied, but it changes the `Global Settings`, not the ones of `admin`.
- `./filebrowser users update admin --commands 'ghdl,python,gtkwave,echo'`. This works. However, note that despite the usage prototype being `filebrowser users add <username> <password> [flags]`, I did not write any password.

So, the procedure requires to:

- Start filebrowser once with quick setup: ./filebrowser --noauth -a 0.0.0.0.
- Wait until filebrowser is started and listening, and close/kill it.
- Run `users update ...`. 
- Start filebrowser again. This time, no arguments are required.

---

Since waiting for the server to start in order to kill is not a clean procedure, I tried to find a work around with `config init`.

> NOTE: I need to copy the parameters from the configuration file to flags manually. Although flag `-c` is shown as inherited in https://docs.filebrowser.xyz/cli/filebrowser-config-init, the content of the configuration file is ignored. However, when filebrowser is started without arguments after `config init`, the configuration file applies. This might be hard to understand. See #653.

``` bash
 ./filebrowser config init --noauth -a 0.0.0.0 --commands 'ghdl,python,gtkwave,echo,bash,sh'
```

However, `--noauth` is not accepted. I can execute the same command without it. Then, even if I start `filebrowser --noauth` credentials are required. Moreover, no valid credentials exist, because the user list is empty. So, I need to add a user:

``` bash
./filebrowser users add admin --commands "sh,ghdl" --perm.admin
```

[...]

PR #1044, opened 2020-07-23
Image resize improvements

o1egl:
- Limit number or image resize workers (#1039)
- Add cli param to disable image thumbnails (#1039)
- Cache resized images (#1033)


ramiresviana:
I think the term img-resize in the cli params can cause confusion. Image resizing is related to the resize process applied on a image, that can be used with different goals. For example, image resize can be used to reduce disk and bandwidth usage or to have many sizes to display it in different screen sizes. I would suggest naming the params with thumbnail or preview words. Also i think the disable word should go first on "img-resize-disable", like "disable-thumbnail", for a more imperative description.

alexisLefebvre:
Hello, how do you enable the cache? I put `"cache-dir": "/cache"` in my config, mounted a directory from the host to Docker, restarted filebrowser and the cache directory is always empty.

niubility000:
> Hello, how do you enable the cache? I put `"cache-dir": "/cache"` in my config, mounted a directory from the host to Docker, restarted filebrowser and the cache directory is always empty.

me too.  I use a router as the server.

o1egl:
@alexandrestein @niubility000 you should use --cache-dir cli param

alexisLefebvre:
Thanks for the answer. How do we do it when using a docker-compose stack? I didn't found a way in the documentation.

alexisLefebvre:
I found how to pass CLI parameters in https://github.com/filebrowser/filebrowser/issues/651#issuecomment-473935396

So we can use cache like this:

```yaml
version: '3.3'

services:
  filebrowser:
    image: filebrowser/filebrowser:v2
    restart: 'no'
    ports:
      - 127.0.0.1:80:80
    command: "--cache-dir=/cache"
    volumes:
      - "/home/me/:/srv/:ro"
      - "./data/database.db:/database.db"
      - "./data/.filebrowser.json:/.filebrowser.json"
      - "./data/cache:/cache"
```

PR #1066, opened 2020-08-18
Preview mobile dropdown and image size button

ramiresviana:
### Preview mobile dropdown
The preview header do not have dropdown for mobile devices, which can cause issues with small screens. It should behave like the file listing header, with its own dropdown.

### Preview image size button
Images in the preview have a fixed size of 1080px, causing bigger images to have a fraction of its size. Having a button for toggling the image size allows users to view the original image.

### DIFF
--- a/frontend/src/components/Header.vue
+++ b/frontend/src/components/Header.vue
@@ -1,5 +1,5 @@
 <template>
-  <header v-if="!isEditor">
+  <header v-if="!isEditor && !isPreview">
     <div>
       <button @click="openSidebar" :aria-label="$t('buttons.toggleSidebar')" :title="$t('buttons.toggleSidebar')" class="action">
         <i class="material-icons">menu</i>
@@ -108,6 +108,7 @@ export default {
       'selectedCount',
       'isFiles',
       'isEditor',
+      'isPreview',
       'isListing',
       'isLogged'
     ]),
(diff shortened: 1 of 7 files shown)


o1egl:
@ramiresviana awesome PR, thanks. One question. What do you mean under ` causing bigger images to have a fraction of its size`. The resize for previews should keep image ratio and just resize it. 
Ideally would be to track `--disable-preview-resize` param and display preview size button only if resizing is enabled.

ramiresviana:
@o1egl this happens when having a fixed size for the previews:

```
original res	 preview res	 preview percentage
1280x720	 1080x608	 84%
1920x1080	 1080x608	 56%
3840x2160	 1080x608	 28%
7980x4320	 1080x608	 14%
```

When the user has a 720p image, the preview will be 84% of the original resolution, since the loss of detail is just 16% the user do not need to download the original file.

Having a 8K image, the preview will be 14% of the original resolution, that is losing 86% of the original image, causing users to be forced to download the original file and use other application to view the image.

> track --disable-preview-resize param and display preview size button only if resizing is enabled.

I will add the integration with --disable-preview-resize param.

o1egl:
@ramiresviana the idea behind this param was to display the image faster on the FE by reducing the size.

ramiresviana:
@o1egl the --disable-preview-resize param can partially achieve what i am proposing with the resize button, except by:

- CLI parameters arent user friendly: running FileBrowser is easy and do not require configuration, cli features may be used only by advanced users.
- Toggle buttons cover different scenarios: when using mobile data you probably want to save data by using resized previews. On a big size TV you probably want to see your photos on full size.

ramiresviana:
@o1egl added conditional for not displaying resize button if --disable-preview-resize is used.

Issue #1084, opened 2020-09-15
Button to show/hide dotfiles


kwsp:
It would be neat if we can add a simple **checkbox in the UI that shows/hides dotfiles**. This feature is available in most file managers (e.g. Nautilus, Dolphin).

I've looked at https://github.com/filebrowser/filebrowser/issues/322, https://github.com/filebrowser/filebrowser/issues/349, https://github.com/filebrowser/filebrowser/issues/489, but all of them are approaching this problem using the disallow/allow rules. 

A minimal version of this feature can be implemented completely in the frontend without touching the backend, although it would be nice if the user's preference is saved to the db or config file. 

I'm happy to work on this feature, but I want to hear what the maintainers think first :) 

Thanks for a great piece of software!

o1egl:
Sounds like a nice feature to add 👍 Feel free to make a PR

PR #1085, opened 2020-09-16
Hiddenfiles

kwsp:
This PR addresses https://github.com/filebrowser/filebrowser/issues/1084 through the following:

Frontend:
* Added a checkbox in Profile Settings to toggle show/hide dotfiles at the user level
* When the files api is called on a directory and `showHidden` is `true`, all directory entries that begin with a `.` are prune from the api results.

Backend:
* Added `showHidden` to the `userDefaults` and `userInfo`
* This attribute defaults to false (by default users do not see hidden files) and is sent along with other `userInfo` from the auth endpoint to the frontend. 

@o1egl please review :)

### DIFF
--- a/frontend/src/views/Files.vue
+++ b/frontend/src/views/Files.vue
@@ -10,13 +10,6 @@
         <router-link :to="link.url">{{ link.name }}</router-link>
       </span>
 
-      <span style="margin-left:auto;">
-        <input 
-          type="checkbox" 
-          v-model="showHidden">
-        <span>{{ $t('files.showHiddenFiles') }}</span>
-      </span>
-
     </div>
 
     <div v-if="error">
@@ -189,7 +182,7 @@ export default {
           return
         }
 
-        if (res.isDir && !this.showHidden) {
+        if (res.isDir && !this.user.showHidden) {
           res = pruneHiddenFiles(res)
         }
 
--- a/frontend/src/views/Files.vue
+++ b/frontend/src/views/Files.vue
@@ -9,6 +9,12 @@
         <span class="chevron"><i class="material-icons">keyboard_arrow_right</i></span>
         <router-link :to="link.url">{{ link.name }}</router-link>
       </span>
+
+      <span style="margin-left:auto;">
+        <input type="checkbox">
+        Show hidden files
+      </span>
+
     </div>
 
     <div v-if="error">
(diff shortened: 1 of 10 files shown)


frontend/src/views/Files.vue
```diff
@@ -9,6 +9,7 @@
         <span class="chevron"><i class="material-icons">keyboard_arrow_right</i></span>
         <router-link :to="link.url">{{ link.name }}</router-link>
       </span>
+
```
o1egl: Why did you add a newline here?

frontend/src/views/Files.vue
```diff
@@ -145,12 +167,16 @@ export default {
...
 
       try {
-        const res = await api.fetch(url)
+        let res = await api.fetch(url)
 
         if (clean(res.path) !== clean(`/${this.$route.params.pathMatch}`)) {
           return
         }
 
+        if (res.isDir && !this.user.showHidden) {
```
o1egl, line 175: why is it applied only for dirs?

o1egl (commented):
Great job. But I think this property should be active by default.

frontend/src/views/Files.vue
kwsp: I think I added some code here in a previous commit and removed it as I changed the approach and accidentally left a new line. Could remove it if you want.

frontend/src/views/Files.vue
kwsp: The frontend only shows the 'file browser' window when the result is a directory. If the result is a file, the previous of the file is shown, so the logic to hide dotfiles in the `browser` view doesn't apply.

kwsp:
> Great job. But I think this property should be active by default.

I've just made 'show hidden files' the default.

o1egl:
1. In current implementation showHidden will be true only for new users. We can reverse this flag to `hideHidden` or `hideDotFiles`. 
2. I don't think that this type filter should be done on client side. Let's move it to the server side.

ramiresviana:
Filtering do not work on the result of files.fetch(), it causes wrong values for items indexes.

For me, the checkbox input feels out of place on the file listing, should be self-describing without text labels.

Since is not an essential feature that needs to be always visible on the listing, i think it should be moved to settings in rules section.

kwsp:
> Filtering do not work on the result of files.fetch(), it causes wrong values for items indexes.
> 
> For me, the checkbox input feels out of place on the file listing, should be self-describing without text labels.
> 
> Since is not an essential feature that needs to be always visible on the listing, i think it should be moved to settings in rules section.

I have not tested `files.fetch()`, will look into it.

I've moved the checkbox to profile settings in the most recent commits, please pull those and see. If you have interest in moving the implementation from the client side to the server side as @o1egl suggested, please feel free to take a look! I might not have time in the coming weeks to do this.

Issue #1088, opened 2020-09-21
disable command execution such that it cannot be enabled by admin


dr-bonez:
**Is your feature request related to a problem? Please describe.**
It seems bad that compromising the admin password gives an attacker remote execution capabilities.

**Describe the solution you'd like**
A configuration option to disable the command execution feature.

dr-bonez:
resolved by #1090

PR #1089, opened 2020-09-21
Add Flag to Disable Exec

dr-bonez:
This PR addresses #1088 through the following:

Frontend:
- Added settings as global state
- Removed all ui elements pertaining to command execution if settings.execEnabled is false

Backend:
- added a config flag: `execEnabled` default `true`
- disabled command runner and hooks if execEnabled is false

I have never used Vue.js before today, so let me know if I did anything un-idiomatic and I'll change it
I have also only dabbled in golang before so same caviat there

### DIFF
--- a/cmd/config.go
+++ b/cmd/config.go
@@ -31,6 +31,7 @@ func addConfigFlags(flags *pflag.FlagSet) {
 	addServerFlags(flags)
 	addUserFlags(flags)
 	flags.BoolP("signup", "s", false, "allow users to signup")
+	flags.BoolP("execEnabled", "", true, "enable the Command Runner feature")
 	flags.String("shell", "", "shell command to which other commands should be appended")
 
 	flags.String("auth.method", string(auth.MethodJSONAuth), "authentication type")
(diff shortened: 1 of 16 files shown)


cmd/config.go
```diff
@@ -31,6 +31,7 @@ func addConfigFlags(flags *pflag.FlagSet) {
 	addServerFlags(flags)
 	addUserFlags(flags)
 	flags.BoolP("signup", "s", false, "allow users to signup")
+	flags.BoolP("execEnabled", "", true, "enable the Command Runner feature")
```
o1egl, line 34: Let's call it `disable-exec`

frontend/src/main.js
```diff
@@ -15,6 +16,7 @@ async function start () {
   } else {
     await login('', '', '')
   }
+  store.commit('setSettings', await settings.get())
```
ramiresviana, line 19: Getting settings requires authentication, causing hangs on login screen.

ramiresviana:
Using settings with states is a nice approach, but this implementation covers only the execEnable param, leaving the remaining settings as constants. Having multiple pieces of code for managing settings adds complexity to the frontend, requiring contributors to know both implementations, that can be bad for maintainability.

Since this is not the scope of this PR, i recommend using only constants for the settings.

dr-bonez:
Can you point me to an example of the constants?

ramiresviana:
@dr-bonez you can implement params and settings like this commit: https://github.com/filebrowser/filebrowser/commit/aa78e3ab1fcae6f618e811ba4e315a7a209f9df2

o1egl (commented):
Totally agree with @ramiresviana

dr-bonez:
closing in favor of #1090

Issue #655, opened 2019-01-31
Need to update Global Settings from the GUI in order to execute commands


eine:
After executing the following sequence:

``` bash
# filebrowser config init
# filebrowser users add user admin --commands '.*' --perm.admin
# filebrowser
2019/01/31 03:47:08 Using config file: /etc/filebrowser/.filebrowser.yml
2019/01/31 03:47:08 Listening on [::]:8080
```

I open the browser and login successfully. If I try to execute a command, I get:

``` bash
2019/01/31 03:47:31 /: 500 172.17.0.1:55494 fork/exec : no such file or directory
```

If I go to Settings > Global Settings and I click UPDATE (without modifying any content), it is fixed. And commands can now be properly executed.

kolaente:
I tried to reproduce this, wasn't able to. It worked without any issues even bevore updating the user via the gui.

eine:
Originally I saw this behaviour in filebrowser/docs#7. I could now reproduce it with:

``` bash
docker pull filebrowser/filebrowser

docker build -t fb_tmp - <<-EOF
FROM alpine
COPY --from=filebrowser/dev /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/ca-certificates.crt
COPY --from=filebrowser/filebrowser /filebrowser /usr/local/bin/filebrowser
EOF

drun="docker run --rm -v /$(pwd)/mydb.db://mydb.db"

touch mydb.db
$drun fb_tmp filebrowser config init -d //mydb.db
$drun fb_tmp filebrowser users add -d //mydb.db user admin --commands '.*' --perm.admin
$(command -v winpty) $drun -itp 8080:8080 fb_tmp filebrowser -d //mydb.db -a 0.0.0.0

docker rmi fb_tmp
```
``` bash
Sending build context to Docker daemon  2.048kB
Step 1/3 : FROM alpine
 ---> caf27325b298
Step 2/3 : COPY --from=filebrowser/dev /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/ca-certificates.crt
 ---> c5d4293376ef
Step 3/3 : COPY --from=filebrowser/filebrowser /filebrowser /usr/local/bin/filebrowser
 ---> e3d6f4ce37b8
Successfully built e3d6f4ce37b8
Successfully tagged fb_tmp:latest
[...]

PR #1106, opened 2020-10-03
fix: run Commands when set by "cmds add" cli (#655)

dannyp11:
### Description
Hello, I came across the exact issue #[655](https://github.com/filebrowser/filebrowser/issues/655) in that when adding cmds via `filebrowser` cli instead of in web UI, filebrowser fails to run the command due to leading whitespace (eg. input command is "`echo hello`" but filebrowser will run "` echo hello`")

### To reproduce
```
rm -f filebrowser.db
filebrowser config init
filebrowser cmds add "before_upload" "echo hello world"
filebrowser -r .
```

Try to upload 1 file
- Expected: file is uploaded successfully, `hello world` is printed to stdout
- Actual: `no such file or directory` error due to running " echo hello world" instead of "echo hello world"
```
2020/10/03 15:00:59 [INFO] Blocking Command: " echo hello world"
2020/10/03 15:00:59 /service-runner: 404 127.0.0.1 fork/exec : no such file or directory
```
### Root cause
- [Settings.Shell](https://github.com/filebrowser/filebrowser/blob/master/cmd/config_init.go#L35) is initiated as `[]string{""}` instead of `[]string{}` due to [strings.Split() expected behavior](https://github.com/golang/go/issues/35130)
- that will result in [`len(Shell)` always >= 1](https://github.com/filebrowser/filebrowser/blob/master/runner/parser.go#L17)
- that leads to `command = append(s.Shell, raw)` which is ["", "echo hello world"] which leads to `cmd := exec.Command("", "echo", "hello", "world")` [here](https://github.com/filebrowser/filebrowser/blob/master/runner/runner.go#L68)
=> runs exec command `" echo hello world"` instead of "echo hello world"

### Testing
1. Try reproduce step
1. File is uploaded successfully and "hello world" is printed
```
2020/10/03 15:03:07 [INFO] Blocking Command: "echo hello world"
hello world
```

:rotating_light: I have followed all the below:

:heart: Thank you!

### DIFF
--- a/cmd/config_init.go
+++ b/cmd/config_init.go
@@ -2,7 +2,6 @@ package cmd
 
 import (
 	"fmt"
-	"strings"
 
 	"github.com/spf13/cobra"
 
@@ -32,7 +31,7 @@ override the options.`,
 		s := &settings.Settings{
 			Key:        generateKey(),
 			Signup:     mustGetBool(flags, "signup"),
-			Shell:      strings.Split(strings.TrimSpace(mustGetString(flags, "shell")), " "),
+			Shell:      convertCmdStrToCmdArray(mustGetString(flags, "shell")),
 			AuthMethod: authMethod,
 			Defaults:   defaults,
 			Branding: settings.Branding{
--- a/cmd/config_set.go
+++ b/cmd/config_set.go
@@ -1,8 +1,6 @@
 package cmd
 
 import (
-	"strings"
-
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
 )
@@ -50,7 +48,7 @@ you want to change. Other options will remain unchanged.`,
 			case "auth.method":
 				hasAuth = true
 			case "shell":
-				set.Shell = strings.Split(strings.TrimSpace(mustGetString(flags, flag.Name)), " ")
+				set.Shell = convertCmdStrToCmdArray(mustGetString(flags, flag.Name))
 			case "branding.name":
 				set.Branding.Name = mustGetString(flags, flag.Name)
 			case "branding.disableExternal":
--- a/cmd/utils.go
+++ b/cmd/utils.go
@@ -7,6 +7,7 @@ import (
 	"log"
 	"os"
 	"path/filepath"
+	"strings"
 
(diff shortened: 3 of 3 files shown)


cmd/config_init.go
```diff
@@ -28,11 +28,12 @@ override the options.`,
 		flags := cmd.Flags()
 		getUserDefaults(flags, &defaults, true)
 		authMethod, auther := getAuthentication(flags)
+		trimmedShellStr := strings.TrimSpace(mustGetString(flags, "shell"))
```
o1egl: I think better would be to initialize correct slice before injecting it into the settings struct. 
```go
var shellCmd []string
trimmedShellStr := strings.TrimSpace(mustGetString(flags, "shell"))
if trimmedShellStr != "" {
	shellCmd = strings.Split(trimmedShellStr, " ")
}
```

o1egl (changes requested):

o1egl:
Good catch!

cmd/config_init.go
dannyp11: I agree, updated

dannyp11:
### Revision 2
- Addressed @o1egl comment

### Testing passed
```
2020/10/04 13:52:28 [INFO] Blocking Command: "echo hello world"
hello world
```

dannyp11:
### Revision 3
- Fix same bug in [config_set.go](https://github.com/filebrowser/filebrowser/blob/ad99bf180197e0e6d82231a86457585de16366a8/cmd/config_set.go#L53)
- Move cmd string split into `cmd/utils.go`

### Testing passed
```
2020/10/04 14:28:38 [INFO] Blocking Command: "echo hello world"
hello world
```

### Side note
@o1egl  why don't we have any unit test? I'm happy to setup simple one in separate PR if the team is ok with it, where can I bring this up?

o1egl:
@dannyp11 I think due to historical reason. I'm trying to add unit test on new code. At the moment I'm working on refactoring which will be released as v3 and after that unit tests will be required for PRs

o1egl (approved):
LGTM

PR #1113, opened 2020-10-07
Fix: WKWebview Custom Protocol

dr-bonez:
**Description**
In WKWebview (the webview for iOS), the Blob data type is not supported for non-http protocols. This can cause issues for custom browsers for iOS that use WKURLSchemeHandler. This change detects if the page is being hosted on a non-http URL scheme and will convert the Blob to an ArrayBuffer if so.

### DIFF
--- a/frontend/src/api/files.js
+++ b/frontend/src/api/files.js
@@ -85,8 +85,9 @@ export function download (format, ...files) {
 export async function post (url, content = '', overwrite = false, onupload) {
   url = removePrefix(url)
 
+  let bufferContent
   if (content instanceof Blob && !['http:', 'https:'].includes(window.location.protocol)) {
-    content = await content.arrayBuffer()
+    bufferContent = await new Response(content).arrayBuffer()
   }
 
   return new Promise((resolve, reject) => {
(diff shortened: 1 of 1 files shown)


o1egl:
@dr-bonez Can you please explain how the procol there could be non http(s)?

dr-bonez:
Sure.
iOS is deprecating the UIWebview and replacing it with WKWebview. WKWebview will only let you override the transport protocol if you use a custom url scheme, so if a browser wants to do anything special with how requests are fetched, the page can't be accessed over http(s). For example, the Consulate browser for iOS uses a custom protocol (onion://) in order to use Tor as well as do network level caching.

frontend/src/api/files.js
```diff
@@ -85,6 +85,11 @@ export function download (format, ...files) {
 export async function post (url, content = '', overwrite = false, onupload) {
   url = removePrefix(url)
 
+  let bufferContent
+  if (content instanceof Blob && !['http:', 'https:'].includes(window.location.protocol)) {
```
o1egl, line 89: I think better would be to compare only prefix with `startsWith`.

dr-bonez: that would be more performant however not strictly correct in what we are trying to catch. Someone could theoretically use a custom protocol `http+foo:` and that would cause this to break, as it is then considered a "non-http protocol" from iOS's standpoint, but would be considered an http protocol from filebrowser's.

Issue #1120, opened 2020-10-14
File handling should be case-insensitive.


vmxr:
**Description**
If a PDF document has an uppercase extension "foo.PDF", it is regarded as a text file instead of a PDF.
Files with lowercase extensions are handled correctly: "bar.pdf" will be displayed in a preview.

**Expected behaviour**
Uppercase extensions should be treated the same as lowercase extensions.
PDF files are then shown in a preview.
![image](https://user-images.githubusercontent.com/71821508/95972934-54837180-0e13-11eb-9b1c-1bcc9c6479af.png)

**What is happening instead?**
PDF with uppercase extensions are either not shown in the preview, or displayed using the text editor.
![image](https://user-images.githubusercontent.com/71821508/95972796-29991d80-0e13-11eb-9b0d-a1060e394713.png)

**Additional context**
Docker image: filebrowser/filebrowser   latest              d92417071f39
File Browser 2.8.0

**How to reproduce?**
Place a PDF document with uppercase extension and try to open it.

vmxr:
Probably related to this line of code:
https://github.com/filebrowser/filebrowser/blob/4c2a09425522c2e81949a0420b52067d7cc3be8f/frontend/src/components/files/Preview.vue#L55

Issue #1119, opened 2020-10-13
Invalid directory URL in search results


wolfpld:
When a directory is entered by navigating the UI, the page URL is as follows:

`https://example.com/files/test/`

However, when a search for `test` is performed, and the directory is entered from search results, the URL is different:

`https://example.com/files/test`

While both URLs present the same contents and everything is seemingly working in both cases, the upload process is broken in the latter one. Uploading a file, let's say `abc` will result in creating a file named `testabc` in the root directory, as the terminating path separator is missing in the malformed search results URL.

Issue #1110, opened 2020-10-07
Windows service capability


Spasman:
Right now, FileBrowser is unable to work as a Windows service, because it needs to be run from command prompt. Trying to add it as a service through a tertiary shortcut or a batch file makes Windows angry, and it feels like this could be resolved if there was some sort of parameter or trick you could use to override the command prompt only error.

Sorry for blowing up the issues tab, but this is a neat app 💯

ramiresviana:
Cobra library has a hook that dont allow running the program if it was started by the Windows Explorer. @o1egl i think we should disabled it, since Filebrowser does not require any CLI parameter to be initialized.

PR #1124, opened 2020-10-19
Some fixes

ramiresviana:
### File upload missing path slash
Ensures that the current path contain a slash at the end.

### Search missing path slash #1119
Adds a trailing slash to the end of the results url.

### Preview case sensitive extension #1120
Compares the file extension in lower case.

### Allow start from Windows explorer #1110
Disables the Mousetrap hook for Windows from Cobra cli library.


o1egl:
Great job!

PR #1127, opened 2020-10-26
add mavon-editor for markdown file

simzhangbest:
to make markdown file more readable and ediable. I embeded the mavon-editor in you filebrowser project. hope to help more developer to write md file. 
图片


ramiresviana:
Ace editor already features markdown syntax highlight. I think adding markdown support on the previewer or extending Ace is a better option than adding a whole new editor.

PR #1129, opened 2020-11-01
Commit sha to sidebar

zoidy:
Add the commit SHA to the sidebar right above the Help link. This makes it easier to track exactly which commit the binary corresponds to when building from source


o1egl:
Thank you for your contribution but I don't think that it make sense to display commit sha in the ui.

Issue #1091, opened 2020-09-23
Go to next Image preview when deleting an Image


debauer:
**Is your feature request related to a problem? Please describe.**
I wanted to sort out Images and after deleting an Image in "Imageviewer?!" the view is closed and i must open it again with the next Image. 

**Describe the solution you'd like**
The better Workflow would be to open the next Image.

PR #1134, opened 2020-11-05
Automatically jumps to the next photo when deleting while previewing

chief8192:
**Description**
Introduces a new `postDeleteLink` property to the store, and uses it to jump to the next photo when deleting while previewing, rather than returning to the files list every time.

Closes issue #1091


chief8192:
Found a bug. Closing this PR, and will submit a revised one.

PR #1139, opened 2020-11-08
Add single click mode

Namaneo:
This PR adds the ability to navigate through the files only using single clicks, using checkboxes to select one or multiple elements. As it directly relates to the user experience, this is configurable per user instead of globally. In my opinion, this improves the user experience on mobile, and that's the main reason I propose those changes. 🙂 

Though, some questions might need to be addressed:
* Would it be a better idea to enable it on mobile and disable it on desktop, instead of manually configuring it?
* I used raw checkboxes, do you want to replace them with Material looking ones instead? I didn't want to alter the package dependencies before asking.

### DIFF
--- a/cmd/users.go
+++ b/cmd/users.go
@@ -27,7 +27,7 @@ var usersCmd = &cobra.Command{
 
 func printUsers(usrs []*users.User) {
 	w := tabwriter.NewWriter(os.Stdout, 0, 0, 2, ' ', 0)
-	fmt.Fprintln(w, "ID\tUsername\tScope\tLocale\tV. Mode\tAdmin\tExecute\tCreate\tRename\tModify\tDelete\tShare\tDownload\tPwd Lock")
+	fmt.Fprintln(w, "ID\tUsername\tScope\tLocale\tV. Mode\tS.Click\tAdmin\tExecute\tCreate\tRename\tModify\tDelete\tShare\tDownload\tPwd Lock")
 
 	for _, u := range usrs {
 		fmt.Fprintf(w, "%d\t%s\t%s\t%s\t%s\t%t\t%t\t%t\t%t\t%t\t%t\t%t\t%t\t%t\t%t\t\n",
--- a/cmd/users.go
+++ b/cmd/users.go
@@ -30,12 +30,13 @@ func printUsers(usrs []*users.User) {
 	fmt.Fprintln(w, "ID\tUsername\tScope\tLocale\tV. Mode\tAdmin\tExecute\tCreate\tRename\tModify\tDelete\tShare\tDownload\tPwd Lock")
 
 	for _, u := range usrs {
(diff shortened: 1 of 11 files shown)


ramiresviana:
The single click idea is good, is the common behavior on mobile apps, but the checkbox dont fit on the listing, is too small for mobile, and is similar to an existing feature, the multiple selection mode that have a button in the header bar. I think we should improve the multiple selection mode to work better on mobile, having a better interface and toggling.

Namaneo:
Oh my bad, I'm pretty new to this project and haven't even seen the multi-selection button. Then, we could just remove the checkboxes to only rely on this button. The configuration entry would concern the single-click behavior on mobiles only. Desktop behavior would stay the same as before. What do you think?

ramiresviana:
Filebrowser on mobile and desktop is the same application and behaves in the same way for both those variations. Adding a feature that changes the behavior only on mobile can be bad for the project maintainability and can make users confused. Just let the users choose if they want to enable the feature.

Namaneo:
Alright, then I will rename the configuration option to something like "Use single clicks to open folders" instead, this will be more accurate. I will also remove the checkboxes: in this mode, only the top-right button will trigger the multi-selection.

cmd/users.go
```diff
@@ -30,12 +30,13 @@ func printUsers(usrs []*users.User) {
 	fmt.Fprintln(w, "ID\tUsername\tScope\tLocale\tV. Mode\tAdmin\tExecute\tCreate\tRename\tModify\tDelete\tShare\tDownload\tPwd Lock")
```
ramiresviana:
```suggestion
	fmt.Fprintln(w, "ID\tUsername\tScope\tLocale\tV. Mode\tS.Click\tAdmin\tExecute\tCreate\tRename\tModify\tDelete\tShare\tDownload\tPwd Lock")
```

frontend/src/i18n/en.json
```diff
@@ -173,6 +173,7 @@
     "globalRules": "This is a global set of allow and disallow rules. They apply to every user. You can define specific rules on each user's settings to override this ones.",
     "allowSignup": "Allow users to signup",
     "createUserDir": "Auto create user home dir while adding new user",
+    "singleClick": "Use single clicks to open folders",
```
ramiresviana:
```suggestion
    "singleClick": "Use single clicks to open files and directories",
```

ramiresviana (changes requested):

frontend/src/i18n/en.json
Namaneo: I will check in the evening, but I think those changes does not affect files, only directories.
I can rename `folders` to `directories` anyway. 🙂 

EDIT: Aaaand I'm wrong, it applies to files as well!

o1egl:
@Namaneo can you please resolve merge conflict?

ramiresviana (approved):

PR #1140, opened 2020-11-08
Automatically jumps to the next photo when deleting while previewing.

### LINKED ISSUE #1091: Go to next Image preview when deleting an Image

chief8192:
**Description**
Deletions made while previewing now immediately load the preview of the next file in the list. When the last file is deleted, previewing ends and returns to the file list, as before.

Closes #1091

### DIFF
--- a/frontend/src/components/files/Preview.vue
+++ b/frontend/src/components/files/Preview.vue
@@ -62,7 +62,7 @@
 </template>
 
 <script>
-import { mapState } from 'vuex'
+import { mapMutations, mapState } from 'vuex'
 import url from '@/utils/url'
 import { baseURL, resizePreview } from '@/utils/constants'
 import { files as api } from '@/api'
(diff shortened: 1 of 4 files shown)


ramiresviana:
I think we can have a better solution for how components react when the delete button is pressed. Using events you wont need those store states for component communication and all the preview logic can be written on the component itself.

chief8192:
It's entirely possible, I don't have a ton of VueJS experience. Is there an example of this being done elsewhere in the codebase? It seems like event communication between sibling components can be a bit tricky.

chief8192:
Yeah, that worked out to be a lot cleaner and simpler. Closing this PR in favor of https://github.com/filebrowser/filebrowser/pull/1143

PR #1143, opened 2020-11-11
Automatically jumps to the next photo when deleting while previewing.

### LINKED ISSUE #1091: Go to next Image preview when deleting an Image

chief8192:
**Description**
Deletions made while previewing now immediately load the preview of the next file in the list. When the last file is deleted, previewing ends and returns to the file list, as before.

Closes #1091

### DIFF
--- a/frontend/src/components/files/Preview.vue
+++ b/frontend/src/components/files/Preview.vue
@@ -133,14 +133,14 @@ export default {
     }
   },
   async mounted () {
-    window.addEventListener('keyup', this.key)
+    window.addEventListener('keydown', this.key)
     this.$store.commit('setPreviewMode', true)
     this.listing = this.oldReq.items
-    this.$root.$on('preview_deleted', this.deleted)
+    this.$root.$on('preview-deleted', this.deleted)
     this.updatePreview()
   },
   beforeDestroy () {
-    window.removeEventListener('keyup', this.key)
+    window.removeEventListener('keydown', this.key)
     this.$store.commit('setPreviewMode', false)
   },
   methods: {
@@ -153,8 +153,10 @@ export default {
 
       if (this.hasNext) {
         this.next()
-      } else {
+      } else if (!this.hasPrevious && !this.hasNext == true) {
         this.back()
+      } else {
+        this.prev()
       }
     },
     back () {
@@ -169,7 +171,6 @@ export default {
       this.$router.push({ path: this.nextLink })
     },
     key (event) {
-      event.preventDefault()
 
       if (this.show !== null) {
         return
--- a/frontend/src/components/files/Preview.vue
+++ b/frontend/src/components/files/Preview.vue
@@ -142,18 +142,15 @@ export default {
   beforeDestroy () {
     window.removeEventListener('keydown', this.key)
     this.$store.commit('setPreviewMode', false)
+    this.$root.$off('preview-deleted', this.deleted)
   },
   methods: {
-    deleted (path) {
-
-      let pieces = path.split('/')
-      let deletedName = decodeURIComponent(pieces[pieces.length - 1])
-
-      this.listing = this.listing.filter(item => item.name !== deletedName)
+    deleted () {
+      this.listing = this.listing.filter(item => item.name !== this.name)
 
       if (this.hasNext) {
         this.next()
-      } else if (!this.hasPrevious && !this.hasNext == true) {
+      } else if (!this.hasPrevious && !this.hasNext) {
         this.back()
       } else {
         this.prev()
--- a/frontend/src/components/files/Preview.vue
+++ b/frontend/src/components/files/Preview.vue
@@ -136,6 +136,7 @@ export default {
     window.addEventListener('keyup', this.key)
     this.$store.commit('setPreviewMode', true)
     this.listing = this.oldReq.items
+    this.$root.$on('preview_deleted', this.deleted)
     this.updatePreview()
   },
   beforeDestroy () {
@@ -143,6 +144,19 @@ export default {
     this.$store.commit('setPreviewMode', false)
   },
   methods: {
+    deleted (path) {
+
+      let pieces = path.split('/')
+      let deletedName = decodeURIComponent(pieces[pieces.length - 1])
+
+      this.listing = this.listing.filter(item => item.name !== deletedName)
+
+      if (this.hasNext) {
+        this.next()
+      } else {
+        this.back()
+      }
+    },
     back () {
       this.$store.commit('setPreviewMode', false)
       let uri = url.removeLastDir(this.$route.path) + '/'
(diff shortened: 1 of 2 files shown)


o1egl:
@ramiresviana could you please review it?

frontend/src/components/prompts/Delete.vue
```diff
@@ -39,7 +38,8 @@ export default {
         if (!this.isListing) {
           await api.remove(this.$route.path)
           buttons.success('delete')
-          this.$router.push({ path: url.removeLastDir(this.$route.path) + '/' })
+
+          this.$root.$emit('preview_deleted', this.$route.path)
```
ramiresviana: Event names should be in kebab-case.

frontend/src/components/files/Preview.vue
```diff
@@ -136,13 +136,27 @@ export default {
     window.addEventListener('keyup', this.key)
     this.$store.commit('setPreviewMode', true)
     this.listing = this.oldReq.items
+    this.$root.$on('preview_deleted', this.deleted)
```
ramiresviana: To listen events of the component you can use the `v-on:preview-deleted` tag or its shorthand `@preview-deleted`.

frontend/src/components/prompts/Delete.vue
ramiresviana: `$root` should be avoided and used only on edge cases, use `this.$emit('preview-deleted')` instead.

frontend/src/components/files/Preview.vue
```diff
@@ -136,13 +136,27 @@ export default {
...
+
+      let pieces = path.split('/')
+      let deletedName = decodeURIComponent(pieces[pieces.length - 1])
+
+      this.listing = this.listing.filter(item => item.name !== deletedName)
+
+      if (this.hasNext) {
+        this.next()
+      } else {
+        this.back()
```
ramiresviana, line 154: `back()` closes the previewer, use `prev()` instead.

ramiresviana: When `!hasPrevious && !hasNext == true` the previewer should be closed.

ramiresviana (changes requested):
There is an issue when pressing enter on delete dialog, the `next()` is being triggered. This should be solved by:

- Moving `this.closeHovers()` to after `this.$emit` on https://github.com/filebrowser/filebrowser/blob/dcbc3286e20aca45a605a3cc8ba741f1fa251424/frontend/src/components/prompts/Delete.vue#L35
- Changing the `keyup` to `keydown` on https://github.com/filebrowser/filebrowser/blob/dcbc3286e20aca45a605a3cc8ba741f1fa251424/frontend/src/components/files/Preview.vue#L136-L142
- Removing `event.preventDefault()` on https://github.com/filebrowser/filebrowser/blob/dcbc3286e20aca45a605a3cc8ba741f1fa251424/frontend/src/components/files/Preview.vue#L158

chief8192:
The problem I'm running into is that after omitting `$root` from `this.$root.$emit(...)` and `this.$root.$on(...)`, event transmission stops working. Seems like it could be because `Delete` and `Preview` are sibling components?

ramiresviana:
Just omitting `$root` from `this.$root.$on(...)` is not going to work, you need to use `v-on` directive on the components you want to listen: https://vuejs.org/v2/guide/events.html

chief8192:
Yeah, so I've tried using `v-on`, in a number of different places. I can verify that the `this.$emit('preview-deleted')` call is working, because the event shows up in the devtools console when I click "DELETE".

The problem seems to be that `prompts/Delete.vue` is not a child component of `files/Preview.vue`. If I'm understanding [this article](https://www.telerik.com/blogs/how-to-emit-data-in-vue-beyond-the-vuejs-documentation) correctly, the `v-on` would have to be added to the `<delete>` tag that gets generated in `prompts/Prompts.vue`. It doesn't look like that could be done without a decent amount of refactoring, and even that wouldn't gain us the ability to interact with `Preview.vue`.

ramiresviana:
Every element that is on template body of the component is considered a child, so `<delete-button>` is child of the preview. A working example is the `<preview-size-button>`, it has `@change-size` event and is being used on the preview.

chief8192:
I agree, what you're saying is true. The difference here is that while the Delete **button** is a child of the Preview component, the Delete **prompt** is not. And it's the Delete **prompt** which is ultimately responsible for making the API call to delete the file, and then calling `$emit()`.

So given the limitation here that an event emitted by a child component has to be handled by the parent component (either via `v-on` or `@event-name`), I'm not clear how this could be easily made to work, given that the parent component of the Delete prompt is `Prompts.vue`, not `Preview.vue`:

filebrowser_components

ramiresviana:
I see now, since VueJS does not provide an proper solution for that case, i think we can use the `$root` instance.

chief8192:
Ok, committed a new round of changes

frontend/src/components/files/Preview.vue
```diff
@@ -133,16 +133,32 @@ export default {
...
   },
   async mounted () {
-    window.addEventListener('keyup', this.key)
+    window.addEventListener('keydown', this.key)
     this.$store.commit('setPreviewMode', true)
     this.listing = this.oldReq.items
+    this.$root.$on('preview-deleted', this.deleted)
     this.updatePreview()
   },
   beforeDestroy () {
```
ramiresviana, line 142: Use `this.$root.$off('preview-deleted', this.deleted)` to stop listening the event.

frontend/src/components/files/Preview.vue
```diff
@@ -133,16 +133,32 @@ export default {
...
+    window.removeEventListener('keydown', this.key)
     this.$store.commit('setPreviewMode', false)
   },
   methods: {
+    deleted (path) {
+
+      let pieces = path.split('/')
+      let deletedName = decodeURIComponent(pieces[pieces.length - 1])
+
+      this.listing = this.listing.filter(item => item.name !== deletedName)
```
ramiresviana:
```suggestion
      this.listing = this.listing.filter(item => item.name !== this.name)
```

frontend/src/components/files/Preview.vue
```diff
@@ -133,16 +133,32 @@ export default {
...
+
+      if (this.hasNext) {
+        this.next()
+      } else if (!this.hasPrevious && !this.hasNext == true) {
```
ramiresviana:
```suggestion
      } else if (!this.hasPrevious && !this.hasNext) {
```

ramiresviana (changes requested):
Navigation stops working when renaming file on preview, can you take a look on that? Similarly to the `prompts/Delete` you will need to create an event for `prompts/Rename`.

chief8192:
I did a clean build from master, and the rename issue is present there as well, so it looks to be an unrelated issue out of scope for this PR.

ramiresviana (approved):
Looking good, thank you for contributing to this project.

chief8192:
Cheers, happy to help!

PR #1148, opened 2020-11-14
feat: implemented hideDotfiles feature in the backend

kwsp:
This PR addresses https://github.com/filebrowser/filebrowser/issues/1084.

Per @o1egl, I've rewritten my previous PR (https://github.com/filebrowser/filebrowser/pull/1085) to move the filtering logic to the backend. 

By default the `hideDotfiles` options is turned off, and it can be turned on in profile settings.

### DIFF
--- a/rules/rules.go
+++ b/rules/rules.go
@@ -22,7 +22,7 @@ type Rule struct {
 // MatchHidden matches paths with a basename
 // that begins with a dot.
 func MatchHidden(path string) bool {
-	return filepath.Base(path)[0] == '.'
+	return strings.HasPrefix(filepath.Base(path), ".")
 }
 
 // Matches matches a path against a rule.
--- a/rules/rules.go
+++ b/rules/rules.go
@@ -1,6 +1,7 @@
 package rules
 
 import (
+	"path/filepath"
 	"regexp"
 	"strings"
 )
@@ -18,9 +19,10 @@ type Rule struct {
 	Regexp *Regexp `json:"regexp"`
 }
 
-var MatchHidden = Rule{
-	Regex:  true,
-	Regexp: &Regexp{Raw: `\/\..*$`},
+// MatchHidden matches paths with a basename
+// that begins with a dot.
+func MatchHidden(path string) bool {
+	return filepath.Base(path)[0] == '.'
 }
 
 // Matches matches a path against a rule.
--- a/rules/rules.go
+++ b/rules/rules.go
@@ -18,6 +18,11 @@ type Rule struct {
 	Regexp *Regexp `json:"regexp"`
 }
 
+var MatchHidden = Rule{
+	Regex:  true,
+	Regexp: &Regexp{Raw: `\/\..*$`},
+}
+
 // Matches matches a path against a rule.
 func (r *Rule) Matches(path string) bool {
 	if r.Regex {
--- a/rules/rules_test.go
+++ b/rules/rules_test.go
@@ -16,8 +16,8 @@ func TestMatchHidden(t *testing.T) {
 
 	for path, want := range cases {
 		got := MatchHidden(path)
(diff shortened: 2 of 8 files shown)


rules/rules.go
```diff
@@ -18,6 +18,11 @@ type Rule struct {
 	Regexp *Regexp `json:"regexp"`
 }
 
+var MatchHidden = Rule{
+	Regex:  true,
+	Regexp: &Regexp{Raw: `\/\..*$`},
+}
```
o1egl: Let's use path prefix instead of regexp

o1egl (changes requested):

rules/rules.go
kwsp: This needs to check if the first character of the base name is a dot. Does not work with `strings.HasPrefix`. Would still need a separate function that uses `filepath.Base`. Is that cool?

o1egl: Yes you are right, better would be to get the base name and then check if it starts from `.`

kwsp:
The path provided to the function are relative paths with directory prefixes, so `strings.HasPrefix("/a/b/c/.docker.json", "/.”)` does not work and we’d need the base name. That’s just an extra import and sort of out of line with the rest of the rules because they’re all Regex, but if you wanna do let this way, I don’t mind.

o1egl:
Prefix based comparison is much faster than the regexp.

rules/rules.go
kwsp: I've updated the PR.

rules/rules.go
```diff
@@ -18,6 +19,12 @@ type Rule struct {
 	Regexp *Regexp `json:"regexp"`
 }
 
+// MatchHidden matches paths with a basename
+// that begins with a dot.
+func MatchHidden(path string) bool {
+	return filepath.Base(path)[0] == '.'
```
o1egl: It will cause panic on empty string. 
```suggestion
	return strings.HasPrefix(filepath.Base(path), ".")
```

rules/rules_test.go
```diff
@@ -0,0 +1,23 @@
...
+		"/src":                false,
+		"/src/":               false,
+		"/.circleci":          true,
+		"/a/b/c/.docker.json": true,
+		".docker.json":        true,
+		"Dockerfile":          false,
+		"/Dockerfile":         false,
+	}
+
+	for path, truth := range cases {
```
o1egl:
```suggestion
	for path, want := range cases {
```

o1egl (commented):
nit

rules/rules_test.go
```diff
@@ -0,0 +1,23 @@
...
+		"/a/b/c/.docker.json": true,
+		".docker.json":        true,
+		"Dockerfile":          false,
+		"/Dockerfile":         false,
+	}
+
+	for path, want := range cases {
+		got := MatchHidden(path)
+		if got != truth {
+			t.Errorf("MatchHidden(%s)=%v; want %v", path, got, truth)
```
o1egl:
```suggestion
		if got != want {
			t.Errorf("MatchHidden(%s)=%v; want %v", path, got, want)
```

kwsp:
Any more changes needed?

o1egl (approved):

smithzss:
@NieTiger @o1egl Is this a configurable option when launching in a docker container?  I looked in the source and I didn't see anything that would suggest it, but I wanted to ask to confirm.  Didn't see it in the documentation as well.  I am looking for dotfiles to be hidden by default.  Appreciate the help!  Thanks.

o1egl:
@smithzss this param is only available in the UI.

pdesgarets:
@o1egl would you consider a PR implementing a flag for this param in the Cmd ?

PR #1168, opened 2020-11-23
Add ablity to disable file detection in listing

Namaneo:
Let's start from the beginning: there might be a better solution to the issue I solve with this PR.

In my situation, I'm accessing folders that contains lots of files (for instance, more than 1000 images). With the current version of filebrowser, it takes ages to display the content of this kind of folder, and I found out after some tests that it was due to the file type detection: after commenting the related lines, the folder appears as soon as I clicked on it.

This PR adds an option to disable this file detection on the listing, making the tool usable in this kind of situation. 🙂 However, I haven't digged that much into the guilty code: if you're not fine with those changes, I can take more time to understand exactly what's going one to find a better way!

Note that the configuration entry is only used for the listing, I believe it makes sense to keep it everywhere else (except for the deletion endpoint).

### DIFF
--- a/files/file.go
+++ b/files/file.go
@@ -163,10 +163,7 @@ func (i *FileInfo) detectType(modify, saveContent bool) error {
 	case strings.HasPrefix(mimetype, "image"):
 		i.Type = "image"
 		return nil
-	case (len(buffer) > 0 && isBinary(buffer)) || i.Size > 10*1024*1024: // 10 MB
-		i.Type = "blob"
-		return nil
-	default:
+	case strings.HasPrefix(mimetype, "text"):
 		i.Type = "text"
 
 		if !modify {
@@ -182,6 +179,9 @@ func (i *FileInfo) detectType(modify, saveContent bool) error {
 
 			i.Content = string(content)
 		}
+	default:
+		i.Type = "blob"
+		return nil
 	}
 
 	return nil
--- a/files/file.go
+++ b/files/file.go
@@ -144,29 +144,12 @@ func (i *FileInfo) detectType(modify, saveContent bool) error {
 	// of files couldn't be opened: we'd have immediately
 	// a 500 even though it doesn't matter. So we just log it.
 
-	readLen := 0
-	buffer := make([]byte, 0)
+	var buffer []byte
 
 	mimetype := mime.TypeByExtension(i.Extension)
 	if mimetype == "" {
-		reader, err := i.Fs.Open(i.Path)
-		if err != nil {
-			log.Print(err)
-			i.Type = "blob"
-			return nil
-		}
-		defer reader.Close()
-
-		buffer = make([]byte, 512)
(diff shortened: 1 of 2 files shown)


o1egl:
I think we shouldn't completely disable file type detections. It still can be done using file ext.

Namaneo:
Indeed, after some investigation, the guilty code was the file reading process, that was only necesserary if the file type could not be deduced from the extension. Now, files are read only when extension identification is failing.

I also removed the configuration entry, since this should transparently help for the vast majority of cases. 🙂

files/file.go
```diff
@@ -143,24 +143,29 @@ func (i *FileInfo) detectType(modify, saveContent bool) error {
...
 
-	buffer := make([]byte, 512)
-	n, err := reader.Read(buffer)
-	if err != nil && err != io.EOF {
-		log.Print(err)
-		i.Type = "blob"
-		return nil
-	}
+	readLen := 0
+	buffer := make([]byte, 0)
```
o1egl:
```suggestion
	var (
		readLen int
		buffer []byte
	)
```

files/file.go
```diff
@@ -143,24 +143,29 @@ func (i *FileInfo) detectType(modify, saveContent bool) error {
...
+
+		buffer = make([]byte, 512)
+		n, err := reader.Read(buffer)
+		if err != nil && err != io.EOF {
+			log.Print(err)
+			i.Type = "blob"
+			return nil
+		}
+
+		readLen = n
```
o1egl: Let's move this part into a separate function to reduce complexity

o1egl (changes requested):
Great job. Please fix linter errors

files/file.go
Namaneo: `readLen` was not required after all, I simplified by just using `buffer`

Namaneo:
With the latest updates, is everything fine in your opinion?

files/file.go
```diff
@@ -175,7 +163,7 @@ func (i *FileInfo) detectType(modify, saveContent bool) error {
 	case strings.HasPrefix(mimetype, "image"):
 		i.Type = "image"
 		return nil
-	case isBinary(buffer[:n], n) || i.Size > 10*1024*1024: // 10 MB
+	case (len(buffer) > 0 && isBinary(buffer)) || i.Size > 10*1024*1024: // 10 MB
```
o1egl: If the file has known extension and it is not video,audio or image and the size is  < 10mb, it will be considered as a text file which is not correct. The logic must be changed to `strings.HasPrefix(mimetype, "text")` with `blob` as a default type

Namaneo: Hey, sorry for the delay! I just updated the PR with your suggestion 🙂

o1egl:
Implemented in https://github.com/filebrowser/filebrowser/pull/1175

Issue #1112, opened 2020-10-07
Disable file header detection


agneevX:
~**Description**~
~I have a mounted cloud drive whose file/folder structure are cached in-memory.
However, FB will take very long to load them, especially if the folder contains many **files**...
For a folder which contains over a thousand sub-folders, it will load them instantly. This is weird.~

~**Expected behaviour**~
~Folders with many files and many sub-directories should load instantly and at the same speed.~

~**What is happening instead?**~
~Loading is slow and NGINX throws HTTP 504.~

---

**Is your feature request related to a problem? Please describe.**
I have cloud drives mounted as FUSE filesystems by rclone.

The entire file structure is entirely cached in-memory, so it would behave like any other filesystem... except File Browser tries to detect the file header of all files in a folder, and as a result, the screen is stuck on `Loading...`. 

As a side effect, a lot of simultaneous API calls are made to the cloud storage provider, causing HTTP 429s.

**Describe the solution you'd like**
Disable file header detection and load files, just like `ls`.

**Describe alternatives you've considered**
None. I can't browse any folder with multiple video files because they're taking too long to load.

**Additional context**
Thumbnails are disabled.

o1egl:
I'm not sure that this issue is related to the filebrowser. It uses standard system syscalls to access filesystem. Supporting cloud storages is out of the scope of this project.

agneevX:
> Supporting cloud storages is out of the scope of this project.

That is not related as it behaves like a normal FS. Is there a way to disable what it tries to access in video files?

o1egl:
No, it's not possible to filter out some files from listing. Can you compare the speed with [nginx file listing](http://nginx.org/en/docs/http/ngx_http_autoindex_module.html) for instance?

agneevX:
Not that. What I'm able to gather is that FB queries video files for certain info. 
I don't see any thumbnails generated but unsure whether its enabled. 

Its because of this querying that the process becomes slow.

o1egl:
It reads file header to [detect](https://github.com/filebrowser/filebrowser/blob/master/files/file.go#L137) it's type. That might be the bottleneck in your case.

agneevX:
Hmm. Maybe the detection could be done after the listing is complete?

agneevX:
I've updated the title and description @o1egl. Please take a look, thanks.

sergwx:
[...]

PR #1175, opened 2020-12-04
allow disabling file detections by reading header

WeidiDeng:
**Description**
This pr solves the [issue](https://github.com/filebrowser/filebrowser/issues/1112) and provides a frontend to set the options.

**Further comments**

:heart: Thank you!


o1egl:
This is actually a duplicate of https://github.com/filebrowser/filebrowser/pull/1168

WeidiDeng:
but that pr doesn't have the updated vue ui, no option exposed yet

o1egl:
I don't think that we need any ui for that. Configuring this option per user doesn't make sense.

WeidiDeng:
Perhaps this can be made in a global option, the one that uses commandline to set options. An array of relatives path to root, that skips type detection.

I tried the other [pr](https://github.com/filebrowser/filebrowser/pull/1168). Mostly it works fine, however I have a directory full of subtitle files (sass extension), the mime type detection fails, directory become very slow to show.

files/file.go
```diff
@@ -261,9 +262,13 @@ func (i *FileInfo) readListing(checker rules.Checker) error {
...
-			err := file.detectType(true, false)
-			if err != nil {
-				return err
+			if detectDisabled {
+				file.Type = "blob"
+			} else {
+				err := file.detectType(true, false)
+				if err != nil {
+					return err
+				}
```
o1egl: We should only disable file detection by header. Simple ext based detection shouldn't be disabled.

o1egl (changes requested):
In my opinion it doesn't make sense to disable it for specific paths only. It should be a global single boolean param.

files/file.go
WeidiDeng: Now I incorporate the changes from [pr](https://github.com/filebrowser/filebrowser/pull/1168).

rules/rules.go
```diff
@@ -9,6 +9,7 @@ import (
 // Checker is a Rules checker.
 type Checker interface {
 	Check(path string) bool
+	ReadHeader() bool
```
o1egl: I don't think that it is the right place to add this method

files/file.go
```diff
@@ -261,7 +269,7 @@ func (i *FileInfo) readListing(checker rules.Checker) error {
 		} else {
 			listing.NumFiles++
 
-			err := file.detectType(true, false)
+			err := file.detectType(true, false, checker.ReadHeader())
```
o1egl: Better would be to add a boolean param to the readListing func

o1egl (changes requested):

o1egl:
Instead of extending Checker interface you can add a boolean param to the FileOptions struct.

o1egl (approved):

PR #1184, opened 2020-12-11
download shared subdirectory

### LINKED ISSUE #1162: some suggestions
1、The shared folder cannot preview the file list. It can only be downloaded

2、Non administrator users cannot have multiple folder permissions

WeidiDeng:
**Description**
Currently shared folder can't be navigated like a real folder, only direct subdirectory and files can be viewed, but not downloadable separately.

This pr fixes [issue](https://github.com/filebrowser/filebrowser/issues/1162#issuecomment-734709064). Folders can be navigated and select files can be downloaded.

**Further comments**

:heart: Thank you!


o1egl (commented):
I found a bug that when you double click on a file which has `#` in its name the blank page is displayed.

ramiresviana:
- The information panel has some inconsistency when selecting a list item, only part of its information is being updated, that can cause user confusion about which item will actually be downloaded.
- The "Select multiple" feels out of place, lets move it to the header bar.
- The current routing allows to insert an arbitrary string to the path, like /share/I2cD5Vvw/\*anything\*. Also i think we should omit the root directory from the url.

WeidiDeng:
> * The information panel has some inconsistency when selecting a list item, only part of its information is being updated, that can cause user confusion about which item will actually be downloaded.

Which parts should be updated?



> * The "Select multiple" feels out of place, lets move it to the header bar.

I plan to make it more like the multiple style in the file listing page when I have time



> * The current routing allows to insert an arbitrary string to the path, like /share/I2cD5Vvw/*anything*. Also i think we should omit the root directory from the url.

Abitratry path can be detected in the backend, I don't think that's an issue, normal user won't bother changing url by hand.

The root directory potion is automatically filled. It's actually essential. According to [this comment](https://github.com/filebrowser/filebrowser/blob/bc4a6462ced43a1f754b8ad4aae2d770a1750051/http/public.go#L43). Say we share a directory a, there are dir a, b, c below it. When a request for /api/public/dl/hash/a came, which a is it, the root a or a below it, removing the trailing filename would fix it, but that leads to ugly download url a/a, 1.mp4/mp4. I did it in the previous commit. Also this means err check can't distinguish between two file with same name.

If root directory is only not shown in the address bar, there is quite a lot to do when someone use the subdirectory url to access the subdirectory, thanks to [this comment](https://github.com/filebrowser/filebrowser/blob/bc4a6462ced43a1f754b8ad4aae2d770a1750051/http/public.go#L43). Not appending filename when downloading can fix them pretty easily.

I think if root directory name is a thing, it can be checked. And normally user won't bother tickering with urls.

ramiresviana:
I see 2 ways to resolve the info panel issue:

1. Update the icon, displayName, lastModified, size, downloadButton and qrcode when selecting an item. Also the panel should have a different state for multiple selections, showing only number of selected items and total size.
2. The panel will show only the information of the current directory/file. Downloading selected items is done by clicking on a button on the header bar, similar to the normal Filebrowser listing.

frontend/src/components/Header.vue
```diff
@@ -49,6 +49,17 @@
           </button>
         </div>
       </template>
+      <template v-if="isSharing">
```
ramiresviana: Instead of creating a whole new dropdown, lets use the `v-show` directive on the wanted elements.

frontend/src/components/Header.vue
```diff
@@ -49,6 +49,17 @@
...
+        <!-- This buttons are shown on a dropdown on mobile phones -->
+        <div class="dropdown" :class="{ active: showMore }">
+          <download-button v-if="sharedSelectedCount > 0"></download-button>
+
+          <button @click="toggleSharedMultipleSelection" :aria-label="$t('buttons.selectMultiple')" :title="$t('buttons.selectMultiple')" class="action" >
```
ramiresviana: Can we reuse the existing selection toggle and download buttons?

frontend/src/store/index.js
```diff
@@ -26,6 +26,9 @@ const state = {
   showConfirm: null,
   previewMode: false,
   shared: {
+    req: {},
```
ramiresviana: I think we can reuse the current existing state properties. Also you can make the hash work as a getter using only the req state.

ramiresviana (changes requested):

PR #1187, opened 2020-12-15
Add LDAP support

### LINKED ISSUE #1164: LDAP Integration
It'd be great to have LDAP integration with local Active Directories so we only have to set up the permissions on filebrowser' side but set an OU to have access.

defkev:
Adds LDAP as an authentication/authorization backend using go-ldap/ldap
Profiles are still stored in bolt to keep things simple and avoid the need to alter the AD scheme.

This works by BINDING the user to the configured `ldap.userou` and, if successful, fetch the matching profile, if the profile doesn't exist it will be created using the configured DEFAULTS.

Access can be further restricted by setting the `ldap.usercn` which will limit access to members of this group.
Admin permission (in File Browser) is controlled by the `ldap.admincn` group which defaults to `file_browser_admins` you will have to create this group under `ldap.groupou` to use this feature.
Note that this only works if your AD server supports the `memberOf` attribute.

This is the most straight forward approach for people who simply want to authenticate File Browser against LDAP without the need to add custom objectClasses, attributesTypes and ACLs.

To get started simple set the `auth.method` to `ldap` and change the defaults as needed, see `config set --help` for all available `--ldap.*` options.

e.g. `config set --auth.method=ldap --ldap.starttls=false --ldap.basedn=dc=filebrowser,dc=org`

Which will connect to the LDAP server running on localhost skipping StartTLS.

Note that you will always need to pass all changed `--ldap.*` options to `config set --auth.method=ldap` or they will be reset to their respective defaults because of #715

There is a special option called `ldap.userhome` which, if set, will be used as the users scope.
By default this is set to the `homeDirectory` attribute, which in my case is pointing to a users config directory since my users don't have local accounts.

You can change this to anything you like as long as it is a TEXT attribute including adding a dedicated attributeType (e.g. fileBrowserScope) using a custom scheme.

Changes to admin permission and user scope are synchronized on every login whenever they change from LDAP to File Browser meaning you CANNOT change any LDAP attributes in File Browser for the sake of the Directory Managers sanity.
e.g. a user previously member of `ldap.admincn` will lose its admin permission on the next login if removed from the group and vice versa.
Same with `ldap.userhome`
The user password change in file browser is disabled by default during profile creation since the (File Browser) password isn't used.

This certainly needs testing, especially against different LDAP servers/version as i only tested it with DS389 (in development) and AD2012 (in production)

Closes #1164 (among other already closed LDAP feature requests)

### DIFF
--- a/auth/auth.go
+++ b/auth/auth.go
@@ -4,12 +4,13 @@ import (
 	"net/http"
 
 	"github.com/filebrowser/filebrowser/v2/users"
+	"github.com/filebrowser/filebrowser/v2/settings"
 )
 
 // Auther is the authentication interface.
 type Auther interface {
 	// Auth is called to authenticate a request.
-	Auth(r *http.Request, s *users.Storage, root string) (*users.User, error)
+	Auth(r *http.Request, s *users.Storage, set *settings.Storage, root string) (*users.User, error)
 	// LoginPage indicates if this auther needs a login page.
 	LoginPage() bool
 }
--- a/auth/json.go
+++ b/auth/json.go
@@ -26,7 +26,7 @@ type JSONAuth struct {
 }
 
 // Auth authenticates the user via a json in content body.
-func (a JSONAuth) Auth(r *http.Request, sto *users.Storage, root string) (*users.User, error) {
+func (a JSONAuth) Auth(r *http.Request, sto *users.Storage, set *settings.Storage, root string) (*users.User, error) {
 	var cred jsonCred
 
 	if r.Body == nil {
--- a/auth/ldap.go
+++ b/auth/ldap.go
@@ -1,0 +1,158 @@
+package auth
+
+import (
+    "net/http"
+    "encoding/json"
+    "crypto/tls"
+    "fmt"
+    "strings"
+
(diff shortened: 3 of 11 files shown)


o1egl:
LDAP support is out of the scope of this project. Let's keep it simple

changchunhua2017:
i like it ，Please consider adding this feature

matthias-colt:
It's a pity that the PR has need rejected, LDAP is something quite essential in my view since local user management is not sustainable.

gaby:
@defkev I'm trying to create a fork with latest `filebrowser` and the changes from this PR but it's failing. is that something you would be interesting in helping? 

The idea of the new fork is to stay up to date with upstream plus add support for LDAP as this PR did and several people have requested.

gaby:
@o1egl After two years, is LDAP still considered out of scope for this project? or should I go ahead with forking `filebrowser` and adding LDAP support?

gaby:
@defkev I'm trying to merge this PR into a fork but it's failing on this line: 

```
Error: http/auth.go:108:37: cannot use d.store.Users (type users.Store) as type *users.Storage in argument to auther.Auth: need type assertion
```

Everything else merged fine.

PR #1224, opened 2021-01-04
optimize gallery preview

WeidiDeng:
**Description**
Currently filebrowser uses the same preview for image with blob, audio, video.
Separate image preview like a gallery, and optimize [interaction](https://github.com/filebrowser/filebrowser/discussions/1200).

**Further comments**

:heart: Thank you!


frontend/src/components/files/ExtendedImage.vue
```diff
@@ -79,12 +82,68 @@ export default {
...
+      this.doMove(target[0], target[1])
+    },
+    fitScreenTarget() {
+      if (this.scale <= this.minScale) {
+        let style = this.$refs.imgex.style
+        let posX = this.pxStringToNumber(style.left)
+        let posY = this.pxStringToNumber(style.top)
+        return [this.position.center.x - posX, this.position.center.y - posY]
+      }
+      else {
```
o1egl, line 129: else is redundant since you return in upper if

WeidiDeng: When image is zoomed, else comes into play.

ramiresviana:
Found some issues with this PR:

- Similar gestures for different actions, the swipe right/left is used to navigate between images and also to move the preview. When doing a swipe you may end navigating or moving the preview, making it unreliable.
- Button icons still can be clicked when the header bar is not visible.
- Header bar stays indefinitely when clicking on a button of the bar.
- Position is reset when moving the image out of the screen, even just by a bit. When doing that on top of the image the preview moves to the bottom.
- The header bar is automatically hidden when entering on preview, that can cause confusion to lay users. Also to show the header on mobile, is required to do a press exactly on the header position.

niubility000:
*Position is reset when moving the image out of the screen, even just by a bit. When doing that on top of the image the preview moves to the bottom.

Just change to the following in ExtendedImage.vue:
 // right out of viewport
        else if (rect.left > 0 && rect.right > width) x = Math.max(-rect.left, width - rect.right)
// bottom out of viewport
        else if (rect.top > 0 && rect.bottom > height) y = Math.max(-rect.top, height - rect.bottom)
*The header bar is automatically hidden when entering on preview, that can cause confusion to lay users. Also to show the header on mobile, is required to do a press exactly on the header position.

Yes, I also think the header bar should be visable all the time.  Just make it transparent. we can see the buttons clearly. （like what they did in Nextcloud.）

niubility000:
> Found some issues with this PR:
> 
> * Similar gestures for different actions, the swipe right/left is used to navigate between images and also to move the preview. When doing a swipe you may end navigating or moving the preview, making it unreliable.
> * Button icons still can be clicked when the header bar is not visible.
> * Header bar stays indefinitely when clicking on a button of the bar.
> * Position is reset when moving the image out of the screen, even just by a bit. When doing that on top of the image the preview moves to the bottom.
> * The header bar is automatically hidden when entering on preview, that can cause confusion to lay users. Also to show the header on mobile, is required to do a press exactly on the header position.

I modified some codes. Now, it works perfect. Just take a look. The head bar is at the bottom and transparent. The pics move as expected. 
filebrowser.zip is for windows. 

[gallery preview1.zip](https://github.com/filebrowser/filebrowser/files/5923021/gallery.preview1.zip)
[filebrowser.zip](https://github.com/filebrowser/filebrowser/files/5923885/filebrowser.zip)

 
The effect is still not as smooth as that one in  https://tank.eyeblue.cn/user/login (demo:123456) or http://demo.kodcloud.com/ （demo:demo）. Maybe we need add a little bit animation.

PR #1225, opened 2021-01-04
Add 3d model preview

vincent:
I plan to use File Browser to list my 3D model files, so I needed a way to preview them.

This PR uses https://github.com/hujiulong/vue-3d-model to preview such files.
filebrowser_3d_model_viewer

If you dont feel a need for this, feel free to discard the PR.


vincent:
> cyclomatic complexity 16 of func `(*FileInfo).detectType` is high (> 15)

Not sure how to fix this one, as I'm not very experienced in Go.
Also I believe this is subject to change in another PR

files/file.go
```diff
@@ -199,6 +202,21 @@ func (i *FileInfo) detectType(modify, saveContent bool) error {
 	return nil
 }
 
+func is3DModelFile(ext string) bool {
+	switch strings.ToLower(ext) {
+	case
+		".obj",
+		".json",
```
o1egl: json file should not be considered as a 3d object

o1egl (commented):
Great job. But I think all the handling can be done on the frontend by checking the file ext. Check out [.pdf](https://github.com/filebrowser/filebrowser/blob/master/frontend/src/components/files/Preview.vue#L53) handler

frontend/src/components/files/Preview.vue
```diff
@@ -230,6 +233,9 @@ export default {
     },
     toggleSize () {
       this.fullSize = !this.fullSize
+    },
+    is3DModelFile (req) {
+      return ThreeViewer.methods.is3DModelFile(req)
```
vincent: Is it the correct way to do this ?

o1egl:
Checkout my refactoring commit. I think it should look something like that. I tested Duck.gltf from examples and it doesn't work properly. it tries to access Duck.bin and DuckCM.png without adding an auth token.

vincent:
Thank you,

.stl and .fbx are working well it seems.
Others are returned as `type: text` however so they wont display.

edit: oh sorry I just get what you meant.
I'll try to check if there is way to workaround subfiles fetching with the token.

Here are some test files if needed: https://www.thingiverse.com/thing:4719541

newhinton:
@vincent Is it possible to extend this to gcode? There are some javascript-libraries that can render gcode, but i dont know how easy it would be to extend.

vincent:
I tried to update the PR with the latest master.

@newhinton I think that's a great idea for 3d models hobbyists, and it should not be that hard to implement.

That being said, I fear it drags unwanted features in the project.

Also, I'm not sure which lib to use
- https://github.com/Cloud-CNC/vue-gcode-viewer is archived
- its replacement https://github.com/cloud-cnc/vue-3d-viewer does not seems to drag much interest (0 issues)
- https://github.com/remcoder/gcode-preview seems nice, and the author has done [some Vue experiments](https://github.com/remcoder/gcode-preview-vue-dem) but could be a bit overkill for a preview

ping @o1egl

NickSica:
Was there a reason this was closed?

PR #1233, opened 2021-01-06
allow navigation information to be cached

WeidiDeng:
**Description**
Save navigation information.
Currently when navigating back from a preview, editor, the view will reset. This pr saves the last view navigation information.

**Further comments**

:heart: Thank you!

### DIFF
--- a/frontend/src/store/getters.js
+++ b/frontend/src/store/getters.js
@@ -13,6 +13,15 @@ const getters = {
 
     let sum = state.upload.progress.reduce((acc, val) => acc + val)
     return Math.ceil(sum / state.upload.size * 100);
+  },
+  getLastViewedDetail: state => (path) => {
+    if (state.lastViewed.details.has(path)) {
+      let i = state.lastViewed.paths.indexOf(path)
+      state.lastViewed.paths.splice(i, 1)
+      state.lastViewed.paths.push(path)
+      return state.lastViewed.details.get(path)
+    }
+    return null
   }
 }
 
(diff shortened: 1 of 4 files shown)


niubility000:
Wow, you really did what I need.
I tried it, and just found the the relocation is not very accurate. Only the first 15 lines works great. After that, it always shows the fifteen lines at the top.  if the folder with like 1000 pics in it, the relocation doesn't work then. I think maybe it's due to the Lazy loading.  Shall we just don't destroy the listing page （keep-alive ）. when we leave it. and we can return back directly （or use scrollBehavior ）.  

And , if in the previewer, the last img I browsered is the 30 line's img, when we return to the list, can we just relocate to the 30 line directly? It will be easier to do, and much useful.

ramiresviana:
For me that's is a very uncommon feature, i would expect to have the previewer state always reset. I don't know if that feature should be enabled by default.

niubility000:
> For me that's is a very uncommon feature, i would expect to have the previewer state always reset. I don't know if that feature should be enabled by default.

And when you browse a large folder with 1000pics, this cache function will be necessary.

niubility000:
Just change to the following in the files.vue. It works fine!
        for (let i = 0; i < 100000; i++) {
          if (this.req.items[i].name === detail.clicked) {
            this.$store.commit('addSelected', i)
            this.$store.commit('resetSelected')
            break

So other ways to to this: https://www.cnblogs.com/zypblogs/p/10481784.html

we also need this https://github.com/filebrowser/filebrowser/pull/1234, it's like "fast listing".

But, I think it's better to relocate to the pic we last browsed in the preview page, not this pic we clicked in this page.

niubility000:
Maybe we can add a choice box in personal settings: enable listing cache. It will be better.

helloray:
I really need this feature.

PR #1234, opened 2021-01-06
Add EXIF thumbnail support for JPEG files

adrium:
**Description**
Compact implementation to support EXIF thumbnails. Listing a directory with a lot of thumbnails will be a lot faster. The following was recorded with the [master branch](https://user-images.githubusercontent.com/23739584/103834336-662b8c80-5083-11eb-9d6d-26d892c20082.mp4) and with the [feature proposed](https://user-images.githubusercontent.com/23739584/103834358-7479a880-5083-11eb-8a0f-b1ef3cfff2d1.mp4) here. The app runs on a Raspberry Pi 4b accessing files on a hard disk.

- [x] continuous integration build passed


adrium:
```
img/service.go:226:11: ineffectual assignment to `err` (ineffassign)
					thm, err = index.RootIfd.NextIfd().Thumbnail()
					     ^
img/service.go:161:7: shadow: declaration of "err" shadows declaration at line 143 (govet)
			_, err := out.Write(thm)
			   ^
img/service.go:221:8: shadow: declaration of "err" shadows declaration at line 202 (govet)
			im, err := exifcommon.NewIfdMappingWithStandard()
			    ^
```

Those assignments are actually effective...

```
img/service.go:14: File is not `goimports`-ed with -local github.com/filebrowser/filebrowser (goimports)
	"github.com/dsoprea/go-exif/v3/common"
```

How can I fix the import error?

adrium:
Fixed CI, all set :rocket:

adrium:
Latest updated includes a fix for thumbnail distortion. `object-fit: cover` is added to stylesheet.
Also rebased to latest master.

niubility000:
It's pretty cool. very very fast and smooth!
But for some pics, it just didn't show the thumbnails. Would you please fix that. Maybe in this situation, just allow the filebrowser to generate the thumknails for those pics.  Here 's the screenshot and some that kind of pics.
[TMP.zip](https://github.com/filebrowser/filebrowser/files/5902262/TMP.zip)

adrium:
Thanks for reporting. Reason was because EXIF exists, but does not contain a thumbnail in the files provided.

This is fixed in the newest version of the pull request. Additionally, it was rebased onto master and added the `feat:` prefix to the commit message.

niubility000:
Wow, I checked it. Perfect!! very fast! especially for the folder with tons of pics. and it 's pretty cool for the low-end server!

niubility000:
Oops, I just found a small bug. For some pictures, It looks like the thumbnails are broken, only shows half of them. Would you please check out these pics? 
[temp.zip](https://github.com/filebrowser/filebrowser/files/5929775/temp.zip)

adrium:
Hi niubility, thanks for testing this PR so thoroughly.
Let's hope it will make into the master branch eventually.

The thumbnails in your files were larger than the preloaded size of 32k.
In the examples I had, a lot of thumbnails were around 16k that's why I chose the value of 32k.

Theoretically, Exif data (including the thumbnails) must fit in the APP1 marker and all JPEG markers can be max 64k.
The preload size is now `0xffff` and it will be hopefully be sufficiently large.

adrium:
Improved code style

niubility000:
Great! I have tried. It's wonderful!
and the file list's loading speed is amazing fast!  I don't need to wait like half a minute for that anymore.

img/service.go
```diff
@@ -183,3 +197,46 @@ func (s *Service) detectFormat(in io.Reader) (Format, io.Reader, error) {
...
 	return format, io.MultiReader(buf, in), nil
 }
+
+func getEmbeddedThumbnail(in io.Reader) ([]byte, io.Reader, error) {
+	buf := &bytes.Buffer{}
+	r := io.TeeReader(in, buf)
+	wrappedReader := io.MultiReader(buf, in)
+
+	offset := 0
+	offsets := []int{12, 30}
```
o1egl, line 207: Can you please explain why are you checking those offsets only?

o1egl (changes requested):
Please add tests

img/service.go
adrium: Those are the two most common offsets where the Exif starts in JPEG files. To avoid more complex parsing (and library dependencies), it only checks those offsets.

If no Exif is found, it will fall back to the downscaling method.

adrium:
Writing tests will require quite some effort... I will look into it when I have time

go.mod
```diff
@@ -10,6 +10,7 @@ require (
 	github.com/dgrijalva/jwt-go v3.2.0+incompatible
 	github.com/disintegration/imaging v1.6.2
 	github.com/dsnet/compress v0.0.1 // indirect
+	github.com/dsoprea/go-exif/v3 v3.0.0-20201216222538-db167117f483 // indirect
```
o1egl: Why not stick to version v3.0.0?

img/service.go
o1egl: I checked that in the provided library the different approach is used https://github.com/dsoprea/go-exif/blob/d154f10435cc/v3/exif.go#L52

adrium: Indeed. And that's the approach in the first (unpublished) attempt. That function tries to find an exif in brute-force fashion byte by byte scanning the whole file: https://github.com/dsoprea/go-exif/blob/d154f10435cc6ae4b64cd68eed04a1d005b73ec7/v3/exif.go#L150

It is more efficient by just checking the most common offsets for exif and there was a noticeable performance increase.

go.mod
adrium: Honestly no clue. That was added by `go get` automatically...

o1egl: Fix it, please

adrium: I looked into this and tried to stick to the version number you mentioned. However, the library does not seem provide tagged releases:

```
:~/Code/filebrowser*master$ go get -d github.com/dsoprea/go-exif/v3@v3.0.0
go get: github.com/dsoprea/go-exif/v3@v3.0.0: invalid version: unknown revision v3.0.0

:~/Code/filebrowser*master$ go get -d github.com/dsoprea/go-exif/v3@3.0.0
go: downloading github.com/dsoprea/go-exif v0.0.0-20200711180620-208788de2815
go get: module github.com/dsoprea/go-exif@3.0.0 found (v0.0.0-20200711180620-208788de2815), but does not contain package github.com/dsoprea/go-exif/v3
```

The library is pulled as recommended. Other projects use the library in a similar way. Examples are PhotoPrism (listed on [Awesome-Selfhosted](https://github.com/awesome-selfhosted/awesome-selfhosted)) and other projects on GitHub:

* [PhotoPrism](https://github.com/photoprism/photoprism/blob/develop/go.mod)
* [GoGallery](https://github.com/robrotheram/gogallery/blob/master/server/go.mod)
* [Collage](https://github.com/fawick/collage/blob/master/go.mod)
* [photoimportd](https://github.com/bwdezend/photoimportd/blob/master/go.mod)

As such, the dependency file should probably be kept as committed.

adrium:
> Please add tests

Constructing pictures programmatically requires a lot of code and is error-prone in itself.

On the other hand, the pictures provided by niubility000 are a very useful set and are "real" data.
I propose to use those pictures to create the tests (if the user agrees). The extracted thumbnails could be compared to the expected hash values.

Do you prefer a specific structure to check in the files?

o1egl:
@adrium I agree that using image files will be better for that. If you plan to add table tests, please look at https://github.com/filebrowser/filebrowser/blob/master/img/service_test.go for a reference.

adrium:
@o1egl Thanks a lot for your feedback.

@niubility000 Do you agree to make use of the pictures you provided for testing and thereby checking them in into this repo?

adrium:
Tests added. Image credits:

* **20130612_142406.jpg** @niubility000 assumed usage permission because of upload in this PR
* **IMG_2578.JPG** @niubility000 assumed usage permission because of upload in this PR
* **gray-sample.jpg** https://github.com/haraldk/TwelveMonkeys

o1egl (approved):
LGTM

niubility000:
I just found a bug: #2137

PR #1252, opened 2021-01-25
Allow to password protect shares

alvaroaleman:
This changes allows to password protect shares. It works by:
* Allowing to optionally pass a password when creating a share
* If set, the password + salt that is configured via a new flag will be
  hashed via bcrypt and the hash stored together with the rest of the
  share
* Additionally, a random 96 byte long token gets generated and stored
  as part of the share
* When the backend retrieves an unauthenticated request for a share
   that has authentication configured, it will return a http 401
* The frontend detects this and will show a login prompt
* The actual download links are protected via an url arg that contains
  the previously generated token. This allows us to avoid buffering the
  download in the browser and allows pasting the link without breaking
  it

Thanks a lot for all the existing work on filebrowser btw, it's great!

Fixes #1223

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,7 +2,7 @@ version: 2
 jobs:
   lint:
     docker:
-      - image: golangci/golangci-lint:v1.27.0
+      - image: golangci/golangci-lint:v1.31.0
     steps:
       - checkout
       - run: golangci-lint run -v
--- a/frontend/src/api/files.js
+++ b/frontend/src/api/files.js
@@ -77,8 +77,13 @@ export function download (format, ...files) {
   if (format !== null) {
     url += `algo=${format}&`
   }
+  if (store.state.jwt !== ''){
+    url += `auth=${store.state.jwt}&`
+  }
+  if (store.state.token !== ''){
+    url += `token=${store.state.token}`
+  }
 
-  url += `auth=${store.state.jwt}`
   window.open(url)
 }
 
--- a/frontend/src/api/share.js
+++ b/frontend/src/api/share.js
@@ -4,8 +4,10 @@ export async function list() {
   return fetchJSON('/api/shares')
 }
 
-export async function getHash(hash) {
-  return fetchJSON(`/api/public/share/${hash}`)
+export async function getHash(hash, password = "") {
+  return fetchJSON(`/api/public/share/${hash}`, {
+    headers: {'X-SHARE-PASSWORD': password},
+  })
 }
 
 export async function get(url) {
@@ -23,14 +25,18 @@ export async function remove(hash) {
   }
 }
 
-export async function create(url, expires = '', unit = 'hours') {
+export async function create(url, password = '', expires = '', unit = 'hours') {
   url = removePrefix(url)
   url = `/api/share${url}`
   if (expires !== '') {
     url += `?expires=${expires}&unit=${unit}`
   }
-
+  let body = '{}';
+  if (password != '' || expires !== '' || unit !== 'hours') {
+    body = JSON.stringify({password: password, expires: expires, unit: unit})
+  }
   return fetchJSON(url, {
-    method: 'POST'
+    method: 'POST',
+    body: body,
   })
 }
--- a/frontend/src/components/prompts/Share.vue
+++ b/frontend/src/components/prompts/Share.vue
@@ -1,14 +1,11 @@
 <template>
-  <div class="card floating" id="share">
+  <div class="card floating share__promt__card" id="share">
     <div class="card-title">
       <h2>{{ $t('buttons.share') }}</h2>
     </div>
 
     <div class="card-content">
       <ul>
-        <li v-if="!hasPermanent">
-          <a @click="getPermalink" :aria-label="$t('buttons.permalink')">{{ $t('buttons.permalink') }}</a>
-        </li>
 
         <li v-for="link in links" :key="link.hash">
           <a :href="buildLink(link.hash)" target="_blank">
@@ -27,6 +24,13 @@
             :title="$t('buttons.copyToClipboard')"><i class="material-icons">content_paste</i></button>
         </li>
 
+        <li v-if="!hasPermanent">
+          <div>
+            <input type="password" :placeholder="$t('prompts.optionalPassword')" v-model="passwordPermalink">
+            <a @click="getPermalink" :aria-label="$t('buttons.permalink')">{{ $t('buttons.permalink') }}</a>
+          </div>
+        </li>
+
         <li>
           <input v-focus
             type="number"
@@ -40,6 +44,7 @@
             <option value="hours">{{ $t('time.hours') }}</option>
             <option value="days">{{ $t('time.days') }}</option>
           </select>
+          <input type="password" :placeholder="$t('prompts.optionalPassword')" v-model="password">
           <button class="action"
             @click="submit"
             :aria-label="$t('buttons.create')"
@@ -72,7 +77,9 @@ export default {
       unit: 'hours',
       hasPermanent: false,
       links: [],
-      clip: null
+      clip: null,
+      password: '',
+      passwordPermalink: ''
     }
   },
   computed: {
@@ -121,7 +128,7 @@ export default {
       if (!this.time) return
 
       try {
-        const res = await api.create(this.url, this.time, this.unit)
+        const res = await api.create(this.url, this.password, this.time, this.unit)
         this.links.push(res)
         this.sort()
       } catch (e) {
@@ -130,7 +137,7 @@ export default {
     },
     getPermalink: async function () {
       try {
-        const res = await api.create(this.url)
+        const res = await api.create(this.url, this.passwordPermalink)
         this.links.push(res)
         this.sort()
         this.hasPermanent = true
--- a/frontend/src/i18n/en.json
+++ b/frontend/src/i18n/en.json
@@ -29,6 +29,7 @@
     "selectMultiple": "Select multiple",
     "share": "Share",
     "shell": "Toggle shell",
+    "submit": "Submit",
     "switchView": "Switch view",
     "toggleSidebar": "Toggle sidebar",
     "update": "Update",
@@ -142,7 +143,8 @@
     "show": "Show",
     "size": "Size",
     "upload": "Upload",
-    "uploadMessage": "Select an option to upload."
+    "uploadMessage": "Select an option to upload.",
+    "optionalPassword": "Optional password"
   },
   "search": {
     "images": "Images",
--- a/frontend/src/views/Share.vue
+++ b/frontend/src/views/Share.vue
@@ -74,6 +74,24 @@
   <div v-else-if="error">
     <not-found v-if="error.message === '404'"></not-found>
     <forbidden v-else-if="error.message === '403'"></forbidden>
+    <div v-else-if="error.message === '401'">
+      <div class="card floating" id="password">
+        <div v-if="attemptedPasswordLogin" class="share__wrong__password">{{ $t('login.wrongCredentials') }}</div>
+        <div class="card-title">
+          <h2>{{ $t('login.password') }}</h2>
+        </div>
+
+        <div class="card-content">
+          <input v-focus type="password" :placeholder="$t('login.password')" v-model="password" @keyup.enter="fetchData">
+        </div>
+        <div class="card-action">
+          <button class="button button--flat"
+            @click="fetchData"
+            :aria-label="$t('buttons.submit')"
+            :title="$t('buttons.submit')">{{ $t('buttons.submit') }}</button>
+        </div>
+      </div>
+    </div>
     <internal-error v-else></internal-error>
   </div>
 </template>
@@ -102,7 +120,9 @@ export default {
   data: () => ({
     error: null,
     path: '',
-    showLimit: 500
+    showLimit: 500,
+    password: '',
+    attemptedPasswordLogin: false
   }),
   watch: {
     '$route': 'fetchData'
@@ -129,7 +149,11 @@ export default {
       return 'insert_drive_file'
     },
     link: function () {
-      return `${baseURL}/api/public/dl/${this.hash}${this.path}`
+      let queryArg = '';
+      if (this.token !== ''){
+        queryArg = `?token=${this.token}`
+      }
+      return `${baseURL}/api/public/dl/${this.hash}${this.path}${queryArg}`
     },
     fullLink: function () {
       return window.location.origin + this.link
@@ -193,8 +217,13 @@ export default {
       this.error = null
 
       try {
-        let file = await api.getHash(encodeURIComponent(this.$route.params.pathMatch))
+        if (this.password !== ''){
+          this.attemptedPasswordLogin = true
+        }
+        let file = await api.getHash(encodeURIComponent(this.$route.params.pathMatch), this.password)
         this.path = file.path
+        this.token = file.token || ''
+        this.$store.commit('setToken', this.token)
         if (file.isDir) file.items = file.items.map((item, index) => {
           item.index = index
           item.url = `/share/${this.hash}${this.path}/${encodeURIComponent(item.name)}`
--- a/http/public.go
+++ b/http/public.go
@@ -1,14 +1,17 @@
 package http
 
 import (
+	"errors"
 	"net/http"
 	"path"
 	"path/filepath"
 	"strings"
 
 	"github.com/spf13/afero"
+	"golang.org/x/crypto/bcrypt"
 
 	"github.com/filebrowser/filebrowser/v2/files"
+	"github.com/filebrowser/filebrowser/v2/share"
 )
 
 var withHashFile = func(fn handleFunc) handleFunc {
@@ -17,6 +20,11 @@ var withHashFile = func(fn handleFunc) handleFunc {
 		link, err := d.store.Share.GetByHash(id)
 		if err != nil {
 			return errToStatus(err), err
+		}
+
+		status, err := authenticateShareRequest(r, link)
+		if status != 0 || err != nil {
+			return status, err
 		}
 
 		user, err := d.store.Users.Get(d.server.Root, link.UserID)
@@ -33,6 +41,7 @@ var withHashFile = func(fn handleFunc) handleFunc {
 			Expand:     true,
 			ReadHeader: d.server.TypeDetectionByHeader,
 			Checker:    d,
+			Token:      link.Token,
 		})
 		if err != nil {
 			return errToStatus(err), err
@@ -48,6 +57,7 @@ var withHashFile = func(fn handleFunc) handleFunc {
 				Modify:  d.user.Perm.Modify,
 				Expand:  true,
 				Checker: d,
+				Token:   link.Token,
 			})
 			if err != nil {
 				return errToStatus(err), err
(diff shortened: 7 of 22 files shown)


http/public.go
```diff
@@ -94,3 +104,26 @@ var publicDlHandler = withHashFile(func(w http.ResponseWriter, r *http.Request,
...
+	password := r.Header.Get("X-SHARE-PASSWORD")
+	if password == "" {
+		return http.StatusUnauthorized, errors.New("unauthorized")
+	}
+	if err := bcrypt.CompareHashAndPassword([]byte(l.PasswordHash), []byte(password+salt)); err != nil {
+		if errors.Is(err, bcrypt.ErrMismatchedHashAndPassword) {
+			return http.StatusUnauthorized, nil
+		}
+		return http.StatusBadRequest, err
+	}
```
ramiresviana: No errors are being returned.

frontend/src/views/Share.vue
```diff
@@ -74,6 +74,23 @@
...
     <not-found v-if="error.message === '404'"></not-found>
     <forbidden v-else-if="error.message === '403'"></forbidden>
+    <div v-else-if="error.message === '401'">
+      <div class="card floating" id="password">
+        <div class="card-title">
+          <h2>{{ $t('buttons.password') }}</h2>
+        </div>
+
+        <div class="card-content">
+          <input v-focus type="password" placeholder="Password" v-model="password" @keyup.enter="fetchData">
```
ramiresviana: The `placeholder` attribute should use localized value.

frontend/src/components/prompts/Share.vue
```diff
@@ -1,14 +1,11 @@
 <template>
-  <div class="card floating" id="share">
+  <div class="card floating" style="max-width: max-content; width: auto" id="share">
```
ramiresviana: Let's avoid inline styling.

frontend/src/components/prompts/Share.vue
```diff
@@ -27,6 +24,13 @@
             :title="$t('buttons.copyToClipboard')"><i class="material-icons">content_paste</i></button>
         </li>
 
+        <li v-if="!hasPermanent">
+          <div style="text-align: right">
```
ramiresviana: Let's avoid inline styling.

ramiresviana (changes requested):
This PR is not working correctly as the way it is, you need to do some testing. Also found some issues:

- Unprotected shares cannot be accessed.
- There's no feedback when user type password wrong.

alvaroaleman:
> This PR is not working correctly as the way it is, you need to do some testing

@ramiresviana can you elaborate on that, what exactly is not working correctly? Or do you just mean the two issues you mentioned?

http/public.go
alvaroaleman: Sorry, I don't understand the comment. Do you mean that I need to check for a non-zero statuscode above?

ramiresviana: When catching the `bcrypt.ErrMismatchedHashAndPassword` error, `nil` are being returned instead of `err`, allowing the request to be processed. That way protected shares can be accessed by providing a wrong password.

ramiresviana:
> > This PR is not working correctly as the way it is, you need to do some testing
> 
> @ramiresviana can you elaborate on that, what exactly is not working correctly? Or do you just mean the two issues you mentioned?

The most important aspect of this feature was not working, the password validation. On my tests i was able to access any share by providing a wrong password. Also found another issue:

- Downloading selected files, using the select button on the header bar, is not possible. Looks like the download url is missing the token.

alvaroaleman:
@ramiresviana addressed all feedback:
* Passwords are now properly checked
* Also added some tests for that
* Shares that are created without password now end up actually not having a password
* Entering a wrong password will result in an error being shown
* Selecting files and then downloading them via the button at the top works
* No inline styling
* Localized placeholder

Please have another look.

http/public.go
alvaroaleman: An error here only indicates that there was an error doing the check, but you are right, I didn't check the returned status code of this func

frontend/src/views/Share.vue
```diff
@@ -129,7 +148,7 @@ export default {
       return 'insert_drive_file'
     },
     link: function () {
-      return `${baseURL}/api/public/dl/${this.hash}${this.path}`
+      return `${baseURL}/api/public/dl/${this.hash}${this.path}?token=${this.token}`
```
ramiresviana: Let's omit the token parameter from the url on non-protected shares.

frontend/src/views/Share.vue
```diff
@@ -193,8 +212,14 @@ export default {
       this.error = null
 
       try {
-        let file = await api.getHash(encodeURIComponent(this.$route.params.pathMatch))
+        const password = this.password || ''
+        if (password !== ''){
+          this.attemptedPasswordLogin = true
```
ramiresviana, line 221: `attemptedPasswordLogin` need to be added to the component state.

frontend/src/views/Share.vue
```diff
@@ -193,8 +212,14 @@ export default {
       this.error = null
 
       try {
-        let file = await api.getHash(encodeURIComponent(this.$route.params.pathMatch))
+        const password = this.password || ''
```
ramiresviana: `password` need to be added to the component state.

frontend/src/views/Share.vue
```diff
@@ -193,8 +212,14 @@ export default {
...
+        if (password !== ''){
+          this.attemptedPasswordLogin = true
+        }
+        let file = await api.getHash(encodeURIComponent(this.$route.params.pathMatch), password)
         this.path = file.path
+        this.token = file.token || ''
+        store.commit('setToken', this.token)
```
ramiresviana: That application store state is not being used, should be removed.

frontend/src/components/prompts/Share.vue
```diff
@@ -130,7 +135,7 @@ export default {
     },
     getPermalink: async function () {
       try {
-        const res = await api.create(this.url)
+        const res = await api.create(this.url, this.passwordpermalink)
```
ramiresviana: `passwordpermalink` need to be added to the component state.

ramiresviana (changes requested):
What about using JWT based authentication? I think would be a more secure approach.

frontend/src/views/Share.vue
alvaroaleman: It is, it is used in the JS that generates the downloadlink when you select a file and click download in the top bar

frontend/src/components/prompts/Share.vue
alvaroaleman: Can you give an example of what you mean by "component state"?

alvaroaleman:
> What about using JWT based authentication? I think would be a more secure approach.

First, if you have conceptional remarks about a change, it would be awesome to give those in the first round of feedback, not in the second one. Second, can you elaborate on how exactly JWT would be more secure? The password is sent exactly once, that will be needed however auth works. The token can be any random value which must be hard to guess but making it a JWT or not will not make anything more or less secure.

ramiresviana:
> > What about using JWT based authentication? I think would be a more secure approach.
> 
> First, if you have conceptional remarks about a change, it would be awesome to give those in the first round of feedback, not in the second one. Second, can you elaborate on how exactly JWT would be more secure? The password is sent exactly once, that will be needed however auth works. The token can be any random value which must be hard to guess but making it a JWT or not will not make anything more or less secure.

Just wanted to hear your opinion about JWT, that's not an requirement. I was worried about the fact that the share token is always the same, with the JWT standard the token needs to be refreshed periodically and the frontend ~do not have access to the payload~ (it has access to payload but cannot generate a valid token).

frontend/src/components/prompts/Share.vue
ramiresviana: https://github.com/filebrowser/filebrowser/blob/f1b7bd59f67e719b7bfd203b0d7ec016fd21ab49/frontend/src/components/prompts/Share.vue#L69-L77

Some errors will appear on the console when `passwordpermalink` is not here.

frontend/src/views/Share.vue
ramiresviana: I see, lets have it set on [/store/index.js](https://github.com/filebrowser/filebrowser/blob/master/frontend/src/store/index.js). Also use `this.$store.commit()` instead of importing the store.

alvaroaleman:
> Just wanted to hear your opinion about JWT, that's not an requirement. I was worried about the fact that the share token is always the same, with the JWT standard the token needs to be refreshed periodically and the frontend do not have access to the payload (it has access to payload but cannot generate a valid token).

Well, I guess that depends on the expectation. Right now the download links from the share will work indefinitely for as long as that share exists, which is nice for example if you want to be able to pass around a link that will work without making the share public. If I wanted to brute force a filebrowser share, I would most likely try to brute force the password instead, because I would assume that it is almost always shorter than the 96 byte of the token and that it will in most cases be easier to guess.

frontend/src/components/prompts/Share.vue
alvaroaleman: @ramiresviana fixed

frontend/src/views/Share.vue
alvaroaleman: @ramiresviana fixed

frontend/src/views/Share.vue
alvaroaleman: @ramiresviana fixed

frontend/src/views/Share.vue
alvaroaleman: @ramiresviana fixed

frontend/src/views/Share.vue
alvaroaleman: @ramiresviana fixed

ramiresviana (approved):

alvaroaleman:
@o1egl Can you have a look, please?

.circleci/config.yml
```diff
@@ -2,7 +2,7 @@ version: 2
 jobs:
   lint:
     docker:
-      - image: golangci/golangci-lint:v1.27.0
+      - image: golangci/golangci-lint:v1.31.0
```
o1egl, line 5: The last version is v1.36.0. Either upgrade to the latest or revert it.

cmd/root.go
```diff
@@ -47,6 +47,7 @@ func init() {
 	flags.Bool("noauth", false, "use the noauth auther when using quick setup")
 	flags.String("username", "admin", "username for the first user when using quick config")
 	flags.String("password", "", "hashed password for the first user when using quick config (default \"admin\")")
+	flags.String("salt", "", "The salt to use when for hashing share password. Can be any value. If changed, existing password-protected shares wil stop working.") //nolint:lll
```
o1egl: Why this salt is needed? According to the https://en.wikipedia.org/wiki/Bcrypt#Description , the salt is automatically included in the output string which means there is no need to add it by yourself. 

```suggestion
	flags.String("salt", "", "The salt is used for hashing share passwords. It can be of any value. If changed, existing password-protected shares will stop working") //nolint:lll
```

frontend/src/api/files.js
```diff
@@ -77,8 +77,13 @@ export function download (format, ...files) {
   if (format !== null) {
     url += `algo=${format}&`
   }
+  if (store.state.jwt !== ''){
+    url += `auth=${store.state.jwt}`
```
o1egl: If both jwt and token are not empty, the generated URL will be incorrect
```suggestion
    url += `auth=${store.state.jwt}&`
```

frontend/src/components/prompts/Share.vue
```diff
@@ -72,7 +77,9 @@ export default {
       unit: 'hours',
       hasPermanent: false,
       links: [],
-      clip: null
+      clip: null,
+      password: '',
+      passwordPermalink: ''
```
o1egl, line 82: Why is it 2 different variables?

frontend/src/i18n/en.json
```diff
@@ -142,7 +143,8 @@
     "show": "Show",
     "size": "Size",
     "upload": "Upload",
-    "uploadMessage": "Select an option to upload."
+    "uploadMessage": "Select an option to upload.",
+    "optionalPassword": "Optional Password"
```
o1egl:
```suggestion
    "optionalPassword": "Optional password"
```

frontend/src/views/Share.vue
```diff
@@ -74,6 +74,24 @@
   <div v-else-if="error">
     <not-found v-if="error.message === '404'"></not-found>
     <forbidden v-else-if="error.message === '403'"></forbidden>
+    <div v-else-if="error.message === '401'">
+      <div class="card floating" id="password">
+        <div v-if="attemptedPasswordLogin" class="share__wrong__password {">{{ $t('login.wrongCredentials') }}</div>
```
o1egl:
```suggestion
        <div v-if="attemptedPasswordLogin" class="share__wrong__password">{{ $t('login.wrongCredentials') }}</div>
```

frontend/src/views/Share.vue
```diff
@@ -203,6 +232,7 @@ export default {
         this.updateRequest(file)
         this.setLoading(false)
       } catch (e) {
+        console.log(e)
```
o1egl: Is it kept on purpose?

http/public.go
```diff
@@ -94,3 +104,26 @@ var publicDlHandler = withHashFile(func(w http.ResponseWriter, r *http.Request,
 
 	return rawDirHandler(w, r, d, file)
 })
+
+func authenticateShareRequest(r *http.Request, l *share.Link, salt string) (int, error) {
```
o1egl: This is not the convenient way of writing go functions. Returning values on error is a bad practice

http/public.go
```diff
@@ -94,3 +104,26 @@ var publicDlHandler = withHashFile(func(w http.ResponseWriter, r *http.Request,
...
+		return 0, nil
+	}
+
+	if r.URL.Query().Get("token") == l.Token {
+		return 0, nil
+	}
+
+	password := r.Header.Get("X-SHARE-PASSWORD")
+	if password == "" {
+		return http.StatusUnauthorized, nil
```
o1egl, line 119:
```suggestion
		return errors.New("password is required")
```

http/public.go
```diff
@@ -19,6 +22,11 @@ var withHashFile = func(fn handleFunc) handleFunc {
 			return errToStatus(err), err
 		}
 
+		status, err := authenticateShareRequest(r, link, d.server.Salt)
+		if status != 0 || err != nil {
+			return status, err
+		}
```
o1egl:
```suggestion
		if err := authenticateShareRequest(r, link, d.server.Salt); err != nil {
			return http.StatusUnauthorized, err
		}
```

http/public_test.go
```diff
@@ -0,0 +1,143 @@
...
+
+func TestPublicShareHandlerAuthentication(t *testing.T) {
+	t.Parallel()
+
+	const passwordBcrypt = "$2y$10$TFAmdCbyd/mEZDe5fUeZJu.MaJQXRTwdqb/IQV.eTn6dWrF58gCSe" //nolint:gosec
+	testCases := []struct {
+		name               string
+		share              *share.Link
+		req                *http.Request
+		expectedStatusCode int
```
o1egl: Please use `map[string]struct{}` for table tests definition. See https://github.com/filebrowser/filebrowser/blob/master/img/service_test.go#L20

http/public_test.go
```diff
@@ -0,0 +1,143 @@
...
+				defer result.Body.Close()
+				if result.StatusCode != tc.expectedStatusCode {
+					t.Errorf("expected status code %d, got status code %d", tc.expectedStatusCode, result.StatusCode)
+				}
+			})
+		}
+	}
+}
+
+func newHTTPRequest(t *testing.T, m ...func(*http.Request)) *http.Request {
```
o1egl:
```suggestion
func newHTTPRequest(t *testing.T, requestModifiers ...func(*http.Request)) *http.Request {
```

http/share.go
```diff
@@ -82,6 +88,10 @@ var sharePostHandler = withPermShare(func(w http.ResponseWriter, r *http.Request
 	rawExpire := r.URL.Query().Get("expires")
 	unit := r.URL.Query().Get("unit")
```
o1egl: Let's use JSON body to send all required params

http/share.go
```diff
@@ -137,3 +164,34 @@ var sharePostHandler = withPermShare(func(w http.ResponseWriter, r *http.Request
 
 	return renderJSON(w, r, s)
 })
+
+func getSharePasswordHash(bodyReader io.Reader, salt string) (data []byte, statuscode int, err error) {
+	if bodyReader == nil {
+		return nil, 0, nil
+	}
+
+	bodyRaw, err := ioutil.ReadAll(bodyReader)
```
o1egl: Please use `json.NewDecoder()` instead

share/storage.go
```diff
@@ -102,7 +102,9 @@ func (s *Storage) Gets(path string, id uint) ([]*Link, error) {
 			if err := s.Delete(link.Hash); err != nil {
 				return nil, err
 			}
-			links = append(links[:i], links[i+1:]...)
+			if len(links) > i+1 {
+				links = append(links[:i], links[i+1:]...)
+			}
```
o1egl: Why this change is introduced?

users/storage.go
```diff
@@ -17,6 +17,15 @@ type StorageBackend interface {
 	DeleteByUsername(string) error
 }
 
+type Interface interface {
```
o1egl: Please give a proper name to this interface. `UserStore`?

o1egl (changes requested):

.circleci/config.yml
alvaroaleman: It might not look like that, but the version was a very conscious choice, it had to be updated to something with go 1.15 to have support for `*testing.T.TempDir` which was added in go 1.15. This is the oldest version of the golangci-lint that uses go 1.15.

Updating it to v1.36.0 results in a bunch of new failures, because filebrowser doesn't have a golangci-lint config, so it uses whatever its default linters are, which is not stable across versions.

frontend/src/components/prompts/Share.vue
alvaroaleman: Because its two different inputs

http/public.go
alvaroaleman: An error here means something went wrong with our code and yes, in that case the status code can be ignored.  Statuscode != 0 means the verification failed. The distinction is important, because they result in different http codes. Our code erroring is a 5XX, the verification failing is a 4XX, if we return an error in both cases, we can not distinguish those two anymore

share/storage.go
alvaroaleman: Because a bunch of methods in storage.go unconditionally access `links[i+1:]` which results in an out of bounds panic if the last link expired. I will remove it from this PR.

users/storage.go
alvaroaleman: The linter will refuse that due to stuttering, package symbols should not start with the package name. This ends up being used as `users.Interface` which IMHO is totally reasonable

alvaroaleman (commented):
@o1egl all feedback addressed, ptal

cmd/root.go
alvaroaleman: Removed it

share/storage.go
o1egl: This check doesn't save from data race

users/storage.go
o1egl: I just gave an example. It can be named Store. An interface name should be meaningful and self explainable. `Interface` name is too generic.

share/storage.go
alvaroaleman: Sorry, what data race do you mean? It stops accessing an array in an unsafe manner and risking an out of bounds

users/storage.go
alvaroaleman: Renamed it to `Store`, please take another look

alvaroaleman:
@o1egl any chance you can have another look?

o1egl:
@alvaroaleman `Get Permanent Link` returns 400 error. It tries to decode an empty body

http/public.go
o1egl: We already have a `errToStatus(err)` function. Let's add a mapping there

frontend/src/api/share.js
```diff
@@ -23,14 +25,18 @@ export async function remove(hash) {
...
 
-export async function create(url, expires = '', unit = 'hours') {
+export async function create(url, password = '', expires = '', unit = 'hours') {
   url = removePrefix(url)
   url = `/api/share${url}`
   if (expires !== '') {
     url += `?expires=${expires}&unit=${unit}`
   }
-
+  let body = '';
```
o1egl: The server expects to receive a JSON. An empty string is not a correct JSON.

o1egl (changes requested):

frontend/src/api/share.js
alvaroaleman: @o1egl fixed

o1egl (approved):

Issue #749, opened 2018-07-23
Workaround for videos that can't stream in browser


francislavoie:
Some videos (mostly .mkv) don't stream correctly in browsers because of missing audio codecs. Often in Chrome, the video will play, but no audio.

Normally, I'd just download the video, but that isn't the best option if the video file is huge and you just want to watch it _right now_ 😄 

My thought would be to have an option in the UI somewhere to make a URL for streaming in VLC (maybe in the File information modal?). VLC supports network streaming, so it should be a fix for the codec issue.

First problem is that if auth is enabled, VLC gets 403s when trying to load the video, obviously.

To get around auth, maybe a session token for the video can be generated and appended to the URL that would skip auth for that request?

eine:
@francislavoie, it seems that xfrr/goffmpeg is just a wrapper, so ffmpeg must be installed in the host. Since docker images are based on `scratch`, we would need either a statically compiled version of ffmpeg or change the base image. Zipping is a different issue, because it is done without any external dependency: https://github.com/mholt/archiver.

I think that using a proper media player, such as VLC, is preferred to adding any heavy dependency (because of either the size or the workload). So, it'd be interesting to have that supported with authentication.

zoidy:
I would also like this functionality. I don't know how to modify the source code so I made a simple userscript to stream the file using VLC. 

The script adds a button to the UI and downloads an M3U playlist which, when opened in VLC, will stream the file. It works with and without authentication enabled in Filebrowser. I've been using it for a while and decided to clean it up a bit and release it.  https://github.com/zoidy/filebrowser-streamlink/

francislavoie:
That's awesome! Thanks for setting that up! @hacdias since that's pure JS, any chance this could be added quickly?

vitaliy-kuzmich:
@francislavoie @zoidy  hey, if your pc runs linux, you can use this : https://github.com/browsefile/backend, and download media files like m3u playlist.

hacdias:
I'm closing this since this repository will be archived. Please read the information on the README for more information. Thanks for all your contributions!

shalva97:
> I'm closing this since this repository will be archived.

Is it still going to be archived? there is no such information in README

Anyway, I was interested if there are any updates to the issue?

francislavoie:
The repo was archived then later revived.

shalva97:
> The repo was archived then later revived.

[...]

Issue #745, opened 2019-01-15
Direct download link


freedomlang:
Is there a wat to get the direct download link of selected file?

kmickeletto:
Does not seem to work for me.

Creating a share link gives me this...
https://filebrowser.mickeletto.com/share/V8kmDiIz

However, changing it to
https://filebrowser.mickeletto.com/dl/V8kmDiIz
simply gives an error stating This location can't be reached.

vitaliy-kuzmich:
@freedomlang if you want to stream video, or something, I have possibility to download m3u as a playlist, or make external share, then you can send that link with external share, and it will be viewed like usual page, and authorized as guest, so only user with link would see that files. https://github.com/browsefile/backend

hacdias:
I'm closing this since this repository will be archived. Please read the information on the README for more information. Thanks for all your contributions!

PR #1285, opened 2021-02-11
Create m3u playlist by selecting files.

g45t345rt:
**Description**
Select video or audio files and click the playlist icon in the header bar.
The program will create a playlist.m3u file and save to the desktop without having to request the server.
You can start watching videos or listening to audio by opening the playlist.m3u with VLC.

**Why**
Avoid copy and paste the raw link address every time you want to open the file in VLC.
Avoid downloading the entire file before opening in VLC.

_Related with_
https://github.com/filebrowser/filebrowser/issues/745
https://github.com/filebrowser/filebrowser/issues/749
https://github.com/zoidy/filebrowser-streamlink/


o1egl:
To access the file, the auth token must be provided. So creating a file with hardcoded links is not the right solution. Also, before merging this type of functionality, we need to be sure that it is in high demand.

zoidy:
This functionality is very useful for me. I ended up implementing it via the backend for better robustness. By doing it this way, it's possible to create playlists of entire directory trees and avoids any auth issues by creating shared links in the backend directly.  You can see the changes in my fork https://github.com/zoidy/filebrowser  ~~The fork is currently out of date from upstream but I will update it~~ fork updated

methatronc:
> 
> 
> To access the file, the auth token must be provided. So creating a file with hardcoded links is not the right solution. Also, before merging this type of functionality, we need to be sure that it is in high demand.

If that could be of any help, I second that request : I focus on optimizing resources on a light server and being able to share terabits of data without encoding them with the only client-requirement of installing vlc would be quite awesome !

snvoid:
This would a great to have so hopefully it gets implemented.

xiaole51:
This is a very needed function, batch exporting and sharing links should be realized, at least with CLI.

PR #1307, opened 2021-03-04
Frontend code quality changes

ramiresviana:
### feat: improved sharing prompt 
Sharing prompt now is more user friendly, having bigger buttons and text, with different views for creation and listing. For simplicity the single permanent share restriction was removed, allowing to have multiple permanent shares.

### feat: dual pane settings view 
Settings cards are now shown side by side when there's enough screen space.

### feat: improved settings navbar 
Added a full width divider and effects for hovering and active state. The navbar will show a scroll bar when there's not enough screen space.

### feat: larger previewer content 
The margins from the previewer content were removed, allowing it to fill the screen. Navigation buttons will now be hidden based on cursor activity.

### chore: header bar component 
There are 3 implementations for the header (Files, Editor and Preview). Having it as a reusable component reduces the amount of code and helps with manutenability and consistency. 

### chore: files pages logic responsability 
The Files view handles logic for other views and has 2 variations of the header for different views (Listing and Share), making hard to add new buttons and custom logic. Having a header for each view allows the view itself to control the behavior of the header. The Editor, Listing, and Preview view were moved to the views folder, following the Settings view structure.

### feat: increased header button counter size 
Make the counter more visible by increasing the spacing and the font weight.

### chore: removed header buttons components 
Most of the header buttons logic are limited for the Files view, they will just open a prompt with simple logic. The new Action component can cover most of the current cases. Further logic for buttons are now written on its parent view.

### chore: files view dynamic component 
Different components are being displayed using conditional render on Files view, which affects readability and makes harder to add new components. Using dynamic components allows having a single function to control the rendered component.

### chore: breadcrumbs component 
Breadcrumbs have its own implementation on 3 different views (Listing, Editor and Share). Having it as a reusable component reduces the amount of code and helps with manutenability and consistency.

### chore: share view logic responsability 
The Share view adds unnecessary data to the store and the api files have logic related to the Share view. The public api should have its own file, having logic related to the Share view. By having the logic on the view itself, unnecessary items can be removed from store.

### chore: remove prompts events 
Delete prompt is using an edge case implementation of events, it can make use of the showConfirm callback.


o1egl:
Well done @ramiresviana .
Nit: I get a warning in the dev console when the `Shares.vue` is rendered.
Screenshot 2021-03-08 at 19 08 26

ramiresviana:
> Well done @ramiresviana .
> Nit: I get a warning in the dev console when the `Shares.vue` is rendered.
> Screenshot 2021-03-08 at 19 08 26

Thanks, just a copy-pasting typo that went unnoticed. Should be good now.

PR #1330, opened 2021-03-15
handles json files like text files to make them editable again

mheers:
**Description**
`.json` files were not editable anymore because the backend used `blob` as mimetype for these files.


ramiresviana:
This issue was introduced in #1175, a proper solution would be to make the file detection behave as it was before.

o1egl:
Fixed in https://github.com/filebrowser/filebrowser/pull/1353

PR #1353, opened 2021-03-26
Some fixes

### LINKED ISSUE #1238: No additional loading triggered when window can accommodate initial files without scrolling.
**Description**
On a large display (1440p) in a browser with default font and zoom settings, in a folder with a lot of files, a large browser window can accommodate all of the 50 files that are loaded initially without any need for scrolling. Because loading of additional files seems to be triggered by the scroll bar hitting the bottom, File Browser will not load additional files unless the window is made smaller so that scrolling occurs.

**Expected behaviour**
I expected File Browser to go beyond the initial 50 files and continue to load files until the screen is filled (that is, until scrolling is necessary to access the overflow).

File Browser only loads 50 files and won't load more unless the window is reduced in size. No log messages are emitted.

**Additional context**
See video of behavior here: https://www.youtube.com/watch?v=-FmH4acRquo

**How to reproduce?**

1. Create a folder with a bunch of files; e.g. `for LETTER1 in {A..Z} ; do for LETTER2 in {A..Z} ; do touch ${LETTER1}${LETTER2} ; done ; done`
2. Launch File Browser with the root dir set to the folder containing the many files.
3. In a large browser window (e.g. maximized at 1440p), access the File Browser web GUI and note that files beyond the initial 50 are not loaded.

**Files**
N/A.

### LINKED ISSUE #1333: Play next song in Safari IPHONE
**Description**
When playing audio files from any folder file browser used to start playing the next song...

**Expected behaviour**
On the last Docker version audio playback skipped to the next song automatically. 

**What is happening instead?
You now have to press play on each skip. 

**Additional context**
I know this is not the core feature of fb but no iOS update since this is happening only updated to latest Docker armv7 release. 

**How to reproduce?**
Start playing a file in safari it will not automatically start for example. 

Thanks team for the arm support and great file browser. It reminds me very much of my WebDAV mod I created many years ago. But based on Java script and css. 

You should add basic zip creation on shares links going forward, this doesn’t need to be cpu intensive. Just normal.

### LINKED ISSUE #1348: Json files an others are not recognized as text files
**Description**
After the release of v2.14.0 we can open and view the json files, but we can't still create a json file directly from filebrowser and edit it.

**Expected behaviour**

1. Open filebrowser and click on `new file`
2. Name it `my-file.json`
3. The file should be opened for edition 

1. The system show the download button instead of edition view

**Additional context**
Version: v2.14.0

**How to reproduce?**
Try to create a json file directly from filebrowser and modify it

**Files**
Nothing

### LINKED ISSUE #1349: [Frontend] Top header not visible in Safari
**Description**
The top banner is not visible when a text file, for e.g., is opened, but I can click the close button to close it.

**Expected behaviour**
Visible top banner. 

image

image

**Additional context**
N/A

**How to reproduce?**
Open any text file. 

**Files**
N/A

ramiresviana:
### fix: error causes panic on upload
A non handled error was causing panic when the upload was not successful, causing the incomplete file to not be deleted after.

### fix: empty text file on editor
The file type detection was preventing to edit empty files. fixes #1348 

### fix: hidden editor header on Safari
The Safari browser has an render issue when using the margin css attribute. fixes #1349 

### feat: dynamic item count on file listing
Calculates the amount of items needed to fill the screen, displaying 2 pages initially. The display of more items is triggered when the scroll reachs 75% of total page height, increasing the number of items by 2 pages. fixes #1238 

### fix: no header button animations on file listing
Progress and success state are not being displayed on header buttons.

### fix: missing bold variation for Roboto font
The bold variation is being used, but is not available, causing the text to be blurred and inconsitent.

### fix: image quality switch on previewer
The quality switch was not working since the addition of tiff support.

### fix: list item interactions on share
Text selection was being triggered when trying to select files on the listing. Also added support for closing prompts using esc key on share view.

### feat: dynamic zoom limit on previewer
The static zoom limit on the image preview was making not possible to scale some images to its full size. The zoom limit is now calculated based on the screen size.

### fix: mouse wheel zoom on previewer
Mouse wheel was not working on previewer on some browsers due to a deprecated implementation.

### fix: buttons without permission on header
Some header buttons are being displayed even if the user was not permission.

### fix: check modify permission on file overwrite
A existing file could be manipulated without the modify permission by uploading a new one. The modify permission will now be checked when overwriting a file.

### fix: empty archive name on directory download
A archive with empty name was being generated when downloading from the root folder from a share or file listing.

### feat: dynamic autoplay on previewer
The previewer will autoplay when the previous media file was playing. fixes #1333


http/preview.go
```diff
@@ -35,9 +35,6 @@ type FileCache interface {
 
 func previewHandler(imgSvc ImgService, fileCache FileCache, enableThumbnails, resizePreview bool) handleFunc {
 	return withUser(func(w http.ResponseWriter, r *http.Request, d *data) (int, error) {
-		if !d.user.Perm.Download {
-			return http.StatusAccepted, nil
-		}
```
o1egl: This means that the user can download any image from the server without corresponding permission.

ramiresviana: Currently the image files are listed, but shows broken img elements on the listing and previewer. With this change, users that dont have the download permission will be allowed to view resized images (thumbnails and previews), but still wont be able to download the raw image. This follows the concept of text files, that users without download permission can view the content. The main idea would be allowing users to view any file handled by the previewer.

Do you think that change should be removed?

o1egl: If resizing is disabled, the previewer will return the original image. The same happens when the HD button is pressed. Giving users access to the content but not letting them download it doesn't make sense. Nothing stops a user from pressing the right button and save the image. In my opinion, previewing text files without download permission is a bug.

ramiresviana: I see, changing the behavior of unprivileged users can also be unexpected to servers admins, since it currently have no customization. I will left it as is since would require a overhaul to make it fit for all use causes.

PR #1356, opened 2021-03-29
Add HEIF support

adrium:
**Description**

Add HEIF/HEIC support, a commonly used format with iPhones.

- [x] File Browser can be successfully built **(amd64 and arm64 tested)**
- [x] continuous integration build passed

**Further comments**

This is an experimental change and may lead to more maintenance overhead than desired.
Anyway, I wanted to try it out and gather some feedback.

It builds successfully on `linux/amd64` and `linux/arm64`
It also builds on `darwin/amd64` with warnings.

It is able to read images 1-3 from https://github.com/tigranbs/test-heic-images but it segfaults for image 4.

Now the following questions arise:

1. Is it a generally embraced addition?
2. Is it possible to catch the SIGSEGV?
3. Should target architectures be conditionally included?


jniggemann:
Would be a great feature to have!

PLK2:
Here's my vote for HEIC support.

pushpinderbal:
Would be a great feature to have in filebrowser.

gaby:
👍🏻 for this feature!

TCB13:
+1 on this feature, please merge.

o1egl:
This feature adds a dependency on CGO. With that our existing release pipeline will be broken. It cannot be merged without rewriting the build pipeline to support CGO cross-compilation.

PR #1373, opened 2021-04-16
Some fixes

### LINKED ISSUE #1369: Support for playing shared videos inline
I wish that I could share video without the user needing to download the actual file. I don't want people needing to download a 500mb file whenever they could just stream it.

The ability to inline embed the file into the share link, or share the file directly (www.domain.com/share/whatever/file.mp4) and have it play in the browser.

I can copy the share download link and hope it embeds into discord or whatever else I am using, but it would nice to see it being able to play in the browser with no need for a download.

If this is already possible with a little bit of code modification, I am happy to change a bit of code on my end.

### LINKED ISSUE #1372: Shared folders are limited to 500 items.
**Description**
I shared a folder with 572 pdf files.

**Expected behaviour**
My friend should be able to access to all files by clicking on "+72".

After the 500th item, there is a "+72" button that doesn't work.

**Additional context**
image

**How to reproduce?**
Write 500+ files in a folder and share the folder.

ramiresviana:
### fix: no items displayed on file listing
Opening a empty directory was causing the number of displayed items to be invalid. The number is now calculated for every directory.

### fix: delete image cache when moving
The cached file was being kept after changing the original file location.

### fix: inconsistent double click on listing item
The dblclick has a different behavior between browsers. Using timeout ensures the double click consistency.

### fix: copying files with special characters
Manipulatingd directories and file with names containing special characters was causing errors.

### feat: support for IE11 browser
Making the frontend work on IE was very simple, requiring just some polyfills, without affecting other browsers and project maintenability.

### chore: split action on resource patch handler
The complexity of the function was triggering some linting errors.

### feat: loading spinner on views navigation
The spinner will is displayed on file listing, previewer and share. The delay prevents it from blinking on good network conditions, making it visible only after 100ms.

### feat: message for connection error
An error message is displayed when there's an issue with the request.

### feat: display error messages on settings
Error messages are now displayed on settings pages.

fixes #1369 fixes #1372

### DIFF
--- a/frontend/src/utils/auth.js
+++ b/frontend/src/utils/auth.js
@@ -12,6 +12,8 @@ export function parseToken(token) {
 
   const data = JSON.parse(Base64.decode(parts[1]));
 
+  document.cookie = `auth=${token}; path=/`;
+
   localStorage.setItem("jwt", token);
   store.commit("setJWT", token);
   store.commit("setUser", data.user);
@@ -81,6 +83,8 @@ export async function signup(username, password) {
 }
 
 export function logout() {
+  document.cookie = "auth=; expires=Thu, 01 Jan 1970 00:00:01 GMT; path=/";
+
   store.commit("setJWT", "");
   store.commit("setUser", null);
   localStorage.setItem("jwt", null);
--- a/http/data.go
+++ b/http/data.go
@@ -49,6 +49,8 @@ func (d *data) Check(path string) bool {
 
 func handle(fn handleFunc, prefix string, store *storage.Storage, server *settings.Server) http.Handler {
 	handler := http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		w.Header().Set("Cache-Control", "no-cache")
+
 		settings, err := store.Settings.Get()
 		if err != nil {
 			log.Fatalf("ERROR: couldn't get settings: %v\n", err)
--- a/http/data.go
+++ b/http/data.go
@@ -49,7 +49,7 @@ func (d *data) Check(path string) bool {
 
 func handle(fn handleFunc, prefix string, store *storage.Storage, server *settings.Server) http.Handler {
 	handler := http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-		w.Header().Set("Cache-Control", "no-cache")
+		w.Header().Set("Cache-Control", "no-cache, no-store, must-revalidate")
 
 		settings, err := store.Settings.Get()
 		if err != nil {
--- a/http/static.go
+++ b/http/static.go
@@ -2,6 +2,7 @@ package http
 
 import (
 	"encoding/json"
+	"fmt"
 	"io/fs"
 	"log"
 	"net/http"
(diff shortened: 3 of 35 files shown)


ramiresviana:
### feat: browser cache directives
Added headers to control the cache behavior on browser.

### chore: split preview creation logic
The complexity of the function was triggering some linting errors.

### chore: set public path on the fly
Seting the public path on the frontend we dont need to replace template strings of javascript files.

### feat: gzip encoding for static js files
Reduces the size of transferred js files.

### feat: open file option on share
Adds a button to open a share file on the browser. fixes #1369 

### feat: mod time title on file info
Shows the date of the file by hovering the date information.

### feat: file name on page title
Adds the name of the current file to the page title on files and share views.

### feat: show more button on share
Allows to show more items on the share listing. fixes #1372

### fix: omit file content
Omit the file content from the response, reducing the download on cases that the content is not required.

williewillus:
Just chiming in to say that I saw that this fixes #1369 and to thank you for implementing that! Would be cool if *all* shared files that can be previewed, are previewed on the share page (I'm not sure if this PR does it for just videos or for all previewable things)

Look forward to using this in the next release.

http/data.go
```diff
@@ -49,6 +49,8 @@ func (d *data) Check(path string) bool {
 
 func handle(fn handleFunc, prefix string, store *storage.Storage, server *settings.Server) http.Handler {
 	handler := http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		w.Header().Set("Cache-Control", "no-cache, no-store, must-revalidate")
```
o1egl, line 52: It doesn't make sense to set no-cache and no-store together. This is the correct headers to wipe existing and prevent to cache upcoming responses.
```suggestion
		w.Header().Set("Cache-Control", "no-store, max-age=0")
```

http/static.go
```diff
@@ -127,7 +131,19 @@ func getStaticHandlers(store *storage.Storage, server *settings.Server, assetsFs
 			return 0, nil
 		}
 
-		return handleWithStaticData(w, r, d, assetsFs, r.URL.Path, "application/javascript; charset=utf-8")
+		fileContents, err := fs.ReadFile(assetsFs, r.URL.Path+".gz")
+		if err != nil {
+			return http.StatusNotFound, err
+		}
+
+		w.Header().Set("Content-Encoding", "gzip")
```
o1egl, line 139: What if the browser doesn't support gzipped response?

frontend/src/utils/auth.js
```diff
@@ -12,6 +12,8 @@ export function parseToken(token) {
 
   const data = JSON.parse(Base64.decode(parts[1]));
 
+  document.cookie = `auth=${token}; path=/`;
```
o1egl, line 15: I don't think that it is a good idea to create an auth cookie accessible by clients. I would suggest generating an anonymous JWT token for unauthorized users. As an option, we can create a special file access token instead of using auth token for that.

http/static.go
ramiresviana: The content encoding and gzip was introduced long time ago in HTTP/1.1 (https://tools.ietf.org/html/rfc2616#section-3.5) and is probably supported by all browsers. (https://caniuse.com/sr_content-encoding-gzip).

We can use one of the following aproaches to make it optional:
- Have two variations of static files, increasing the shipped build size.
- Decode the gzip files on the fly, affecting performance.
- Encode the static files on the fly, affecting performance.

frontend/src/utils/auth.js
ramiresviana: The token is already acessible by the client by being stored on localStorage. What makes the cookie approach inferior to the one that we already have? The objetive of the cookie is authenticate browser requests without the auth query param, which is required to make the cache work.

http/static.go
ramiresviana: I think that decoding gzip on the fly is a good fallback and wont affect performance since majority of the requests will be served as gzip. Decoding will only happen when the request does not support gzip.

PR #1409, opened 2021-05-21
Use config port for docker healthcheck

mRoca:
This PR allows to use the configured port in docker healthcheck.

There is a big issue with the default docker config, for now: if we want to use another user than `root` (`--user 1000:1000`), we obviously cannot use any reserved port (`0-1023`), so we must specify a port bigger than 1023.

The default healthcheck always targets the 80 port. The container will be unhealthy if we choose a custom port.

**The best way to fix this issue would be to always use a port > 1023 (as `8080` or `3000`), and to remove the `port` option, but this will be a bc break for all people using this image**

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,10 +1,11 @@
 FROM alpine:latest
 RUN apk --update add ca-certificates \
                      mailcap \
-                     curl
+                     curl \
+                     jq
 
 HEALTHCHECK --start-period=2s --interval=5s --timeout=3s \
-  CMD curl -f http://localhost/health || exit 1
+  CMD curl -f http://localhost:$(jq '.port' /.filebrowser.json)/health || curl -f http://localhost/health || exit 1
 
 VOLUME /srv
 EXPOSE 80


Dockerfile
```diff
@@ -1,10 +1,11 @@
 FROM alpine:latest
 RUN apk --update add ca-certificates \
                      mailcap \
-                     curl
+                     curl \
+                     jq
 
 HEALTHCHECK --start-period=2s --interval=5s --timeout=3s \
-  CMD curl -f http://localhost/health || exit 1
+  CMD curl -f http://localhost:$(jq '.port' /.filebrowser.json || 80)/health || curl -f http://localhost/health || exit 1
```
o1egl: 1) The default port is 8080
2) `|| curl -f http://localhost/health` is redundant

mRoca: 1. The default port is 80 with your docker config: https://github.com/filebrowser/filebrowser/blob/master/.docker.json
2. If the `.filebrowser.json` does not contain `"port" : 1234`, the `jq` command will fail, and no port will be returned. That's why we need a default function call.

o1egl: You already do it in this part `curl -f http://localhost:$(jq '.port' /.filebrowser.json || 80)/health`

mRoca: Indeed, the `|| 80`  was useless, as the command returns a `null` string if the `.port` value is not found. I've removed it, but we have to keep the fallback method in order to deal with json files without `port` value.

imho, you should definitely choose a port > 1023, and remove the `port` option from the json file. There is no reason to allow to change the port inside the container, as we can bind any host port to a container's one, and we can choose any container port in a proxy config (Traefik, ...).

The only problem I can see is the BC break for all people using the existing image with the port `80`

Dockerfile
```diff
@@ -1,10 +1,11 @@
 FROM alpine:latest
 RUN apk --update add ca-certificates \
                      mailcap \
-                     curl
+                     curl \
+                     jq
 
 HEALTHCHECK --start-period=2s --interval=5s --timeout=3s \
-  CMD curl -f http://localhost/health || exit 1
+  CMD curl -f http://localhost:$(jq '.port' /.filebrowser.json)/health || curl -f http://localhost/health || exit 1
```
o1egl, line 8:
```suggestion
  CMD curl -f http://localhost:$(jq '.port' /.filebrowser.json || 80)/health || exit 1
```

mRoca: As I explained it in my last comment, it's  impossible to do it like this: if the json node `port` does not exist, the `jq` command will still return something (the `null` string) with a `0` exit code. The `|| 80` won't work.

The only thing we could do here would be `curl -f http://localhost:$(jq '.port' /.filebrowser.json || echo '80')/health`, but this will only catch a `jq` error, as a file not found.

jhollowe:
just a note: the use of `jq` completely assumes (and limits) the format of the config file to JSON. It currently works for multiple config file types (YAML, JSON, TOML) and I think forcing a JSON config just for this one feature seems a shame.

I would just set the port to a common non-root port (i.e. 8080), set the default config's port to it, set the EXPOSE to it, and set the HEALTHCHECK to use it. If someone really wants to change the port inside the container, they can deal with changing it in the healthcheck.

hepto:
I know this doesn't solve the main issue that the HEALTHCHECK should reflect the desired/configured port rather than be statically set, but as a workaround you can override the image CMD via `run`:

https://docs.docker.com/engine/reference/run/#healthcheck

For example:

`--health-cmd "curl -f http://localhost:8080/health || exit 1"`

PR #1410, opened 2021-05-21
Use database dir for docker image

mRoca:
For now, it's impossible to start the docker image with another user than `root` without a previously created database file mouted as a volume.

With this PR, it'll be possible to use a custom user and docker volumes (directories), or no volume at all.

The doc must be updated with the followed command:
```
mkdir /path/filebrowserdb && chmod a+w /path/filebrowserdb

docker run \
    -v /path/to/root:/srv \
    -v /path/filebrowserdb:/var/lib/database.db \
    -v /path/.filebrowser.json:/.filebrowser.json \
    --user $(id -u):$(id -g)
    -p 80:80 \
    filebrowser/filebrowser
```

### DIFF
--- a/.docker.json
+++ b/.docker.json
@@ -3,6 +3,6 @@
   "baseURL": "",
   "address": "",
   "log": "stdout",
-  "database": "/database.db",
+  "database": "/var/lib/filebrowser/database.db",
   "root": "/srv"
 }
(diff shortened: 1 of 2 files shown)


jhollowe:
I think it would be better to change the default to something like `/config/database.db`.

If `/config/` was also added to the default search locations for the config file, you could just mount the `/config/` directory containing the config file and the DB file would be created in the folder.

daiaji:
It seems that neither the database nor the configuration file location comply with the XDG Base Directory Specification.

PR #1495, opened 2021-08-03
Fix 401 error in share view open file button

andrew-kennedy:
This link would be generated with a `?token=<hash>?inline=true` when it should be `?token=<hash>&inline=true`. This was causing 401 Unauthorized errors because `URL.Query().Get("token")` would return the empty string and so [authenticateShareRequest](https://github.com/filebrowser/filebrowser/blob/dee465ab865ba5bf96fbfbe5888f4e95731d3701/http/public.go#L126) would try to verify the password in the header, which doesn't exist for these requests.

### DIFF
--- a/frontend/src/views/Share.vue
+++ b/frontend/src/views/Share.vue
@@ -93,7 +93,7 @@
             </a>
             <a
               target="_blank"
-              :href="link + '&inline=true'"
+              :href="inlineLink"
               class="button button--flat"
               v-if="!req.isDir"
             >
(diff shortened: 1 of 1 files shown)


frontend/src/views/Share.vue
```diff
@@ -93,7 +93,7 @@
             </a>
             <a
               target="_blank"
-              :href="link + '?inline=true'"
+              :href="link + '&inline=true'"
```
ramiresviana: This will make the link invalid on shares without password.

andrew-kennedy: Good point, I really need to properly construct the query string using official builder methods.

ramiresviana:
Please make sure that all checks are passing.

ramiresviana:
You will need to squash your commits, taking into account the commintlint rules.

PR #1499, opened 2021-08-06
Some fixes

### LINKED ISSUE #1497: Video preview
播放不了1080p的视频 一直加载
6BPY{6(UT%})XEYOXRIW5O7

ramiresviana:
### fix: file caching directive
Using etag header with `http.ServeContent` causes unexpected behavior. The etag was removed in favor of the `http.ServeContent` implementation. fixes #1497

### feat: open file option on preview
Allows users to open the file when the preview is not available.

### DIFF
--- a/http/preview.go
+++ b/http/preview.go
@@ -80,33 +80,26 @@ func handleImagePreview(w http.ResponseWriter, r *http.Request, imgSvc ImgServic
 		return errToStatus(err), err
 	}
 
-	isFresh := checkEtag(w, r, file.ModTime.Unix(), file.Size)
-	if isFresh {
-		return http.StatusNotModified, nil
-	}
-
 	cacheKey := previewCacheKey(file.Path, file.ModTime.Unix(), previewSize)
-	cachedFile, ok, err := fileCache.Load(r.Context(), cacheKey)
+	resizedImage, ok, err := fileCache.Load(r.Context(), cacheKey)
 	if err != nil {
 		return errToStatus(err), err
 	}
-	if ok {
-		_, _ = w.Write(cachedFile)
-		return 0, nil
+	if !ok {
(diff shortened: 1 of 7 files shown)


http/static.go
```diff
@@ -77,7 +77,7 @@ func handleWithStaticData(w http.ResponseWriter, _ *http.Request, d *data, fSys
 		return http.StatusInternalServerError, err
 	}
 
-	data["Json"] = string(b)
+	data["Json"] = strings.ReplaceAll(string(b), `'`, `\'`)
```
o1egl, line 80: Encoded json is already escaped. Why this is needed?

http/preview.go
```diff
@@ -80,33 +80,30 @@ func handleImagePreview(w http.ResponseWriter, r *http.Request, imgSvc ImgServic
...
 	cacheKey := previewCacheKey(file.Path, file.ModTime.Unix(), previewSize)
 	cachedFile, ok, err := fileCache.Load(r.Context(), cacheKey)
 	if err != nil {
 		return errToStatus(err), err
 	}
 	if ok {
-		_, _ = w.Write(cachedFile)
-		return 0, nil
-	}
+		resizedImage = cachedFile
```
o1egl: It can be rewritten to:

```go 
resizedImage, ok, err := fileCache.Load(r.Context(), cacheKey)
if err != nil {
    return errToStatus(err), err
}
if !ok {
    resizedImage, err = createPreview(imgSvc, fileCache, file, previewSize, enableThumbnails, resizePreview)
    if err != nil { 
        return errToStatus(err), err
    }
}
```

http/static.go
ramiresviana: It escapes double quotes but not single quotes. Since single quotes is used as delimiter in the index template, it needs to be escaped manually.

https://github.com/filebrowser/filebrowser/blob/25c87883908babde073390a2e2320a8e5880a87c/frontend/public/index.html#L33

Issue #1525, opened 2021-08-30
Display exact date


thoernle:
Currently, when viewing a file list, only rounded dates are displayed. Is there a way to display the exact timestamps? In many cases the rough indication is not enough.

orazmyradov:
what about setting a class dedicated to tthe layout of the timestamps?

Get Outlook for Android<https://aka.ms/AAb9ysg>
________________________________
From: thoernle ***@***.***>
Sent: Monday, August 30, 2021 4:52:32 PM
To: filebrowser/filebrowser ***@***.***>
Cc: Subscribed ***@***.***>
Subject: [filebrowser/filebrowser] Display exact date (#1525)


Currently, when viewing a file list, only rounded dates are displayed. Is there a way to display the exact timestamps? In many cases the rough indication is not enough.

PR #1536, opened 2021-09-03
Mod modified time format

Lilihx:
**Description**
Add setting for exact date format.

Default: previous "from now" format
After setting: display "YYYY-MM-DD HH:mm" format

截屏2021-09-07 下午1 53 14

#1525


o1egl:
Some people like the current behavior. It must be set as a user config param

frontend/src/i18n/en.json
```diff
@@ -214,6 +215,7 @@
     "rules": "Rules",
     "rulesHelp": "Here you can define a set of allow and disallow rules for this specific user. The blocked files won't show up in the listings and they wont be accessible to the user. We support regex and paths relative to the users scope.\n",
     "scope": "Scope",
+    "setDateFormat": "Set exact date format",
```
o1egl, line 221: duplicate of line 30

Lilihx: Actually, I copy the feture "hit dot files" to implement the issue.
there are two "hit dot files" settings in en.json
I'll commit that

frontend/src/components/files/ListingItem.vue
```diff
@@ -106,6 +106,9 @@ export default {
       return filesize(this.size);
     },
     humanTime: function () {
+      if (this.user.dateFormat) {
+        return moment(this.modified).format("YYYY-MM-DD HH:mm");
```
ramiresviana:
```suggestion
        return moment(this.modified).format("L LT");
```
FileBrowser supports multiples languages, dates should be displayed with a localized format.

Lilihx:
great

Ramires Viana ***@***.***> 于2021年9月9日周四 下午8:30写道：

PR #1542, opened 2021-09-07
make theme_color of Manifest can be set

bekcpear:
The 'theme_color' of Manifest affects the color of status bar
of mobile view, however it's static before. Make it can be set
via `filebrowser config set --branding.color "<hex-color-code>"`.

Signed-off-by: bekcpear <i@bitbili.net>

### DIFF
--- a/cmd/config.go
+++ b/cmd/config.go
@@ -41,7 +41,7 @@ func addConfigFlags(flags *pflag.FlagSet) {
 	flags.String("recaptcha.secret", "", "ReCaptcha secret")
 
 	flags.String("branding.name", "", "replace 'File Browser' by this name")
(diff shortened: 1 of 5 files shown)


ramiresviana (changes requested):
Chrome desktop/mobile ignores the theme-color from dynamic manifest. You have to use template variables on [index.html#L19](https://github.com/filebrowser/filebrowser/blob/3768e3345f071c9a7ae80b7c75092000df9afc1b/frontend/public/index.html#L19) and [index.html#L29](https://github.com/filebrowser/filebrowser/blob/3768e3345f071c9a7ae80b7c75092000df9afc1b/frontend/public/index.html#L29).

ramiresviana (approved):

PR #1573, opened 2021-09-22
"back" button function fixed

### LINKED ISSUE #1562: Press phone browser's "back" button to go to upper directory.
The problem is : 
After I browsed and viewed hundreds of pics and files, I wanna click the "back" button on my phone's web browser to move to upper directory, and finally close the browser. But it always jump to the last pics or files or dir I visited. So, I need click hundreds of times to close this website. and there is no easy way to close it. Compare to other projects, they can do it easily by pressing the "back" button on the browser.

niubility000:
fixed #1562

### DIFF
--- a/frontend/src/views/files/Preview.vue
+++ b/frontend/src/views/files/Preview.vue
@@ -241,11 +241,11 @@ export default {
     },
     prev() {
       this.hoverNav = false;
-      this.$router.push({ path: this.previousLink });
+      this.$router.replace({ path: this.previousLink });
     },
     next() {
       this.hoverNav = false;
-      this.$router.push({ path: this.nextLink });
+      this.$router.replace({ path: this.nextLink });
     },
     key(event) {
       if (this.show !== null) {
@@ -341,7 +341,7 @@ export default {
       this.$store.commit("updateRequest", {});
 
       let uri = url.removeLastDir(this.$route.path) + "/";
-      this.$router.push({ path: uri });
+      this.$router.replace({ path: uri });
     },
     download() {
       api.download(null, this.$route.path);
--- a/frontend/src/views/files/Preview.vue
+++ b/frontend/src/views/files/Preview.vue
@@ -339,7 +339,7 @@ export default {
     }, 500),
     close() {
       this.$store.commit("updateRequest", {});
-      this.$router.go(-1);
+      history.back();
     },
     download() {
       api.download(null, this.$route.path);
--- a/frontend/src/views/files/Preview.vue
+++ b/frontend/src/views/files/Preview.vue
@@ -339,9 +339,7 @@ export default {
     }, 500),
     close() {
       this.$store.commit("updateRequest", {});
-
-      let uri = url.removeLastDir(this.$route.path) + "/";
-      this.$router.push({ path: uri });
+      this.$router.go(-1);
     },
     download() {
       api.download(null, this.$route.path);
--- a/frontend/src/views/files/Preview.vue
+++ b/frontend/src/views/files/Preview.vue
@@ -339,7 +339,9 @@ export default {
     }, 500),
     close() {
       this.$store.commit("updateRequest", {});
-      history.back();
+
+      let uri = url.removeLastDir(this.$route.path) + "/";
+      this.$router.push({ path: uri });
     },
     download() {
       api.download(null, this.$route.path);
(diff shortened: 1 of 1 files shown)


frontend/src/views/files/Preview.vue
```diff
@@ -341,7 +341,7 @@ export default {
       this.$store.commit("updateRequest", {});
 
       let uri = url.removeLastDir(this.$route.path) + "/";
-      this.$router.push({ path: uri });
+      this.$router.replace({ path: uri });
```
ramiresviana: Assuming that the current directory is already on the history, `replace` will insert another history entry for the same directory, making users to press the back button twice to go back from that directory. Use the `push` method instead to avoid a duplicated history entry.

niubility000: No. Look at this pic, if we are in the preview layer, the background layer(listingitem) 's link is not a folder, it's a file's link.    If we use push here, we click the "close" button to trigger this close(), , it will go to folder "b". But after that, we click "back" button of the browser, it will revisit the c.jpg, the pic will show up again.

niubility000: If we wanna make it perfect, we need to change some other files which also have "@router.push" except "listingitem.vue". 


When I got time. I will test it one by one. 
But only changing the file "preview.vue" have already avoided most of the "back" clicks.

niubility000: oh, I've tested it. it's true. 
But if we use "push", it will revisit the last pic or file if you click the "back" button. 
Commonly, we use "this.$router.go(-1)" here, it's perfect now. 
fixed!

frontend/src/views/files/Preview.vue
```diff
@@ -339,9 +339,7 @@ export default {
     }, 500),
     close() {
       this.$store.commit("updateRequest", {});
-
-      let uri = url.removeLastDir(this.$route.path) + "/";
-      this.$router.push({ path: uri });
+      this.$router.go(-1);
```
ramiresviana: The `go` method cannot be used because:

- We do not have control of the user history, the previous entry is not necessarily from FileBrowser.
- The user history is empty when accessing the page directly by its URL, since there's no previous page nothing will happen.

ramiresviana (changes requested):

frontend/src/views/files/Preview.vue
niubility000: Thank you for your review. 
Then, what should I use? Or, just use the same default func as the "back" button: history.back(); it also works here. Hope this time it's ok. when there is no history, it will close the browser.
Or if it still has problems, hope you can make a PR for it.  🥇

ramiresviana: I think that the changes addressed on the methods `prev()` and `next()`, keeping `close()` untouched, are enough for this PR. If some browser is not behaving correctly with these changes, please add information how to reproduce the issue and will take a look.

niubility000: Ok, I will change it back.

ramiresviana (approved):

ehsundar (approved):

o1egl:
I'm not sure that it works as expected. Tested on latest chrome

https://user-images.githubusercontent.com/966539/135824038-06dbcb40-2804-41a1-9aad-f833597e9586.mov

ramiresviana:
These changes affects the back/previous buttons from the previewer. Instead of adding an history entry for every item when navigating on the previewer, it will keep only a single entry of the latest opened file.

For example: navigating through 10 images using the next button will add 10 entries to the history. With this PR, you will need to press the back button only once to exit from the preview, instead of pressing it 10 times.

niubility000:
> I'm not sure that it works as expected. Tested on latest chrome
> 
>  filebrowser.mov

Yeah, o1egl, you are testing the editor.vue, not the preview.vue.  This PR only fixed the next and prev buttons in preview.vue. When you browse the pictures using prev and next buttons, and then click back button, you will see the difference. 
As you tested, the close() in the editor.vue and preview.vue also has this problem. It can be fixed by changing the following:
    close() {
      this.$store.commit("updateRequest", {});
      let uri = url.removeLastDir(this.$route.path) + "/";
      this.$router.push({ path: uri });
    },

to 
    close() {
      this.$store.commit("updateRequest", {});
      history.back();
    },
ramiresviana recommends to make another PR for this later.

PR #1576, opened 2021-09-23
fixed sidebar function

### LINKED ISSUE #1575: a small bug in "sidebar" function
The small bug is :
Using a phone browser, when you are already on the "my files" page, you open the sidebar and click the "my files" button, nothing happened, the sidebar is still there and can't close itself because the webpage doesn't refresh. then, you will feel it's a bug. 
actually, that's a bug in Vue, the official website has two solutions.

niubility000:
fixed #1575

### DIFF
--- a/frontend/src/components/Sidebar.vue
+++ b/frontend/src/components/Sidebar.vue
@@ -4,6 +4,7 @@
       <router-link
         class="action"
         to="/files/"
+        @click.native="refresh"
         :aria-label="$t('sidebar.myFiles')"
         :title="$t('sidebar.myFiles')"
       >
@@ -37,6 +38,7 @@
         <router-link
           class="action"
           to="/settings"
+          @click.native="refresh"
           :aria-label="$t('sidebar.settings')"
           :title="$t('sidebar.settings')"
         >
@@ -127,6 +129,10 @@ export default {
   methods: {
     help() {
       this.$store.commit("showHover", "help");
+    },
+
+    refresh:function(){
+      this.$router.go(0);  
     },
     logout: auth.logout,
   },


frontend/src/components/Sidebar.vue
```diff
@@ -4,6 +4,7 @@
       <router-link
         class="action"
         to="/files/"
+        @click.native="refresh"
```
ramiresviana, line 7: The `router-link` component already handles the click event, adding another click directive will cause unexpected behavior. Instead of using a component, you should have a new element and handle its click behavior.

frontend/src/components/Sidebar.vue
```diff
@@ -128,6 +130,10 @@ export default {
     help() {
       this.$store.commit("showHover", "help");
     },
+
+    refresh:function(){
+      this.$router.go(0);
```
ramiresviana, line 135: Using `go(0)` is inefficient, causing the whole page to be reloaded. You should use the `showHover` state to control the sidebar visibility.

ramiresviana (changes requested):

frontend/src/components/Sidebar.vue
niubility000: Good point. But if using showHover, the under layer (files) will still keep inactive. I need to click the screen to make it active. I don't know how to solve it.

frontend/src/components/Sidebar.vue
niubility000: Would you please fix it and pull a request?  I just don't know where to put the element.

frontend/src/components/Sidebar.vue
niubility000: Yes, and the current filebrowser also has this problem. When you route back from detail page(previewer) to list page (files or item listing), it always reload the list page, and then we can't come back to the exact location on the list page when we leave. that's inefficient. Hope we can fix it.

ramiresviana: The whole page is not reloaded when navigating from Preview back to Listing, in this case we have a partial reload where only information required to display the files is gathered from the server.

frontend/src/components/Sidebar.vue
ramiresviana: I will add it to my issues backlog and fix it when possible.

niubility000: Thank you so much.

frontend/src/components/Sidebar.vue
niubility000: oh, I got it. instead of using go(0), we need use:  "$store.commit('closeHovers')" , this will not reload the page.

PR #1591, opened 2021-10-02
show"upload process percentage" at the center of the screen.

### LINKED ISSUE #1590: Hope to let filebrowser display the transfer rate when uploading.

niubility000:
fixed #1590 
screen1
Screen2


o1egl:
I think the UX should be revisited

niubility000:
> I think the UX should be revisited

Yes, but I am not good at the design. -:)

PR #1596, opened 2021-10-04
Show "upload percentage" at the center of the screen. 

### LINKED ISSUE #1590: Hope to let filebrowser display the transfer rate when uploading.

niubility000:
fixed #1590

### DIFF
--- a/frontend/src/views/Layout.vue
+++ b/frontend/src/views/Layout.vue
@@ -9,7 +9,15 @@
       <shell v-if="isExecEnabled && isLogged && user.perm.execute" />
     </main>
     <prompts></prompts>
+
+    <div v-if="this.progress !==0" >
+       <div style=" height: 8rem;  width: 100%;  z-index: 99999;  position: fixed; top:40%; text-align:center; font-size:6em; "> {{ this.progress }}%</div>
+    </div>
+
   </div>
+
+
+  
 </template>
 
 <script>


niubility000:
Yes, it may look a little bit ugly. Who can make it better?  I am not good at it.

frontend/src/views/Layout.vue
```diff
@@ -9,7 +9,15 @@
       <shell v-if="isExecEnabled && isLogged && user.perm.execute" />
     </main>
     <prompts></prompts>
+
+    <div v-if="this.progress !==0" >
+       <div style=" height: 8rem;  width: 100%;  z-index: 99999;  position: fixed; top:40%; text-align:center; font-size:6em; "> {{ this.progress }}%</div>
```
mckaygerhard, line 14: i think is not so ugly.. the only hint is move the style to the css at `frontend/src/css/styles.css` and that's all

daiaji:
The percentage of uploads that are always placed in the center of the screen may distract users from doing other things.

mckaygerhard:
> The percentage of uploads that are always placed in the center of the screen may distract users from doing other things.

ok but i guess that know if upload are doing well is more important.. of course the percent can be moved away to an area of notification but will need more coding and this merge request is a collaboration.. 

so then you can propose some changes later

o1egl:
> > The percentage of uploads that are always placed in the center of the screen may distract users from doing other things.
> 
> ok but i guess that know if upload are doing well is more important.. of course the percent can be moved away to an area of notification but will need more coding and this merge request is a collaboration..
> 
> so then you can propose some changes later

Merging this PR will block the release process until we find a better UX solution for it. This is not acceptable

PR #1613, opened 2021-10-13
Search box border fixed

niubility000:
current:

fixed:

### DIFF
--- a/frontend/src/css/header.css
+++ b/frontend/src/css/header.css
@@ -88,7 +88,6 @@ header .menu-button {
   padding: 0em 0.75em;
   border-radius: 0.3em;
   transition: .1s ease all;
-
   align-items: center;
   z-index: 2;
 }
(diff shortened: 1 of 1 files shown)


frontend/src/css/header.css
```diff
@@ -84,7 +84,8 @@ header .menu-button {
 #search #input {
   background-color: #f5f5f5;
   display: flex;
-  padding: 0.75em;
+  height: 3em;
```
ramiresviana:
```suggestion
  height: 100%;
```

ramiresviana (changes requested):

frontend/src/css/header.css
niubility000: Done.

ramiresviana (approved):
This PR fixes the issue where search box is misaligned when the browser preferred font size is other than 16px.

PR #1614, opened 2021-10-13
Editor height fixed

niubility000:
**before:**
large bottom:

Can't see and edit the last line in mobile browser:

**after:**

### DIFF
--- a/frontend/src/css/styles.css
+++ b/frontend/src/css/styles.css
@@ -291,6 +291,8 @@ main .spinner .bounce2 {
 /* EDITOR */
 
 #editor-container {
+  display: flex;
+  flex-direction: column;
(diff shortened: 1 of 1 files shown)


ramiresviana (changes requested):
The relative height size causes inconsistency between different screen sizes. Using flexbox, we can have the breadcrumb height fixed with the editor filling the remaining space.

niubility000:
> The relative height size causes inconsistency between different screen sizes. Using flexbox, we can have the breadcrumb height fixed with the editor filling the remaining space.

Such a good point. Thank you! Fixed.

ramiresviana (approved):
This PR fixes the issue where editor does not fill the screen when the browser preferred font size is other than 16px.

PR #1631, opened 2021-10-20
Copy function fixed

niubility000:
The current copy function has two bugs:
1. When you copy a file or files to the current folder, it will not show you the conflict, and ask you to choose "rename" or "cancel". It just rename it and make a copy. 
2. When you copy a file or files to a different folder, and there are no same name file(s)，it will copy directly. But the prompt copy box doesn't close, and the button still can be clicked. It will confuse the users. They may think the copying process has not started.

This PR fixed both of them.


ramiresviana (commented):
> When you copy a file or files to the current folder, it will not show you the conflict, and ask you to choose "rename" or "cancel". It just rename it and make a copy.

When the user copy and paste a file on same folder, the intention is to duplicate that file, that's how it works on Windows file explorer. I see no advantage by adding another step to that process.

> When you copy a file or files to a different folder, and there are no same name file(s)，it will copy directly. But the prompt copy box doesn't close, and the button still can be clicked. It will confuse the users. They may think the copying process has not started.

I cannot reproduce this behavior. The Replace / Rename prompt is displayed only when there is a conflict with a filename. FileBrowser will check each filename when copying multiple files between folders and will prompt you once on the first occurrence of a conflict.

niubility000:
> > When you copy a file or files to the current folder, it will not show you the conflict, and ask you to choose "rename" or "cancel". It just rename it and make a copy.
> 
> When the user copy and paste a file on same folder, the intention is to duplicate that file, that's how it works on Windows file explorer. I see no advantage by adding another step to that process.
> 
Ok, I will put it back. 

> > When you copy a file or files to a different folder, and there are no same name file(s)，it will copy directly. But the prompt copy box doesn't close, and the button still can be clicked. It will confuse the users. They may think the copying process has not started.
> 
> I cannot reproduce this behavior. The Replace / Rename prompt is displayed only when there is a conflict with a filename. FileBrowser will check each filename when copying multiple files between folders and will prompt you once on the first occurrence of a conflict.

You misunderstand what I mean. I am saying when there is no conflict and the copying process starts, but the box is still there. It should auto hidden.  Test it with large files and you can see it clearly. The process alrealy starts, but the box still there. You still can click the copy and cancel buttons, but nothing happens. 
When you click the "copy" button, the box should disappear immediately and the copying process starts, just like the way when you copy it to the same dir(or click the rename/replace button when there is a conflict.), the box disappears immediately.

ramiresviana:
I see now, copying files causes the prompt to stay open until the process is finished. There's some problems by not blocking the user interaction when using the copy/move prompt:

- The user was no information if the process is still going.
- When the process finishes, the user is suddenly redirected to the destination folder.

mckaygerhard:
same problem happened to me of course only with big files.. mostly notable in slow connections (phones) 

> I see now, copying files causes the prompt to stay open until the process is finished. There's some problems by not blocking the user interaction when using the copy/move prompt:
> 
>     * The user was no information if the process is still going.

the information is on the upper corner right? as i can see here in this meerge request!

>     * When the process finishes, the user is suddenly redirected to the destination folder.

compared to a blocked ui this is dessirable.. rather a blocked ui..  @ramiresviana

niubility000:
How about disable the copy and cancel buttons when copying? or just show up a message box when copying? after copy process finished, just pop up a box to ask the user to switch to the new dir? because  here we use $route.push, it's very easy togo back to the origin dir by clicking 'back' button.

mckaygerhard:
ping @github-actions this is valid issue and i try to stay it alive

Issue #1380, opened 2021-04-20
Download / Upload Progress Bar with percentage completed


dualznz:
**Is your feature request related to a problem? Please describe.**
There is no way of knowing how far the download / upload progress is going 

**Describe the solution you'd like**
Some sort of progress bar maybe in a model that shows  on bottom right hand corner

ramiresviana:
Filebrowser has a progress bar for uploads that is displayed on top of the screen when uploading files. Downloads are handled by the browser, you can check the download progress on the downloads section of the browser you're using.

Frank4411:
> 
> 
> Filebrowser has a progress bar for uploads that is displayed on top of the screen when uploading files. Downloads are handled by the browser, you can check the download progress on the downloads section of the browser you're using.

@dualznz 
Here is an example of what you might see when uploading a file(see link below). I do think that it wouldn't hurt to make the bar a bit taller, maybe double the size? The first time I was uploading a file it took me a minute to figure out where the progress bar was.

[https://i.imgur.com/aQXuHel.png](https://i.imgur.com/aQXuHel.png)

o1egl:
I think better would be to create a small window which displays current queue. Something like that
<img width="399" alt="Screenshot 2021-04-23 at 13 54 57" src="https://user-images.githubusercontent.com/966539/115867550-8832fa00-a43b-11eb-9462-19801470ceb9.png">

dualznz:
yea something like that would make it more streamlined

sintaxx:
How do I enable the percentage indicator? are there plans to show the transfer rate as well ?

ramiresviana:
This feature was introduced on Filebrowser v2.21.0.

sintaxx:
> This feature was introduced on Filebrowser v2.21.0.

oh ok i guess i'll mess around with it some more but I just deployed this in a container v2.21.1 and saw no transfer speed or percentage indicator just a progress bar

Pendrag00n:
> This feature was introduced on Filebrowser v2.21.0.

A progress bar was introduced, but not actual transfer rate information. This would be pretty useful

zeezhao:
Please could you confirm if transfer rate addition is in the works? Thanks

PR #1632, opened 2021-10-21
Show upload progress percentage

### LINKED ISSUE #1590: Hope to let filebrowser display the transfer rate when uploading.

niubility000:
fixed #1590
fixed #1596
o1egl, how about it this time? It fits the default color design of filebrowser. 
And when you scroll, the bar will fix in the windows. you can do other operations.

PC:

.
.
Mobile:

### DIFF
--- a/frontend/src/css/base.css
+++ b/frontend/src/css/base.css
@@ -86,6 +86,7 @@ main {
 .breadcrumbs {
   height: 3em;
   border-bottom: 1px solid rgba(0, 0, 0, 0.05);
+  overflow: hidden;
 }
 
 .breadcrumbs span,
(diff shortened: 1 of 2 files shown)


ramiresviana:
I don't think it fits the current design. A suitable implementation would be a window for upload information as mentioned in https://github.com/filebrowser/filebrowser/issues/1380#issuecomment-825607117

mckaygerhard:
umm @ramiresviana and @niubility000 personally i prefers this event the thin bar that is not perceptible @o1egl

of course the windows upload progress as show in the comment is better but will be too complicated to make and we expected to have a better advertise of uploads SOON!

niubility000:
> I don't think it fits the current design. A suitable implementation would be a window for upload information as mentioned in [#1380 (comment)](https://github.com/filebrowser/filebrowser/issues/1380#issuecomment-825607117)
>
Yes. you mean if we choose multiple files at the same time and upload them, we wanna display every file's uploading progress. 
But we don't have those progress viarables currently for each file. I only can get the whole progress percentage from the source code.

PR #1677, opened 2021-11-22
Show pic's location

niubility000:
fixed: #1564
When you browse the pic, it will show the location where the pic was taken if the pic has the gps info in the exif.

I use windows-docker-ubuntu to compile the filebrowser, after I run "npm install --save exif-js", it can't generate the file "package-lock.json". I only submit package.json. So it can't pass the Checks. Please help me to generate the "package-lock.json"

------------------------------------------
You can try the version I modified. I've made a lot of changes, and have fixed #1224, #1225, #1233, #1631:
https://gitee.com/niubilitypdj/filebrowser/releases/
(It's very hard for me to upload large files to github, so I put them in gitee.)


o1egl:
Great job! I think we can make it extensible and have an info button that opens a popup with file-related metadata. So this information should be provided by the backend

niubility000:
> Great job! I think we can make it extensible and have an info button that opens a popup with file-related metadata. So this information should be provided by the backend

Yes! This api http://nominatim.openstreetmap.org/reverse?format=xml&lat=117.1111111&lon=37.1111111&zoom=18&addressdetails=0 works great!
Maybe we can store the address info in server's cache, then we don't need to fetch it from openstreetmap every time.  But I know nothing about goLang.

adrium:
That's a great idea and I was already thinking about that when I was implementing #1234.

I think it would be nice to provide the location already in the ```FileInfo struct``` for the listing so the list of files can also be displayed on a map.

What are your thoughts?

niubility000:
> That's a great idea and I was already thinking about that when I was implementing #1234.
> 
> I think it would be nice to provide the location already in the `FileInfo struct` for the listing so the list of files can also be displayed on a map.
> 
> What are your thoughts?

Sure, it's great if we can check the img's gps info from backend and show the locations in a map. and then we can show the location for the HD images (read from the original pics without downloading) which is hard to do in the frontend .

PR #1700, opened 2021-12-08
Fix encoding issue on password for shared files.

natchan0:
**Description**

At the moment, when using special characters in password for file sharing (e.g. accentuated letters "éèà"), it causes problems with encoding. When accessing the shared link and asked for the password, it always returns "Wrong credentials" even if the password is correct due to UTF-8/ISO-8859-1 incompatibilities.  
This PR fixes this issue.

### DIFF
--- a/frontend/src/api/pub.js
+++ b/frontend/src/api/pub.js
@@ -5,7 +5,7 @@ export async function fetch(url, password = "") {
   url = removePrefix(url);
 
   const res = await fetchURL(`/api/public/share${url}`, {
-    headers: { "X-SHARE-PASSWORD": password },
+    headers: { "X-SHARE-PASSWORD": unescape(encodeURIComponent(password)) },
   });
 
   if (res.status === 200) {
--- a/frontend/src/api/pub.js
+++ b/frontend/src/api/pub.js
@@ -3,9 +3,11 @@ import { baseURL } from "@/utils/constants";
 
 export async function fetch(url, password = "") {
   url = removePrefix(url);
+  let deiso = new TextDecoder("iso88591");
+  let enutf8 = new TextEncoder();
 
   const res = await fetchURL(`/api/public/share${url}`, {
-    headers: { "X-SHARE-PASSWORD": unescape(encodeURIComponent(password)) },
+    headers: { "X-SHARE-PASSWORD": deiso.decode(enutf8.encode(password)) },
   });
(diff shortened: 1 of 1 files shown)


frontend/src/api/pub.js
```diff
@@ -5,7 +5,7 @@ export async function fetch(url, password = "") {
   url = removePrefix(url);
 
   const res = await fetchURL(`/api/public/share${url}`, {
-    headers: { "X-SHARE-PASSWORD": password },
+    headers: { "X-SHARE-PASSWORD": unescape(encodeURIComponent(password)) },
```
o1egl: Looks like it is not recommended to use [unescape](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/unescape) function

natchan0:
I changed to use `TextDecoder` and `TextEncoder` instead of `unescape`.

This solves the issue for some characters like `é, è, ç, à` but some stil cause problems (`The server can't be reached`) like `€, Æ, É, È`. 

One solution would be to hash passwords in the JS before sending them to the server (to have only ASCII).   
Another solution would be to use a POST request and pass the password in the body (like when creating the share).

I read that sending passwords using GET requests is not safe, so the second solution might be better. I am no expert, so please verify this.

frontend/src/api/pub.js
```diff
@@ -3,9 +3,11 @@ import { baseURL } from "@/utils/constants";
 
 export async function fetch(url, password = "") {
   url = removePrefix(url);
+  let deiso = new TextDecoder("iso88591");
```
o1egl, line 6: On the server side, we support uft8 only. We can't use `iso88591` for that

o1egl: I think we can try to use https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURIComponent and decode it on the server side

o1egl:
I think we can use https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURIComponent and decode it on the server side https://go.dev/play/p/wkRvWPla3pJ

o1egl:
fixed in https://github.com/filebrowser/filebrowser/commit/c782f21b0fa4511a15e7015117d075eaf5ea332c

PR #1716, opened 2021-12-16
Users with different scopes will read the same cached files in the server. 

### LINKED ISSUE #1709: Can we reuse the cached HD images or thumbnails for different users?
Currently, different users will generate different cached HD images or thumbnails in the server's cache folder. 
For example, if I log in as user A, and browse some HD images, the server will generate these HD images at the first time and save them in the server's cache folder. 
Then, I logged in as user B, and wanna browse these HD images, it will genarate them again. But they are already generated for user A. I hope it can read the cached images directly. It will much faster for low-end server, and will save the space of server.  I think we do not need to genarate these HD images and cache them again and again for different users.

niubility000:
Fixed: #1709

Before:
Users with different scopes will generate different cache files  for the same image to the server's cache folder . Only the users with same scope can read the same cache file. 
This will cause the cache folder much larger and make the browse performance  worse for users with different scope when they browse it at the first time.

After: 
All users can read the same cached file for the same image. It will not be generated again and again for users with different scope.
-----------
You can try this first in my modified version. 
https://gitee.com/niubilitypdj/filebrowser/releases/


http/preview.go
```diff
@@ -152,5 +152,6 @@ func createPreview(imgSvc ImgService, fileCache FileCache,
 }
 
 func previewCacheKey(fPath string, fTime int64, previewSize PreviewSize) string {
+        fPath = filepath.Base(fPath)
```
ramiresviana: Filenames are not unique and can result in multiple files having the same cache key, the full path should be used instead.

niubility000: Yes, it's a bug. (rarely happens).
For example: There are two different images with same name 1.jpg, they are  a/b/1.jpg and a/c/1.jpg. And in filebrowser, just copy folder a to another place (Both images will have same modified time.), browse the images in the new place. the two 1.jpgs will show the same thumbnails. 
 
I have tried to use the fullpath: "d.user.FullPath(r.URL.Path)" instead of file.Path. 
and fmt.Println(d.user.FullPath(r.URL.Path)) will print different path for different users. The fullPath includes the previewSize (big or thumb) in it. and the position of it is different when the scope is different. Just like 
a/b/big/c/1.jpg when scope is /a/b, and /a/big/b/c/1.jpg when scope is /a. 
It's hard to delete this /big, because when the user really has a folder with name "big", and the assigned scope is /big, it will be a bug. 
What else can we use then? Or we can add the realsize of the image as another identifer.

o1egl: I agree with @ramiresviana. https://pkg.go.dev/github.com/spf13/afero#BasePathFs.RealPath should be used instead.

niubility000: Fixed again. I just use this one: d.user.Scope + fPath = realpath

For example:
test this image ./a/b/IMG_0027.JPG
Just run:
    fmt.Println(file.Path)
    fmt.Println(d.user.Scope)
    fmt.Println(d.user.Scope+file.Path)

for user with scope"./a", it's:
/b/IMG_0027.JPG
./a
./a/b/IMG_0027.JPG

for user with scope "." , it's:
/a/b/IMG_0027.JPG
.
./a/b/IMG_0027.JPG

d.user.Scope+file.Path can get the correct realpath. 
I think it's as good as BasePathFs.RealPath. 
And RealPath is a function, maybe will get a path like d.user.FullPath(r.URL.Path).

ramiresviana: What is the behavior of that implementation when the scope is empty?

niubility000: >What is the behavior of that implementation when the scope is empty?


I just fixed it. Using this:         fPath = filepath.Clean("/"+ d.user.Scope + fPath)
Filebrowser's root path can be "." or "/" or  "///" or ".///" or "/.//" or "\\\" or "abc" or empty ...

filepath.Clean is good enough to handle them.
In general, if the path starts with "." or "abc" or empty, filepath.Clean will give us a path like a/b/c... ,
if the path starts with "/ or " \ " , filepath.Clean will give us a path like /a/b/c...

It's very common we set the admin user's scope as ".", and set the common user's scope as "/a/b/". This will be a bug if only use "d.user.Scope + fPath".
So, we just add a "/" in front of the path, then we will get the path always like /a/b/c..., no matter what the scope starts with.

ramiresviana: You're making use of relative file paths. We need the real file path, relative to the OS filesystem root, to ensure that users will hit the correct cache entries and there will be no duplicates. Afero `RealPath()` should be able to solve it and ensure that will work on different platforms.

niubility000: Yes. ramiresviana, you mean it will be a bug when filebrowser runs in docker or something like that? I have tried this "filepath.Clean("/"+ d.user.Scope + fPath)" in windows and Ubuntu, it works.
I just don't know how to use this RealPath function. name string is the filename? It's not unique. (I know little about goLang.)
func (b *BasePathFs) RealPath(name string) (path string, err error) {
	if err := validateBasePathName(name); err != nil {
		return name, err
	}

	bpath := filepath.Clean(b.path)
	path = filepath.Clean(filepath.Join(bpath, name))
	if !strings.HasPrefix(path, bpath) {
		return name, os.ErrNotExist
	}

	return path, nil
}

func validateBasePathName(name string) error {
	if runtime.GOOS != "windows" {
		// Not much to do here;
		// the virtual file paths all look absolute on *nix.
		return nil
	}

	// On Windows a common mistake would be to provide an absolute OS path
	// We could strip out the base part, but that would not be very portable.
	if filepath.IsAbs(name) {
		return os.ErrNotExist
	}

	return nil
}

ramiresviana: I think we can have something like:
```go
func previewCacheKey(file *files.FileInfo, previewSize PreviewSize) (string, error) {	
	if fPath, err = file.Fs.(*afero.BasePathFs).RealPath(file.Path); err != nil {
		return nil, err
	}
	return fmt.Sprintf("%x%x%x", fPath, file.ModTime.Unix(), previewSize), nil
}
```
The usage should be:
```go
if cacheKey, err := previewCacheKey(file, previewSize); err != nil {
	return nil, err
}
```

niubility000: Fixed.
Now we just get the path like "D:\a\b\1.jpg", it's a realpath.

BTW, when I use "return nil, err", it fails to compile, nil is not an int or string. So I make some changes. I am not sure if it's correct. But it works.

http/preview.go
```diff
@@ -151,6 +155,10 @@ func createPreview(imgSvc ImgService, fileCache FileCache,
 	return buf.Bytes(), nil
 }
 
-func previewCacheKey(fPath string, fTime int64, previewSize PreviewSize) string {
-	return fmt.Sprintf("%x%x%x", fPath, fTime, previewSize)
+func previewCacheKey(file *files.FileInfo, previewSize PreviewSize) (string, error) {	
+	fPath, err := file.Fs.(*afero.BasePathFs).RealPath(file.Path)
```
o1egl, line 159: This cast will panic if file.Fs is not an instance of `afero.BasePathFs`. I would define a `RealPath` method on `files.FileInfo` which returns a `BasePathFs.RealPath` if fs is an instance of `BasePathFs` or file.Path as a fallback

niubility000: > This cast will panic if file.Fs is not an instance of `afero.BasePathFs`. I would define a `RealPath` method on `files.FileInfo` which returns a `BasePathFs.RealPath` if fs is an instance of `BasePathFs` or file.Path as a fallback

o1egl, would you please make a PR for this? It's beyond my ability in golang.

niubility000: Personally, I will use this one:
func previewCacheKey(file *files.FileInfo, previewSize PreviewSize) string {	
	return fmt.Sprintf("%x%x%x", filepath.Base(file.Path), file.ModTime.Unix(), previewSize, file.Size)
I think filename & modifed time & file size is enough to tell different images. and the good thing of using this is: when we move or rename a folder with tons of images, it will not regenerate the cache files.

o1egl:
fixed in https://github.com/filebrowser/filebrowser/commit/c1987237d05adcce77c614e5247a181ae5cdfacd

PR #1723, opened 2021-12-19
Detect multiple subtitle languages

### LINKED ISSUE #1702: Detect multiple subtitle languages
It would be nice to support subtitle files (i.e. `.vtt`) for **multiple languages** by detecting filename with language codes like `videoname.en.vtt` as shown in this [TODO](https://github.com/filebrowser/filebrowser/blob/aa52b07bb169e53bc1b656cbf88262e808c10ae3/files/file.go#L274) for the native video player.

mazenbesher:
Resolves #1702 as suggested in [this comment](https://github.com/filebrowser/filebrowser/issues/1702#issuecomment-991083038).


files/file.go
```diff
@@ -271,7 +272,23 @@ func (i *FileInfo) detectSubtitles() {
...
 
-	// TODO: detect multiple languages. Base.Lang.vtt
+	// detect multiple languages. Base.Lang.vtt
+	// TODO: give subtitles descriptive names (lang) and track attributes
+	afs := &afero.Afero{Fs: i.Fs}
+	parentDir := strings.TrimRight(i.Path, i.Name)
+	dir, err := afs.ReadDir(parentDir)
+	if err == nil {
+		base := strings.TrimSuffix(i.Name, ext)
+		r := regexp.MustCompile(base + `\.(.*\.)?vtt`)
```
ramiresviana: Use `QuoteMeta` to prevent specials characters in the filename being evaluated. Asserting the start and the end of the string prevents invalid matches with filenames contaning a suffix or prefix.

```suggestion
		r := regexp.MustCompile("^" + regexp.QuoteMeta(base) + `\.(.*\.)?vtt$`)
```

ramiresviana (changes requested):

files/file.go
```diff
@@ -271,7 +272,23 @@ func (i *FileInfo) detectSubtitles() {
 	i.Subtitles = []string{}
 	ext := filepath.Ext(i.Path)
 
-	// TODO: detect multiple languages. Base.Lang.vtt
+	// detect multiple languages. Base.Lang.vtt
+	// TODO: give subtitles descriptive names (lang) and track attributes
+	afs := &afero.Afero{Fs: i.Fs}
```
o1egl: Why the original fs is wrapped?

files/file.go
```diff
@@ -271,7 +272,23 @@ func (i *FileInfo) detectSubtitles() {
...
+	parentDir := strings.TrimRight(i.Path, i.Name)
+	dir, err := afs.ReadDir(parentDir)
+	if err == nil {
+		base := strings.TrimSuffix(i.Name, ext)
+		r := regexp.MustCompile("^" + regexp.QuoteMeta(base) + `\.(.*\.)?vtt$`)
```
o1egl: `MustCompile` must not be used for user-provided data. Use `Compile` instead. I think we can use a simple substring check which would be ~200 times faster than regexp https://go.dev/play/p/HOZBYFaGH1Z

files/file.go
mazenbesher: No reason other than my illiteracy. I guess `afero.ReadDir(i.Fs, parentDir)` should be the correct way?

o1egl: Yes

files/file.go
mazenbesher: I agree. This would eliminate the restriction to `Base.Lang.vtt`.

PR #1743, opened 2021-12-28
Show video seek bar in mobile browser

### LINKED ISSUE #1724: No seek bar for videos on mobile browsers
There doesn't seem to be a seek bar for videos on mobile browsers. I tried Firefox for Android and Chrome for Android. On Chrome you can pause and unpause videos. On Firefox you can't. Neither had a seek bar for videos. Does anyone else have this problem?

niubility000:
fixed: #1724
When the address bar of the browser is hidden, we can see the seek bar of the video. 

But when the address bar is not hidden, we can't see the seek bar. 

==================
after the change:


armandleopold:
Yes very anoying bug !

o1egl:
I think there should be a better solution instead of setting the height for all the browsers. Also when the panel is hidden, the media bar jumps up to the height of the top bar.

Issue #1698, opened 2021-12-07
FEATURE REQUEST: Folder and File Icon Differentiation


hernandito:
Hi.

I would love to see better visual differentiation between files and folders. Being that the folder and file icons are the same color and look very close to each other; please add option to better distinguish... a quick Photoshop below illustrates what I suggest:
**BEFORE:**
![FB-Before](https://user-images.githubusercontent.com/787317/144958145-4129c670-8a98-4038-81ee-2c70baf0d123.jpg)

**AFTER:**

![FB-After2](https://user-images.githubusercontent.com/787317/144960147-67fc7bca-9774-4a78-bf66-c2872a44e6b7.jpg)
Similar to how in any OS it's easy to distinguish between a folder and a file.  

If you strongly object to adding colors, at a bare minimum, please use "**view_headline**" instead of "_insert_drive_file_" for the font icon:

![image](https://user-images.githubusercontent.com/787317/144959070-e87e2eb6-08a8-4167-b665-300b8cc77172.png)

This should be a very easy change... I hope you will consider this.

Thank you!

jniggemann:
Filebrowser is really great! My feeling is though, that many good feature requests go unoticed and as such, are not implemented. This is one of them, most probably an easy fix that greatly helps vision impaired people and all other users, too.

hernandito:
> Filebrowser is really great! My feeling is though, that many good feature requests go unoticed and as such, are not implemented. This is one of them, most probably an easy fix that greatly helps vision impaired people and all other users, too.

Thank you...! It is indeed awesome.

I have tried editing the .vue files, but for the life of me, I have not been able to build the single binary executable with npm and go. At the end, the is no new "filebrowser" file to be found. I have spent 3 days trying. 

I created a new docker container (with an unRAID template as well) to make the user settings (and custom.css) persistent.

hernandito:
After many days trying to recompile… without success, I just found this post that achieves exactly what I propose……

https://github.com/filebrowser/filebrowser/discussions/1256

Thank you to snokin!!!

niubility000:
I have written a doc about how to compile filebrowser in windows pc. But it's in Chinese. 
https://gitee.com/niubilitypdj/filebrowser/releases/
If you need, I can translate it to English when I have time.

niubility000:
@hernandito, You could start a new topic to mention that you can't compile the filebrowser. Everyone would notice it earlier. Then you don't need to try many days.

[...]

Issue #1668, opened 2021-11-15
Use images instead of fonts for the icons


DUOLabs333:
**Is your feature request related to a problem? Please describe.**
I force the same font for all websites for usability purposes, and that obviously makes this unusable.

**Describe the solution you'd like**
Instead of using text icons, maybe use images

**Describe alternatives you've considered**
You can use Unicode for some of them

**Additional context**

o1egl:
We don't plan to use images. It will slow down the frontend. You can create custom CSS styles and override the default theme https://filebrowser.org/configuration/custom-branding

PR #1769, opened 2022-01-12
Add colorized file type icons

adrium:
I stumbled upon issue #1698. The listing view has made it difficult for me too finding files that are not images, and it inspired me to implement this.

The general approach is to fully define icon rendering in CSS.
Icons are based on the type and file suffix, colors are based on simple definitions, and the palette is from [KDE](https://develop.kde.org/hig/style/color/default/). I tried to include the most relevant types considering modern operating systems, formats, and languages. In the end though, it is subjective what should be considered relevant...

Also related #1256 and #1668.

The colors can be disabled by adding the following [css](https://filebrowser.org/configuration/custom-branding):

```css
#listing i { color: inherit !important }
```

- [x] continuous integration build

**Example:**

preview

### DIFF
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -15,7 +15,7 @@
         "js-base64": "^2.5.1",
         "lodash.clonedeep": "^4.5.0",
         "lodash.throttle": "^4.1.1",
-        "material-design-icons": "^3.0.1",
+        "material-icons": "^1.10.5",
         "moment": "^2.24.0",
         "normalize.css": "^8.0.1",
         "noty": "^3.2.0-beta",
@@ -8962,10 +8962,10 @@
         "node": ">=0.10.0"
       }
     },
-    "node_modules/material-design-icons": {
-      "version": "3.0.1",
-      "resolved": "https://registry.npmjs.org/material-design-icons/-/material-design-icons-3.0.1.tgz",
-      "integrity": "sha1-mnHEh0chjrylHlGmbaaCA4zct78="
+    "node_modules/material-icons": {
+      "version": "1.10.5",
+      "resolved": "https://registry.npmjs.org/material-icons/-/material-icons-1.10.5.tgz",
+      "integrity": "sha512-+4l0xbfKlA2eSXJrhISYvCvT8e9HqQk3BLdOCPGal5qPYAQoTI8TTacYI0yXyf7WS6BpKrhQemUenhmR0IoS8A=="
     },
     "node_modules/md5.js": {
       "version": "1.3.5",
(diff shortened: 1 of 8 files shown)


o1egl:
@adrium well done! Let's wait for community feedback

adrium:
Thanks for the feedback - glad you like it! I actually got some ideas for slight improvements such as less opacity for hidden files. I will update the code within the next couple of days...

niubility000:
Is there a way to set this one as a theme? Currently we have "light","dark", maybe we can add this one as a new theme. So users can make their own choice.  
Or just let the users to customize their own theme(or skin), including these icons and colors. Then everyone will be satisfied. see here:#1256

adrium:
The intention is to set the icons and their colors as defaults - like operating systems that provide a default icon theme.
It works well for the light and dark theme. The screenshot is updated in the PR description.

adrium:
I updated some details:

- Sort definitions by color and alphabet in the stylesheet
- Selected icons are always white
- Hidden and .bak items have less opacity
- Add some more types

adrium:
Updated with minor last changes:

- CSS selector `i` instead of `.material-icons`
- Remove unnecessary `:class="type"` in `<i>`

ramiresviana:
The behavior i expect of files that are not supported by the system is them to be displayed as a blank file icon. FileBrowser has preview only for text, pdf, image, audio and video files, i think that adding specific colors and icons for other file types can lead to confusion.

We could try implementing this gradually and checking the user feedback.

adrium:
Hi Ramires, thanks for your feedback. I generally like your suggestion of having icons for types that are "supported by the system".

Let's take a closer look:

1. File Browser determines file types based on the go mime module, specifically by [file extension](https://pkg.go.dev/mime#TypeByExtension). This is afaik how all popular operating systems handle file identification. Only, if the extension is unknown, it tries to [detect it from the content](https://github.com/filebrowser/filebrowser/blob/v2.20.1/files/file.go#L206).
However, it depends on the system where File Browser is installed. On my server, the file type of `go.mod` is `application/xml-dtd` due to the extension. However, File Browser will still open it as text despite being classified in the broad `application/*` type. It seems to inspect the file in this case and probably `isBinary(buffer)` returns false.
When I run File Browser on my desktop, the same file is `audio/x-mod` and has a speaker icon because VLC is installed and modifies the file type library. When opening the file in File Browser, a seek bar is displayed and nothing useful can be done with the file which is in fact text.
**File Browser does not handle file identification robustly.**

2. Related to the above, if there is a `sound.wma` file, it is most likely a Windows Audio file. On any Debian based system, this will be recognized as `audio/x-ms-wma` because of the `media-types` package. If File Browser is run on Windows, it will also be recognized as an audio file. Again, when opening the file in File Browser, a seek bar is displayed but the file can not be played back because it is not supported by the browser.
**File preview depends on Browser support.**

3. Suppose the clicks on a `sound.opus` file. It can be previewed in File Browser successfully thanks to browser support. However, if you download to file on Windows 10, you can not open it anymore, because there is no application registered for it (as opposed to webp which is opened by Edge). Even if the file can easily be opened in the web UI, it may cause difficulties on the client when downloaded.
**Files can be more difficult to open when downloaded.**

There is already much potential for confusion. On the other hand, if File Browser (as per readme a "file managing interface") gives a visual hint of what the file is supposed to be, it helps to manage files - just think of the _Downloads_ folder. It is up to the user to install programs that can handle the files on a desktop. If File Browser is even able to provide a preview - perfect! If it can not, a nice prompt to download the file is displayed - no confusion there.

This why I think icons and colors should be provided independent of preview, by default and without gradual changes.

jniggemann:
> We could try implementing this gradually and checking the user feedback.

I welcome this and think it is a good idea. I've gotten a lot of feedback for the CSS I published in #1256. There are definetly a lot of people out there that want to distinguish files more easily using colorized icons .

PR #1777, opened 2022-01-17
Add gallery view mode

adrium:
**Description**

When organizing pictures, the "mosaic" view mode is still a bit too small.
This PR adds new gallery view mode that focuses on pictures.

- [x] continuous integration build

**Further comments**

- Thumbnail size is increased from 128x128 to 256x256
- Columns are a bit smaller to fit 2 columns in landscape mobiles

Screenshot:

gallery-mode

### DIFF
--- a/frontend/src/css/listing.css
+++ b/frontend/src/css/listing.css
@@ -110,6 +110,41 @@
   width: calc(100% - 5vw);
 }
 
+#listing.mosaic.gallery .item div:first-of-type {
+  width: 100%;
+  height: 12em;
+}
+
+#listing.mosaic.gallery .item div:last-of-type {
+  position: absolute;
+  bottom: 0.5em;
+  padding: 1em;
+  width: calc(100% - 1em);
+  text-align: center;
+}
+
+#listing.mosaic.gallery .item[data-type=image] div:last-of-type {
+  color: white;
+  background: linear-gradient(#0000, #0009);
+}
+
+#listing.mosaic.gallery .item i {
+    width: 100%;
+    margin-right: 0;
+    font-size: 8em;
+    text-align: center;
+}
+
+#listing.mosaic.gallery .item img {
+  width: 100%;
+  height: 100%;
+}
+
+#listing.gallery .size,
+#listing.gallery .modified {
+  display: none;
+}
+
 #listing.list {
   flex-direction: column;
   width: 100%;
--- a/frontend/src/views/files/Listing.vue
+++ b/frontend/src/views/files/Listing.vue
@@ -283,6 +283,7 @@ export default {
   data: function () {
     return {
       showLimit: 50,
+      columnWidth: 280,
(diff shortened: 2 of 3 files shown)


niubility000:
It's just like what I did here:
https://github.com/filebrowser/filebrowser/issues/1709#issuecomment-992821302

adrium:
Fixed lint errors in new commits

niubility000:
> Fixed lint errors in new commits

Test my version here:
http://ppddjj.tpddns.cn:2008    share:share
http://ppddjj.tpddns.cn:2008/share/NiOzqIQV   (add preview for shared files.)

niubility000:
> Fixed lint errors in new commits

Test my version here:
http://ppddjj.tpddns.cn:2008    share:share
http://ppddjj.tpddns.cn:2008/share/NiOzqIQV   (add preview for shared files.)

o1egl:
It looks great for images, but do you think we can improve it for regular files and dirs?
Screenshot 2022-02-08 at 20 20 38

adrium:
The idea is to explicitly enable gallery mode only in folders with many pictures, but files could look better indeed...

The tile size would need to stay the same in order to line up propery. The style could use some finetuning.

How do you like this?

filebrowser-gallery

o1egl:
It looks cleaner. Can you push those changes, so we can merge this pr

adrium:
Thanks for the feedback Oleg. I am preparing the changes.

adrium:
Oleg, working on a new feature, I noted that I should have added the gallery view mode here:
https://github.com/filebrowser/filebrowser/blob/v2.20.1/users/users.go#L19

Sorry for missing this. How do you want to proceed? Should I open a new PR or can we reopen this one and do a fixup?

niubility000:
The problems are:
1. The thumbnails we get from the exifs are not clear enough to show in  256x256 format. It may look ugly. But If we don't use the exif to get the thumbnails, the listing speed will be very slow in low-end server (e.g.  a router) for the first time.   
2. As I tested, If we wanna show thumbnails in  256x256 clearly, we need a 320*320 high quality thumbnails.

So my solution is: 
Just keep the original version, and add an option to switch to "Large thumbnails".  For these large thumbnails, we can use all 320*320 high quality thumbnails (don't use the thumbnails from exif).


And to show these 320*320 high quality thumbnails faster at the first time, we need add a function to generate all these thumbnails to server's cache in advance. I hided this function in the search-image for admin users.




Another problem is: the current listing is slow when showing large number of pics. we need to swipe the list down again and again to show them all, and the DOM becomes larger and larger. I think we need to use "virtual list" to make it perform better. 

I have solves all the above problems in my modified version. Feel free to modify and use these source code.
[filebrowser.zip](https://github.com/filebrowser/filebrowser/files/8045494/filebrowser.zip)

By the way, the following is the large thumbnails I used. They are much clearer.

PR #1825, opened 2022-02-14
feat: add upload file list with progress

filippofinke:
**Description**
Add the list of files that are being uploaded and their relative progress.

Reference to issue #1380

out


ramiresviana:
While it looks functional, the design is in a rough state and needs improvement. The progress indicator takes up a large part or even all of the screen space, depending on the number of files being uploaded. We need a compact design that doesn't get in the way of users and fits different screen sizes.

o1egl:
I'd suggest something like that one https://github.com/filebrowser/filebrowser/issues/1380#issuecomment-825607117 . Ideally, this window should be minimized and expanded when the user clicks on it

filippofinke:
> I'd suggest something like that one [#1380 (comment)](https://github.com/filebrowser/filebrowser/issues/1380#issuecomment-825607117) . Ideally, this window should be minimized and expanded when the user clicks on it

Thank you for the feedback.

I'll work on it as soon as I have time.

filippofinke:
@ramiresviana @o1egl 

I have updated the design, still some work has to be done.

out




It's giving errors because my server disk is full.

ramiresviana:
Please make sure that you're using the project ESLint rules, making the style of your code consistent with the current codebase.

When there's a upload in progress, users cannot upload more files since the header action shows only the progress prompt.

filippofinke:
> Please make sure that you're using the project ESLint rules, making the style of your code consistent with the current codebase.
> 
> When there's a upload in progress, users cannot upload more files since the header action shows only the progress prompt.

I have fixed the code style. 

I agree that users can't upload more files because of the progress prompt.

I'll try to create a prompt on the bottom right that doesn't block user navigation and that can be maximized/minimized.

Something like this one:

image

filippofinke:
I am currently working on the icons on the left side of the file name, some other designs improvement and then I'll do some refactoring.

Current status:
out

filippofinke:
Added icons:

image

PR #1850, opened 2022-03-05
Feature/1849/windowtitle

### LINKED ISSUE #1849: add branding.name to window title

I would like to be able to use this application with my password manager that uses window detection based on the window title. Filebrowser does not have a unique windowtitle (Login is the default) and therefore it does not detect the correct entry.

The most straigthforward option would be to add branding.name to the windowtitle if set. So the default would still be 'Login', but if branding was enabled it would be 'Login - branding.name'. It would also be an option to always make the windowtitle 'Login - Filebrowser' and only change Filebrowser to the branding if set.

newhinton:
**Description**
adds branding to document.title for branding/passwordmanager-use
closes #1849

### DIFF
--- a/frontend/src/views/Settings.vue
+++ b/frontend/src/views/Settings.vue
@@ -50,9 +50,12 @@
 </template>
 
 <script>
-import { mapState } from "vuex";
+import {mapMutations, mapState} from "vuex";
 
 import HeaderBar from "@/components/header/HeaderBar";
+import {
+  name,
+} from "@/utils/constants";
 
 export default {
   name: "settings",
@@ -61,6 +64,25 @@ export default {
   },
   computed: {
     ...mapState(["user", "loading"]),
+    name: () => name,
   },
+  created() {
+    this.fetchData();
+  },
+  watch: {
+    $route: "fetchData",
+    reload: function (value) {
+      if (value === true) {
+        this.fetchData();
+      }
+    },
+  },
+  methods: {
+    ...mapMutations(["setLoading"]),
+    async fetchData() {
+        document.title = document.title + ` - ${this.name}`;
+    },
+  },
+
 };
 </script>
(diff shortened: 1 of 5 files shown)


frontend/src/views/Settings.vue
```diff
@@ -61,6 +62,24 @@ export default {
...
+      if (value === true) {
+        this.fetchData();
+      }
+    },
+  },
+  methods: {
+    ...mapMutations(["setLoading"]),
+    async fetchData() {
+      document.title = document.title + ` - ${this.name}`;
+    },
```
o1egl: Why is this added?

newhinton: For consistentcy

it is not required to solve my linked issue, but without it we have different titles:

Login - Filebrowser
Files - /Path
Settings

with it we always append the title


Or do you mean the watch-function etc?

o1egl: The watch function

newhinton: Switching tabs in the settings also updates the title. If the watch-function is omitted, the instance-name only gets added to the initial load of the settings, but switching tabs removes it again

ramiresviana:
Instead of manipulating the document title on every view, a better approach would be setting the title correctly on the first place. This can be achieved by changing this line:

https://github.com/filebrowser/filebrowser/blob/b2ad3f73686a2abaa4fc62963fba6f83c9da9b5e/frontend/src/router/index.js#L143

newhinton:
@ramiresviana That does not seem to work properly. While you can briefly see the changed name (using `document.title = to.name + "TEST";`) it gets overriden as soon as the page is finished loading. 

If there is a better way to do this i am all ears, but this seemed the most straitforward way to me, especially for the file-view

Edit: it seems to be working partially, but not for the file-view because that overrides its title depending on the path. I will investigate further.

newhinton:
@ramiresviana Thanks for your suggestion!

PR #1883, opened 2022-03-28
feat: add new event supporting for the action login.

hunt978:
**Description**

**Further comments**

I want to create a soft link to my users automatically for sharing common file storage among them as FTP does, by executing commands before or after certain events. While I cannot find the suitable one from the default sets. Compared with events like 'upload', 'save', 'remove', etc. 'USER LOGIN' is a more promising event for such needs.

Brokenly, this approach was summited to address the issue here: `printToken` has a different return type with the callback `RunHook` required. So, my implementation of `event_login` *DOES NOT* obeys the current rule of 'before' and 'after' schema, which is the same actually.  But it is fine to me, and I am also very glad to take advice to make things right.

Furthermore, I have considered the next feature following the thoughts of `common share storage` that maybe there should be a user interface for the maintainer to add a list of folders to share among users and set up each every folder read/write attributes also.


o1egl (approved):

o1egl (changes requested):
Please fix lint errors

PR #1930, opened 2022-04-27
A few improvements | theme color

ij369:
Change the theme color by media query.


ramiresviana (commented):
The dark theme requires additional css to work and you removed the support for custom theme color.

PR #1942, opened 2022-05-06
Some fixes

### LINKED ISSUE #1839: You cannot directly drag and drop folders to upload, and you cannot upload multiple folders at the same time.
**Description**

You cannot directly drag and drop folders to upload, and you cannot upload multiple folders at the same time.

**Expected behaviour**

Drag and drop folders to upload directly, upload multiple folders at the same time

Uploading folders is harder than before.

**Additional context**

Without.

**How to reproduce?**

Anytime on v2.21.1.

**Files**

Without.

### LINKED ISSUE #1919: You don't have permissions to access this - after some hours of uploading large file
**Description**

Upload a large file to my NAS using NGINX Reverse Proxy

**Expected behaviour**

Upload any file size with any speed

Upload is interrupted, message: "You don't have permissions to access this."

**Additional context**

I use NGINX as reverse proxy for my NAS. NGINX has buffering turned off to pass on the file traffic to the upstream right away. No issue when I upload smaller files that takes less time to upload.

No exact amount of time is measured, but after some hours of non interaction with FileBrowser webUI (while uploading) the page shows "You don't have permissions to access this.". Same file can be uploaded with no issue if I interact with the page in some way or from additional tab.

I suspect that FileBrowser kicks out the session and it stop 'sending' the upload, although no error message is shown on the page, it still keeps showing that it is uploading. If the webUI is refreshed, "You don't have permissions to access this." message shows.

NOTE: The PC with the upload webUI does not go to sleep or the connection doesn't drop in any way.

**How to reproduce?**

Upload large file that takes many hours, and don't interact with the FileBrowser webUI in any way.
In my case, this is cause due to high latency. My VPN has about 130ms latency, which causes slow upload. But this is not relevant as FileBrowser should upload no matter the latency or the time it takes to upload any given size of a file.

**Files**

N/A

ramiresviana:
### chore: refactor url creation
Removed the logic for URL creation from the views and created reusable functions. Should prevent inconsistency of enconding and the URL structure.

### fix: drag-and-drop folder upload
Uploading folder via drag-and-drop was not working and throwing silent errors because of file type detection. Since folders do not require the type information, such field is now omitted for folders. fixes #1839

### fix: folder info on upload list
Folders are displayed as empty items on the uploading file list. The required information is now gathered and displayed correctly.

### chore: refactor response error handling
Response errors are handled by having logic on every API function, even that it works in the same way for all of them. Errors should be handled in a general way, improving maintainability.

### fix: network error object message
Network errors causes notifications with vague object messages to be shown. Custom error codes are added to make easier to identify problems caused by network.

### fix: expired token error
Forbidden error messages are displayed when the auth token is expired. The user is now redirected to the login page when the token was expired. fixes #1919

### feat: page title localization
Page titles are using hard coded names, even that localization is supported. Existing values were used to avoid creating a new category only for the titles.

### feat: invalid symlink icon
When the backend cant follow a symlink, it is displayed as a regular file. Such items are now displayed with custom icon and message to make clear for the users.

### fix: allow CSP inline styling
Somes browsers uses custom CSS when opening files on a new tab. The CSP was adjusted to allow these styles.

### DIFF
--- a/files/file.go
+++ b/files/file.go
@@ -322,7 +322,7 @@ func (i *FileInfo) readListing(checker rules.Checker, readHeader bool) error {
 			continue
 		}
 
-		isSymlink := false
+		isSymlink, isInvalidLink := false, false
 		if IsSymlink(f.Mode()) {
 			isSymlink = true
 			// It's a symbolic link. We try to follow it. If it doesn't work,
@@ -330,6 +330,8 @@ func (i *FileInfo) readListing(checker rules.Checker, readHeader bool) error {
 			info, err := i.Fs.Stat(fPath)
 			if err == nil {
 				f = info
+			} else {
+				isInvalidLink = true
 			}
 		}
 
@@ -350,9 +352,13 @@ func (i *FileInfo) readListing(checker rules.Checker, readHeader bool) error {
 		} else {
 			listing.NumFiles++
 
-			err := file.detectType(true, false, readHeader)
-			if err != nil {
-				return err
+			if isInvalidLink {
+				file.Type = "invalid_link"
+			} else {
+				err := file.detectType(true, false, readHeader)
+				if err != nil {
+					return err
+				}
 			}
 		}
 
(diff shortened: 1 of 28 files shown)


o1egl (approved):

adpavlov:
dear maintainers, when is it going to be released?

PR #1950, opened 2022-05-13
Encrypt the password obtained from the flags

VolanDeVovan:
The password set via flags (--password) does not work because it must be encrypted
I don't know if this is a bug, but I decided to fix it, since the description of the flag does not indicate that it should be encrypted


o1egl:
As mentioned in the `password` flag's description, it should be provided as a hashed string, not a plaintext. The `filebrowser hash my-secret-password` command can be used to hash the password

Issue #1970, opened 2022-06-05
baseurl doesn't re-map all api urls?


jaz000:
**Description**
filebrowser runing in a docker container
using a reverse proxy (nginx) to access it
using FB_BASEURL env  or settings json file to set base url to "/filebrowser"

**Expected behaviour**
perfection :-s

**What is happening instead?**
So most things are working, it generally looks good, except eventually I notice image preview and playing video files are not working

**Additional context**
looking at urls loaded browser in dev mode:

displaying a directory gives:

https://myserver.net/filebrowser/api/usage/mydirectory/
https://myserver.net/filebrowser/api/resources/mydirectory/

etc.
all looks good so far but then the thumbnails (fail with 404 through proxy or they work when accessed directly):

https://myserver.net/api/preview/thumb/mydirectory/file-fanart.jpg?auth=<token>&inline=true&key=1642449078616
https://myserver.net/api/preview/thumb/tv-japan/Angel%20Heart%20(2015)/poster.jpg?auth=<token>&inline=true&key=1642477709602

note: the /api at the root, it's missing the base /filebrowser 
playing a video gives the same:

https://myserver.net/api/raw/mydirectory/mydirectory/myfile.mkv?auth=<token>&inline=true

**How to reproduce?**

It does this when accessed directly - it doesn't need to go through a reverse proxy to reproduce. Just set the baseurl option and look at the urls (those not remapped will work when accessed directly, just not through the proxy).

Possibly in frontend/src/api/files.js  the functions like getPreviewURL / getDownloadURL need baseURL added

stahiga:
same problem with v2.22.0 and downgrading it back to v2.21.1 temporarily solves it. note that this upgrade breaks the db structure which means it's not backward compatible so you need to reconfigure from start.

o1egl:
> same problem with v2.22.0 and downgrading it back to v2.21.1 temporarily solves it. note that this upgrade breaks the db structure which means it's not backward compatible so you need to reconfigure from start.

v2.22.0 is backward compatible version and doesn't break the DB

PR #1971, opened 2022-06-05
Fix createURL

jisqyv:
The Javascript URL object takes a second argument which is the base to
use for the URL. However it only accepts the "root" of the base. So
arguments such as http://localhost/example get silently turned into
http://localhost/

This change places the baseURL within the path element, after making
sure that it ends with a slash (/).


jisqyv:
Fixes #1970

Issue #1972, opened 2022-06-06
500 Internal Server Error 


hungxqt:
**Description**
500 Internal Server Error box when viewing an image or video.

**Expected behaviour**
No error box appears when viewing.

**What is happening instead?**
Red box with "500 Internal Server Error" text.
"500 192.168.193.191 The directory name is invalid" in terminal.

**Additional context**
![image](https://user-images.githubusercontent.com/75252773/172189232-dfda5595-c979-4eef-a2f8-42f155a71450.png)
![image](https://user-images.githubusercontent.com/75252773/172190096-5acd1e0c-9750-4b24-8644-40dc8b5ba1d9.png)

Issue #1967, opened 2022-06-03
Frontend Vue error


langren1353:
**Description**
\filebrowser\frontend\src\components\Sidebar.vue:88
param large is not defined on the instance but referenced during render.
![image](https://user-images.githubusercontent.com/14857127/171922569-a94c578d-cd64-4cfe-ad40-7cf4e67756ec.png)

PR #1973, opened 2022-06-06
fix: use incorrect suffix and return no 500(#1972、#1967)

langren1353:
When watch Video at filebrowser, there will be a `500` server error at right bottom of the page #1972 
The reason is SideBar.vue will check disk useage when route changes, but when path is a file, server will return a `500` error
Before:
    file: 500 error
    dir: true useage
After:
    file: 0 useage
    dir: true useage

### DIFF
--- a/frontend/src/components/Sidebar.vue
+++ b/frontend/src/components/Sidebar.vue
@@ -85,7 +85,7 @@
       v-if="$router.currentRoute.path.includes('/files/')"
       style="width: 90%; margin: 2em 2.5em 3em 2.5em"
     >
-      <progress-bar :val="usage.usedPercentage" :size="large"></progress-bar>
+      <progress-bar :val="usage.usedPercentage" size="large"></progress-bar>
       <br />
       {{ usage.used }} of {{ usage.total }} used
     </div>
@@ -102,9 +102,9 @@
         >
         <span> {{ version }}</span>
       </span>
-      <span
-        ><a @click="help">{{ $t("sidebar.help") }}</a></span
-      >
+      <span>
+        <a @click="help">{{ $t("sidebar.help") }}</a>
+      </span>
     </p>
   </nav>
 </template>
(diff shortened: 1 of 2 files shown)


frontend/src/components/Sidebar.vue
```diff
@@ -85,7 +85,7 @@
       v-if="$router.currentRoute.path.includes('/files/')"
       style="width: 90%; margin: 2em 2.5em 3em 2.5em"
     >
-      <progress-bar :val="usage.usedPercentage" :size="large"></progress-bar>
+      <progress-bar :val="usage.usedPercentage" size="large"></progress-bar>
```
ramiresviana:
```suggestion
      <progress-bar :val="usage.usedPercentage" size="small"></progress-bar>
```

ramiresviana (changes requested):
Retrieving the usage by providing a path to a file causes this error on Windows. I think the proposed solution is good for now.

http/resource.go
```diff
@@ -351,12 +351,19 @@ var diskUsage = withUser(func(w http.ResponseWriter, r *http.Request, d *data) (
...
+		return renderJSON(w, r, &DiskUsageResponse{
+			Total: usage.Total,
+			Used:  usage.Used,
+		})
+	} else {
+		return renderJSON(w, r, &DiskUsageResponse{
+			Total: 0,
+			Used:  0,
+		})
 	}
```
o1egl:
```suggestion
	if !file.IsDir {
		return renderJSON(w, r, &DiskUsageResponse{
			Total: 0,
			Used:  0,
		})
		
	}
	
	usage, err := disk.UsageWithContext(r.Context(), fPath)
	if err != nil {
		return errToStatus(err), err
	}
	return renderJSON(w, r, &DiskUsageResponse{
		Total: usage.Total,
		Used:  usage.Used,
	})
```

PR #2050, opened 2022-07-21
An extension for Aria2 and Youtube-dl

shiningw:
Hi,
I have created [an extension for filebrowser ](https://github.com/shiningw/yadownloader)and it is just a frontend for Aria2 and youtube-dl. It is a port from [ncdownloader for Nextcloud ](https://github.com/shiningw/ncdownloader). The current version has all the basic features and I can download torrents as well as youtube videos without issue.  

This PR adds only a few lines of code to FB and all it does is to provide an entry point to connect yadownloader
yadownloader-1  
yadownloader2


o1egl:
This might be useful for some people, but I don't think this is something we want to be merged with the main filebrowser repository. The way it looks and how it integrates with the app doesn't look native. We are trying to keep it as simple as possible.

xor-gate:
This is highly specific for a certain use case. Filebrowser has no plugin/extension/app model as far as I can tell and this looks like a prototype which I also think is not good for merging into the main repo.

shiningw:
Ok, I'm closing it. Thank you!

mckaygerhard:
this is a great feature.. the ui implementation can look ugly but that do not care cos provide great functional feature.. thanks @shiningw  i will implement in my fork rihgt now

Issue #2038, opened 2022-07-13
Avoid losing ongoing edits when navigating page by alerting user


Rearden21M:
**Is your feature request related to a problem? Please describe.**
In the current state of this program, it is way too easy to lose what you were working on. If you don't save your work and you accidentally close your window or go to the previous page, then your work will be completely lost with no warning whatsoever. 

**Describe the solution you'd like**
At the very least, there ought to be a message warning you that you have unsaved work and asking you if you are sure that you want to leave the page. This is the bare minimum. 

**Describe alternatives you've considered**
Another more advanced solution would be to offer a setting whereby your work is automatically saved, and you don't need to constantly worry about it.

Rmiller5466:
I assume this issue is specific to when a user is in the file editor?  If not, where else would this apply?  Also the editor already has a "Are you sure you wish to discard the changes you've made?" prompt, which I believe achieves what this issue is requesting?  Unless it is something else, this issue can probably be closed

hacdias:
@Rmiller5466 the prompt exists, indeed, but it isn't triggered when going back or forward in history, or refreshing the page. We should actually show it also when the user tries to leave the page in any way that could lead to content loss.

Rmiller5466:
Gotcha, I follow.  I'll look into it and see what I can do!

PR #2123, opened 2022-09-08
feat: add editor auto save

Lilihx:
Description

Add feature: Editor Auto Save.

1, Auto Save triggers for now:

click "close" button
auto save after interval, while the interval is set by user
2, Add editor settings card in Profile Settings, because I think there will be more configs about editor in the future.

https://github.com/filebrowser/filebrowser/issues/2038

Looking forward any better suggestions.


frontend/src/views/settings/Profile.vue
```diff
@@ -89,6 +114,7 @@ export default {
       singleClick: false,
       dateFormat: false,
       locale: "",
+      editorAutoSaveInterval: 60,
```
ramiresviana, line 117: Why 60 is being used as the default value? What's the default behavior for existing users?

PR #2145, opened 2022-09-25
Integrating tus.io for resumable and chunked uploads

TobiasGoerke:
First off: thank you for creating and maintaing this project.

**Description**

Currently, Filebrowser uploads files in one request. While this is inefficient in terms of upload speed it also breaks in some environments where large file uploads are not allowed (see [discussion here](https://github.com/filebrowser/filebrowser/discussions/1182)).

Chunked uploads were necessary in my case and I was able to integrate them into filebrowser using the suggested tus.io framework. As a sideeffect, this brings us resumable uploads. 
So, I thought I might as well contribute these changes here as other seem to require them likewise. 

I don't consider these changes ready to merge since I'm not too familiar with either the project, Go or Javascript. So, I'm looking for your support in case you like these changes. 

In case these changes won't get merged or you'd like to test my PR: you may build the filebrowser by using my fork and the following Dockerfile:

```
FROM alpine/git:latest as src

RUN git clone -b feature/tus-uploads https://github.com/TobiasGoerke/filebrowser.git /src

FROM node:16.20.0-alpine as nodejs-build

WORKDIR /frontend
COPY --from=src /src/frontend/package.json ./
RUN npm install
COPY --from=src /src/frontend ./
RUN npm run build


FROM golang:1.20.3 as builder
WORKDIR /src
COPY --from=src /src /src
RUN go mod download

COPY --from=nodejs-build /frontend ./frontend
RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build


FROM alpine:latest
RUN apk --update add ca-certificates \
                     mailcap \
                     curl

HEALTHCHECK --start-period=2s --interval=5s --timeout=3s \
  CMD curl -f http://localhost/health || exit 1

VOLUME /srv
EXPOSE 80

COPY --from=builder /src/docker_config.json /.filebrowser.json
COPY --from=builder /src/filebrowser /filebrowser

ENTRYPOINT [ "/filebrowser" ]
```

Thanks!


frontend/src/api/tus.js
```diff
@@ -0,0 +1,158 @@
...
+      httpStack: new CustomHTTPStack(),
+      endpoint: tusEndpoint,
+      chunkSize: chunkSize,
+      retryDelays: retryDelays,
+      parallelUploads: parallelUploads,
+      metadata: {
+        filename: content.name,
+        filetype: content.type,
+        overwrite: overwrite,
+        destination: removePrefix(url),
```
c0re100: I discovered `destination` haven't decode url.
If an user want to upload a file, maybe a file named `[ABC].txt` or something else.
It will become `%5BABC%5D.txt` after upload, not original filename `[ABC].txt`.

c0re100 (changes requested):
Thanks for your great work~

frontend/src/api/tus.js
TobiasGoerke: Thanks for reviewing and testing my PR. You are right, I'm not correctly decoding the URL encoded metadata values.
I've fixed this issue so that files with special characters can now be uploaded.

http/tus.go
```diff
@@ -0,0 +1,201 @@
...
+	uploadDir := filepath.Join(d.user.FullPath("/"), th.uploadDirName)
+	uploadedFile := filepath.Join(uploadDir, event.Upload.ID)
+	fullDestination := filepath.Join(d.user.FullPath("/"), destination)
+
+	log.Printf("Upload of %s (%s) is finished. Moving file to destination (%s) "+
+		"and cleaning up temporary files.\n", filename, uploadedFile, fullDestination)
+
+	// Check if destination file already exists. If so, we require overwrite to be set
+	if _, err := os.Stat(fullDestination); !errors.Is(err, os.ErrNotExist) {
+		if overwrite, err := strconv.ParseBool(overwrite); err != nil {
```
c0re100: Maybe you can naming it as `overwriteStr`.
Cause `if overwrite, err := strconv.ParseBool(overwrite); err != nil {` would make it confused.
So `overwriteStr` = it's a string type of `overwrite` and `overwrite` is bool.

just my two cents :)
FYI: https://github.com/moby/moby/blob/master/daemon/logger/splunk/splunk.go#L176

c0re100 (approved):

http/tus.go
TobiasGoerke: Fair point. Changed that as well.

CEbbinghaus:
What's the status? Anything holding this back? I have reverted to running without a cloudflare proxy but I am keen to be able to turn it back on

TobiasGoerke:
> What's the status? Anything holding this back? I have reverted to running without a cloudflare proxy but I am keen to be able to turn it back on

There's nothing holding this back from my side. I'm about to improve a detail in the implementation in the next few days though. 

Edit: I've refactored some parts, removed code and am confident to submit the PR to fhe filebrowser team. One could think about making some tus settings configurable. However, right now, things do work fine. 

@c0re100 how do we proceed with this PR?

Pikamander2:
We're running into this issue as well due the Cloudflare upload limit, which can't be raised without paying for Enterprise.

It would be fantastic if this software supported resumable, chunk-based uploads.

+1 for getting this merged soon, assuming that it's stable.

frontend/src/api/files.js
```diff
@@ -78,6 +79,22 @@ export function download(format, ...files) {
 }
 
 export async function post(url, content = "", overwrite = false, onupload) {
+  // Use the pre-existing API if:
+  const useResourcesApi =
```
CEbbinghaus, line 83: Do we want to add a setting to allow the admin to enable/disable this functionality?

frontend/src/api/tus.js
```diff
@@ -0,0 +1,50 @@
+import * as tus from "tus-js-client";
+import { tusEndpoint } from "@/utils/constants";
+import store from "@/store";
+import { removePrefix } from "./utils";
+
+// Make following configurable by envs?
```
CEbbinghaus: Yes. Maybe another section in the Global Settings tab?

http/tus.go
```diff
@@ -0,0 +1,242 @@
...
+
+func getBasePathFromRequest(r *http.Request, apiPath string) (*url.URL, error) {
+	// The tus protocol is designed to return a location header in its response, guiding the client to the correct endpoint.
+	// However, in proxied environments, we cannot know the correct URL at compile time
+	// We therefore make use of the Request-URI as sent by the client (https://www.w3.org/Protocols/rfc2616/rfc2616-sec5.html)
+	// Since we know this URI always contains the api path "/api/tus" we configured, we can form the basePath for tus accordingly
+	// In case this is a scheme-less URI, we prepend the origin header to form a full URL that has initially been requested
+	if idx := strings.Index(r.RequestURI, apiPath); idx < 0 {
+		return nil, fmt.Errorf("Expected URI to contain " + apiPath)
+	} else {
```
CEbbinghaus: Else is not needed and just adds needless Indenting

http/tus.go
```diff
@@ -0,0 +1,242 @@
...
+		if basePath, err := url.Parse(r.RequestURI[:(idx + len(apiPath))]); err != nil {
+			return nil, err
+		} else {
```
CEbbinghaus: Same as above

http/tus.go
```diff
@@ -0,0 +1,242 @@
...
+			if len(basePath.Scheme) != 0 {
+				return basePath, nil
+			}
+			if origin, ok := r.Header["Origin"]; !ok || len(origin) == 0 || len(origin[0]) == 0 {
+				return nil, fmt.Errorf("Client sent a request to a relative tus URL. " +
+					"Expected Origin header to be set in this case to form an absolute URL.")
+			} else {
```
CEbbinghaus: Same as above

http/tus.go
```diff
@@ -0,0 +1,242 @@
...
+		}
+	}
+}
+
+func (th tusHandler) getOrCreateTusHandler(d *data, r *http.Request) (*tusd.UnroutedHandler, error) {
+	if handler, ok := th.handlers[d.user.ID]; !ok {
+		log.Printf("Creating tus handler for user %s\n", d.user.Username)
+		if basePath, err := getBasePathFromRequest(r, th.apiPath); err != nil {
+			return nil, err
+		} else {
```
CEbbinghaus: Same as above

CEbbinghaus: In fact this whole function could be refactored to be cleaner

http/tus.go
```diff
@@ -0,0 +1,242 @@
...
+	// Move uploaded file from tmp upload folder to user folder
+	if err := os.Rename(uploadedFile, fullDestination); err != nil {
+		return err
+	}
+
+	// Remove uploaded tmp files for finished upload (.info objects are created and need to be removed, too))
+	for _, partialUpload := range append(event.Upload.PartialUploads, event.Upload.ID) {
+		if filesToDelete, err := filepath.Glob(filepath.Join(uploadDir, partialUpload+"*")); err != nil {
+			return err
+		} else {
```
CEbbinghaus: Again the else can be deleted

http/tus.go
```diff
@@ -0,0 +1,242 @@
...
+					return err
+				}
+			}
+		}
+	}
+
+	// Delete folder basePath if it is empty
+	if dir, err := ioutil.ReadDir(uploadDir); err != nil {
+		return err
+	} else if len(dir) == 0 {
```
CEbbinghaus: else is not necessary in this case. It doesn't do much because the if statement is still needed but it serves no functional purpose

frontend/src/api/files.js
CEbbinghaus: Should also consider using `tus.isSupported` to ensure that we don't try to use it when it cannot be used. Prompt the user with an error if the file is too big and they cannot use Tus?

frontend/src/api/tus.js
```diff
@@ -0,0 +1,50 @@
...
+        }
+      },
+      onSuccess: function () {
+        resolve();
+      },
+    });
+
+    upload.findPreviousUploads().then(function (previousUploads) {
+      if (previousUploads.length) {
+        upload.resumeFromPreviousUpload(previousUploads[0]);
```
CEbbinghaus: Will this only retry the very first upload? Since previousUploads is an array I would expect to have to iterate and retry them all. It looks like the intention is to prompt the user to select which upload they want to resume so we might need to add a popup to let the user manually resume uploads. Usage guide on this can be found [here](https://github.com/tus/tus-js-client/blob/master/docs/usage.md#example-let-user-select-upload-to-resume)

CEbbinghaus (changes requested):
Overall its a good start. With some settings added to allow configuration of the hard coded values and some code cleanliness it looks good. Ofc apart from the resume but I think the team will have to weigh in on that. All I can say is that how its written right now it doesn't stand.

http/tus.go
TobiasGoerke: Well, it's either using else or not being able to assign idx in the if. But I'll admit, assigning the var is nicer.

frontend/src/api/tus.js
TobiasGoerke: FindPreviousUploads retrieves a list of uploads that have been previously started for this file. According to the documentation, a fingerprint of the file is used for findPreviousUploads. The fingerprint doesn't include the file's hash (see [here](https://github.com/tus/tus-js-client/blob/aed3e8420e3105610bebdb6f6de3322495079484/lib/browser/fileSignature.js#L27)) 
So, which information would you like to show the user before letting them make a decision on which upload to resume? Why not simply select the most recent upload or the largest one?

frontend/src/api/files.js
TobiasGoerke: +1 for letting an admin or user disable/enable tus. Where would one best implement such a feature?

frontend/src/api/tus.js
TobiasGoerke: I'll see if I can add them there

TobiasGoerke:
> Overall its a good start. With some settings added to allow configuration of the hard coded values and some code cleanliness it looks good. Ofc apart from the resume but I think the team will have to weigh in on that. All I can say is that how its written right now it doesn't stand.

Thanks for taking the time to review my code. I've changed some minor requests and will look into implementing configurable settings somewhen over the next few weeks.
I'd be glad for support on things like user dialogs as I don't know filebrowser enough to do this efficiently. 
Also, given these things are implemented. who else needs to be in the loop and give feedback allowing this PR to move forward?
Thanks!

CEbbinghaus (commented):
Thanks for the changes, Looking a lot better

frontend/src/api/files.js
CEbbinghaus: https://github.com/filebrowser/filebrowser/blob/master/frontend/src/views/settings/Global.vue 
is the frontend view which just sends its settings as JSON to the API
and the following file contains the settings definitions in GO
https://github.com/filebrowser/filebrowser/blob/master/settings/settings.go

TobiasGoerke:
Is there any way we can move forward with this PR? [We're currently considering using the Filebrowser for the Machine Learning Platform Kubeflow](https://github.com/kubeflow/kubeflow/pull/6876). This use-case would require users uploading large masses of data efficiently, hence tus.io. Thanks!

arter97:
Haven't tried out @TobiasGoerke's fork yet myself, but tus.io is by far the most reliable method to upload big data.

This will make filebrowser a full-fledged web-based file explorer comparable to SMB, and we'd definitely appreciate merging this and improving it further down the road (if there are any room left).

Thanks!
(And thanks @TobiasGoerke for working on the code!)

Pikamander2:
Just to confirm interest, we're still awaiting something like Tobias's solution to be able to manually upload large site backups while our site is connected to Cloudflare. If the change is able to be rolled out in v2.24.0 then that would help us a lot.

frontend/src/api/tus.js
```diff
@@ -48,3 +46,40 @@ export async function upload(url, content = "", overwrite = false, onupload) {
...
+
+function computeRetryDelays(tusSettings) {
+  if (!tusSettings.retryCount || tusSettings.retryCount < 1) {
+    // Disable retries altogether
+    return null;
+  }
+  return Array.apply(null, { length: tusSettings.retryCount }).map(
+    Number.call,
+    (idx) => (idx + 1) * tusSettings.retryBaseDelay * tusSettings.retryBackoff
+  );
```
CEbbinghaus: This section should be simplified to reduce confusion. Also a comment to explain what its doing would be helpful.
```suggestion
  return Array.apply(null, { length: tusSettings.retryCount }).map(
    (_, idx) => (idx + 1) * tusSettings.retryBaseDelay * tusSettings.retryBackoff
  );
```

frontend/src/api/tus.js
```diff
@@ -48,3 +46,40 @@ export async function upload(url, content = "", overwrite = false, onupload) {
...
+}
+
+export async function useTus(content) {
+  if (!(isTusSupported() || content instanceof Blob)) {
```
CEbbinghaus: Why are we returning false if the content is a Blob?

frontend/src/views/settings/Global.vue
```diff
@@ -92,6 +92,91 @@
...
+          <p>
+            <input
+              type="checkbox"
+              v-model="settings.tus.enabled"
+              id="tus-enabled"
+            />
+            {{ $t("settings.tusUploadsEnabled") }}
+          </p>
+
+          <p v-if="settings.tus.enabled">
```
CEbbinghaus: will these if statements cause Vue to hide the whole setting if its not enabled? If so that might cause a large content shift when pressing the button & lead to a worse UX. We could create one wrapper div that has an optional class that marks all other settings as disabled (for styling) & just define the disabled property on each of the inputs to reduce any layout shifting but still convey that they are no longer active since Tus is disabled.

CEbbinghaus (commented):
Looks pretty good. The settings are going to go a long way to make this a lot more usable. Do you have a screenshot of what they look like in the current UI? it would be great to see them in context.

Other than the comments I think this is pretty close to being ready (after a team member has reviewed it)

TobiasGoerke:
> Looks pretty good. The settings are going to go a long way to make this a lot more usable. Do you have a screenshot of what they look like in the current UI? it would be great to see them in context.
> 
> Other than the comments I think this is pretty close to being ready (after a team member has reviewed it)

Thank you for your repeated review and feedback. 

The settings previously looked like this:
grafik

Disabling the "Enable Tus" checkbox led to the whole block disappearing. As you suggested, I've now changed this to disable the fields only, leading to a better UX:
grafik

One thing left from my side: 
The frontend needs to fetch the tus settings before uploading files. It stores them temporarily on client-side to not have an additional query for each uploaded file. Thus, changes to the settings are only propagated, once the user refreshed the page. Is this an issue?

TobiasGoerke:
> Looks pretty good. The settings are going to go a long way to make this a lot more usable. Do you have a screenshot of what they look like in the current UI? it would be great to see them in context.
> 
> Other than the comments I think this is pretty close to being ready (after a team member has reviewed it)

As this PR is staling again: is there any team member we can ping or ask for a review?

CEbbinghaus:
Whoops sorry. Seemed to have missed the reply to my comment (my bad). The UI changes look fantastic & will go a long way to making this more usable. I'll give it a review but I think @o1egl might be the person to ping.

CEbbinghaus (approved):
Looks great. 👍🏻 from my side.

CEbbinghaus (approved):

mcorrigan:
This is exactly what we've been looking for! +1 to having this go into the next release.

j4pa0:
How's thing going for this pull? Would like to have this in the next release 😄.

CEbbinghaus:
@j4pa0 The PR requires a contributor to approve it & clear it to run unit tests. so at this point we are waiting on @o1egl or any of the other contributors

TobiasGoerke:
Pinging @o1egl: could we please get your review/call on this PR? Thanks!

frontend/src/api/tus.js
```diff
@@ -0,0 +1,86 @@
...
+
+// Temporarily store the tus settings stored in the backend
+// Thus, we won't need to fetch the settings every time we upload a file
+var temporaryTusSettings = null;
+
+async function getTusSettings() {
+  if (temporaryTusSettings) {
+    return temporaryTusSettings;
+  }
+  const fbSettings = await settings.get();
```
c0re100: I discovered 1 upload bug.

That line is affecting normal user, it leads to 403 forbidden for upload file.
Cause user can't access `/api/settings`.
To access this API, you must have admin permission.

For example, a user do not have admin permission but user can create(upload) files.
{DDDFD64E-8CB2-4444-97CA-2F9733D3589A}

Without admin permission, then user can't upload anything to server.

Just my two cents:
Expose one more API like `/api/tus_settings` to access tus settings.

c0re100 (changes requested):

frontend/src/api/tus.js
CEbbinghaus: Good catch. Thanks for that ❤️

CEbbinghaus (approved):
lgtm
Thanks for the quick change. Now all we need is @o1egl to approve & merge

Emojigit:
I've tested it and it works perfectly. Wait for approval.

arter97:
Bump....

arter97:
So I've decided to try it out by myself.

The node and golang versions were a bit out of date from @TobiasGoerke's original Dockerfile.
The Nodejs v16's latest LTS version - 16.20.0 and golang's latest version - 1.20.3 seems to work just fine to build filebrowser.

Hope this helps for someone else too.

```
FROM node:16.20.0-alpine as nodejs-build
FROM golang:1.20.3 as builder
```

TobiasGoerke:
> So I've decided to try it out by myself.
> 
> The node and golang versions were a bit out of date from @TobiasGoerke's original Dockerfile. The Nodejs v16's latest LTS version - 16.20.0 and golang's latest version - 1.20.3 seems to work just fine to build filebrowser.
> 
> Hope this helps for someone else too.
> 
> ```
> FROM node:16.20.0-alpine as nodejs-build
> FROM golang:1.20.3 as builder
> ```

Thanks, updated the description. 
I haven't seen a maintainer's reaction since I've opened this PR seven months ago. I'm afraid building the image yourself will sadly remain the way to go here..

Is anyone aware of the project's state? Is filebrowser dead?

arter97:
Well, things happen, especially on FOSS projects with no clear fundings from big companies :(

It's a bit worrying that a manual `npm install` spits out a bunch of security vulnerability warnings, especially considering that this is a file-hosting web server.

We're currently evaluating how to securely sandbox filebrowser so that it cannot access the system even when it's exploited.
Releases are made with static binaries, so we are thinking of boxing it under chroot as a non-root user.

arter97:
@TobiasGoerke It's only been a few minutes playing around with your fork (which, I thank you again), but it seems that the current UI lacks pause (or cancel) and a speed indicator.

We were thinking of deploying this to upload 10-30 GiB of files regularly, so a pause button and a speed indicator would be a nice touch.

Would this be hard to implement?

TobiasGoerke:
> Well, things happen, especially on FOSS projects with no clear fundings from big companies :(
> 
> It's a bit worrying that a manual `npm install` spits out a bunch of security vulnerability warnings, especially considering that this is a file-hosting web server.
> 
> We're currently evaluating how to securely sandbox filebrowser so that it cannot access the system even when it's exploited. Releases are made with static binaries, so we are thinking of boxing it under chroot as a non-root user.

Executing the container as a non-root user work fine without any modifications to the Dockerfile. We'ven been executing our filebrowsers that way for months.

TobiasGoerke:
> @TobiasGoerke It's only been a few minutes playing around with your fork (which, I thank you again), but it seems that the current UI lacks pause (or cancel) and a speed indicator.
> 
> We were thinking of deploying this to upload 10-30 GiB of files regularly, so a pause button and a speed indicator would be a nice touch.
> 
> Would this be hard to implement?

Are you sure the speed indicator doesn't work for you? They do for me. 
I don't think filebrowser currently offers an option to pause uploads but it should be doable. However, I'd see it as out of scope for this PR and might join in on efforts implementing this after this PR has been merged

arter97:
> > @TobiasGoerke It's only been a few minutes playing around with your fork (which, I thank you again), but it seems that the current UI lacks pause (or cancel) and a speed indicator.
> > We were thinking of deploying this to upload 10-30 GiB of files regularly, so a pause button and a speed indicator would be a nice touch.
> > Would this be hard to implement?
> 
> Are you sure the speed indicator doesn't work for you? They do for me. I don't think filebrowser currently offers an option to pause uploads but it should be doable. However, I'd see it as out of scope for this PR and might join in on efforts implementing this after this PR has been merged

Hm, this is what I get:
image

No percentage, no speed indicator. Same thing when I scroll to the right. Is this something wrong on my end?

I've never used filebrowser's upload functionality before (because it was so limited), so maybe I'm using it wrong.

TobiasGoerke:
> > > @TobiasGoerke It's only been a few minutes playing around with your fork (which, I thank you again), but it seems that the current UI lacks pause (or cancel) and a speed indicator.
> > > We were thinking of deploying this to upload 10-30 GiB of files regularly, so a pause button and a speed indicator would be a nice touch.
> > > Would this be hard to implement?
> > 
> > 
> > Are you sure the speed indicator doesn't work for you? They do for me. I don't think filebrowser currently offers an option to pause uploads but it should be doable. However, I'd see it as out of scope for this PR and might join in on efforts implementing this after this PR has been merged
> 
> Hm, this is what I get: image
> 
> No percentage, no speed indicator. Same thing when I scroll to the right. Is this something wrong on my end?
> 
> I've never used filebrowser's upload functionality before (because it was so limited), so maybe I'm using it wrong.

This is how it looks for me:
grafik

Are you sure your file upload is actually progressing? Have you checked the networking tools?

arter97:
> > > > @TobiasGoerke It's only been a few minutes playing around with your fork (which, I thank you again), but it seems that the current UI lacks pause (or cancel) and a speed indicator.
> > > > We were thinking of deploying this to upload 10-30 GiB of files regularly, so a pause button and a speed indicator would be a nice touch.
> > > > Would this be hard to implement?
> > > 
> > > 
> > > Are you sure the speed indicator doesn't work for you? They do for me. I don't think filebrowser currently offers an option to pause uploads but it should be doable. However, I'd see it as out of scope for this PR and might join in on efforts implementing this after this PR has been merged
> > 
> > 
> > Hm, this is what I get: image
> > No percentage, no speed indicator. Same thing when I scroll to the right. Is this something wrong on my end?
> > I've never used filebrowser's upload functionality before (because it was so limited), so maybe I'm using it wrong.
> 
> This is how it looks for me: grafik
> 
> Are you sure your file upload is actually progressing? Have you checked the networking tools?

Oh yeah, the speed is awesome. It's saturating my 500 Mbps link over the Internet.

By speed indicator, I meant MiB/s. I'm not seeing it on your screenshot either. Are we talking about the same thing? 🙃

TobiasGoerke:
> > > > > @TobiasGoerke It's only been a few minutes playing around with your fork (which, I thank you again), but it seems that the current UI lacks pause (or cancel) and a speed indicator.
> > > > > We were thinking of deploying this to upload 10-30 GiB of files regularly, so a pause button and a speed indicator would be a nice touch.
> > > > > Would this be hard to implement?
> > > > 
> > > > 
> > > > Are you sure the speed indicator doesn't work for you? They do for me. I don't think filebrowser currently offers an option to pause uploads but it should be doable. However, I'd see it as out of scope for this PR and might join in on efforts implementing this after this PR has been merged
> > > 
> > > 
> > > Hm, this is what I get: image
> > > No percentage, no speed indicator. Same thing when I scroll to the right. Is this something wrong on my end?
> > > I've never used filebrowser's upload functionality before (because it was so limited), so maybe I'm using it wrong.
> > 
> > 
> > This is how it looks for me: grafik
> > Are you sure your file upload is actually progressing? Have you checked the networking tools?
> 
> Oh yeah, the speed is awesome. It's saturating my 500 Mbps link over the Internet.
> 
> By speed indicator, I meant MiB/s. I'm not seeing it on your screenshot either. Are we talking about the same thing? 🙃

Ah, sorry. I see: speed, not progress indicator. 
No, there's currently no speed indicator implemented, only a progress bar which you could visually read the percentage off. Implementing a speed indicator should also be doable. Why don't you go ahead and open up a feature request issue with both the speed indicator and the pause button so we can track this separately? Thanks 😀

arter97:
@TobiasGoerke Maybe 2 separate issue may be an overkill, but as it's a 2 separate functionality, I hope the maintainers will be ok with it.

#2415 #2416

I wish I knew enough of Vue and Go to help implementing these :(

Anyways, this PR seems good enough for us to use as-is. My org will have a brief period of testing and push to production.

Thanks again, @TobiasGoerke :)

arter97:
Bump.

Pikamander2:
We can also confirm that we're still in need of this.

CEbbinghaus:
Ok this is getting a bit ridiculous. @o1egl reviewed #2403 just 3 weeks ago but this PR hasn't even gotten a comment. What does it take to get this approved by one of the maintainers?
image

http/settings.go
```diff
@@ -27,13 +28,21 @@ var settingsGetHandler = withAdmin(func(w http.ResponseWriter, r *http.Request,
...
 
 	return renderJSON(w, r, data)
 })
 
+// Tus settings are queried by the frontend before uploading files using the tus protocol.
+// While these settings are included with the getSettings handler, it requires admin privileges.
+// However, all users should be able to query these settings in order to upload data.
+var tusSettingsGetHandler = withUser(func(w http.ResponseWriter, r *http.Request, d *data) (int, error) {
+	return renderJSON(w, r, d.settings.Tus)
+})
```
o1egl: How about injecting them into the index.html? The same way we handle other settings?

http/tus.go
```diff
@@ -0,0 +1,247 @@
...
+	if !ok || len(origin) == 0 || len(origin[0]) == 0 {
+		return nil, fmt.Errorf("Client sent a request to a relative tus URL. " +
+			"Expected Origin header to be set in this case to form an absolute URL.")
+	}
+	parsedOrigin, err := url.Parse(origin[0])
+	if err != nil {
+		return nil, err
+	}
+	return parsedOrigin.ResolveReference(basePath), nil
+}
```
o1egl: Why do we need all of those? We know the baseurl provided on startup

http/tus.go
```diff
@@ -0,0 +1,247 @@
...
+type tusHandler struct {
+	store         *storage.Storage
+	server        *settings.Server
+	settings      *settings.Settings
+	uploadDirName string
+	handlers      map[uint]*tusd.UnroutedHandler
+	apiPath       string
+}
+
+var mutex sync.Mutex
```
o1egl: Don't use global variables

http/tus.go
```diff
@@ -0,0 +1,247 @@
...
+
+	return handler, nil
+}
+
+func (th tusHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+
+	code, err := withUser(func(w http.ResponseWriter, r *http.Request, d *data) (int, error) {
+		// Create a new tus handler for current user if it doesn't exist yet
+		// Use a mutex to make sure only one tus handler is created for each user
+		mutex.Lock()
```
o1egl: I think it should be used inside the method that modifies the data

http/tus.go
```diff
@@ -0,0 +1,247 @@
...
+		store:    th.store,
+		settings: th.settings,
+		server:   th.server,
+	})
+
+	if err != nil {
+		http.Error(w, err.Error(), code)
+	} else if code >= 400 {
+		http.Error(w, "", code)
+	}
```
o1egl:
```go
switch {
    case err != nil:
        http.Error(w, err.Error(), code)
    case code >= 400:
        http.Error(w, "", code)
}
```

http/tus.go
```diff
@@ -0,0 +1,247 @@
...
+	composer := tusd.NewStoreComposer()
+	tusStore.UseIn(composer)
+
+	handler, err := tusd.NewUnroutedHandler(tusd.Config{
+		BasePath:              basePath,
+		StoreComposer:         composer,
+		NotifyCompleteUploads: true,
+	})
+	if err != nil {
+		panic(fmt.Errorf("Unable to create handler: %s", err))
```
o1egl: Why not return this error?

http/tus.go
```diff
@@ -0,0 +1,247 @@
...
+
+	// Goroutine to handle completed uploads
+	go func() {
+		for {
+			event := <-handler.CompleteUploads
+
+			if err := th.handleTusFileUploaded(handler, d, event); err != nil {
+				log.Printf("ERROR: couldn't handle completed upload: %s\n", err)
+			}
+		}
```
o1egl:
```suggestion
		for event := range handler.CompleteUploads {
			if err := th.handleTusFileUploaded(handler, d, event); err != nil {
				log.Printf("ERROR: couldn't handle completed upload: %s\n", err)
			}
		}
```

http/tus.go
```diff
@@ -0,0 +1,247 @@
...
+		BasePath:              basePath,
+		StoreComposer:         composer,
+		NotifyCompleteUploads: true,
+	})
+	if err != nil {
+		panic(fmt.Errorf("Unable to create handler: %s", err))
+	}
+
+	// Goroutine to handle completed uploads
+	go func() {
```
o1egl: What if there are 10k users on the platform? The app will spawn 10k permanent goroutines doing nothing but consuming resources

http/tus.go
```diff
@@ -0,0 +1,247 @@
...
+	return handler
+}
+
+func readMetadata(metadata tusd.MetaData, field string) (string, error) {
+	if value, ok := metadata[field]; ok {
+		return value, nil
+	} else {
+		return "", errors.New(fmt.Sprintf("Metadata field %s not found in upload request", field))
+	}
+}
```
o1egl:
```suggestion
func getMetadataField(metadata tusd.MetaData, field string) (string, error) {
	if value, ok := metadata[field]; ok {
		return value, nil
	} else {
		return "", fmt.Errorf("metadata field %s not found in upload request", field)
	}
}
```

http/tus.go
```diff
@@ -0,0 +1,247 @@
...
+
+	log.Printf("Upload of %s (%s) is finished. Moving file to destination (%s) "+
+		"and cleaning up temporary files.\n", filename, uploadedFile, fullDestination)
+
+	// Check if destination file already exists. If so, we require overwrite to be set
+	if _, err := os.Stat(fullDestination); !errors.Is(err, os.ErrNotExist) {
+		if overwrite, err := strconv.ParseBool(overwriteStr); err != nil {
+			return err
+		} else if !overwrite {
+			return fmt.Errorf("Overwrite is set to false while destination file %s exists. Skipping upload.\n", destination)
```
o1egl:
```suggestion
			return fmt.Errorf("overwrite is set to false while destination file %s exists", destination)
```

Pikamander2:
Thanks for the feedback @o1egl . Hopefully @TobiasGoerke can take a look at it soon and review/correct it.

If there's going to be a release soon, would it be possible to merge the existing code and refine it later? This is a really important feature that's needed for Cloudflare compatibility and I would hate to see it miss this release cycle.

TobiasGoerke:
The feedback was very valuable, thanks, @o1egl! 
I've resolved all of your notes and suggestions, except one that I commented on. 
What do you think?

CEbbinghaus (commented):
Thanks for the better error handling 👍🏻

CEbbinghaus (approved):
Thanks for the better error handling 👍🏻

CEbbinghaus (approved):
Thanks for the better error handling 👍🏻

CEbbinghaus:
Thank you @o1egl for giving this PR a review. Hopefully it can be merged as soon as all comments are resolved :3

TobiasGoerke:
I've just managed to also resolve your last comment - tusd now uses the baseUrl provided through filebrowser.
I'd be thankful if you gave this another review @o1egl

CEbbinghaus (approved):
Looks cleaner. Nice job refactoring

TobiasGoerke:
Any chance for a new review, @o1egl?

Kirari04:
> First off: thank you for creating and maintaing this project.
> 
> **Description**
> 
> Currently, Filebrowser uploads files in one request. While this is inefficient in terms of upload speed it also breaks in some environments where large file uploads are not allowed (see [discussion here](https://github.com/filebrowser/filebrowser/discussions/1182)).
> 
> Chunked uploads were necessary in my case and I was able to integrate them into filebrowser using the suggested tus.io framework. As a sideeffect, this brings us resumable uploads. So, I thought I might as well contribute these changes here as other seem to require them likewise.
> 
> I don't consider these changes ready to merge since I'm not too familiar with either the project, Go or Javascript. So, I'm looking for your support in case you like these changes.
> 
> In case these changes won't get merged or you'd like to test my PR: you may build the filebrowser by using my fork and the following Dockerfile:
> 
> ```
> FROM alpine/git:latest as src
> 
> RUN git clone -b feature/tus-uploads https://github.com/TobiasGoerke/filebrowser.git /src
> 
> FROM node:16.20.0-alpine as nodejs-build
> 
> WORKDIR /frontend
> COPY --from=src /src/frontend/package.json ./
> RUN npm install
> COPY --from=src /src/frontend ./
> RUN npm run build
> 
> 
> FROM golang:1.20.3 as builder
> WORKDIR /src
> COPY --from=src /src /src
> RUN go mod download
> 
> COPY --from=nodejs-build /frontend ./frontend
> RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build
> 
> 
> FROM alpine:latest
> RUN apk --update add ca-certificates \
>                      mailcap \
>                      curl
> 
> HEALTHCHECK --start-period=2s --interval=5s --timeout=3s \
>   CMD curl -f http://localhost/health || exit 1
> 
> VOLUME /srv
> EXPOSE 80
> 
> COPY --from=builder /src/docker_config.json /.filebrowser.json
> COPY --from=builder /src/filebrowser /filebrowser
> 
> ENTRYPOINT [ "/filebrowser" ]
> ```
> 
> Thanks!


Hi @TobiasGoerke
I built the docker image locally, but when running the container and trying to upload I get the following error message:
```
Upload failed: Error: tus: unexpected response while uploading chunk, originated from request (method: PATCH, url: https://media.senpai.one//api/tus/9418294d1f06c4ff0df6195392b68185, response code: 404, response text: 404 Not Found , request id: n/a)
```

Logs:
```
2023/06/26 12:25:53 //api/tus/0304ee67e01112b3ea8a21a6d6ce1dd7: 404 194.209.11.14 <nil>

2023/06/26 12:25:53 //api/tus/2b6ac98494a4b21f9484602afc8da56c: 404 194.209.11.14 <nil>

2023/06/26 12:25:54 Creating tus handler for user admin on path https://media.senpai.one//api/tus

2023/06/26 12:25:54 http: panic serving 172.17.0.1:40010: runtime error: invalid memory address or nil pointer dereference

goroutine 39 [running]:

net/http.(*conn).serve.func1()

	/usr/local/go/src/net/http/server.go:1854 +0xb0

panic({0x6435a0, 0x1533190})

	/usr/local/go/src/runtime/panic.go:890 +0x248

github.com/tus/tusd/pkg/handler.(*UnroutedHandler).PostFile(0x0, {0x111ead0, 0x40000f2700}, 0x400024c000)

	/go/pkg/mod/github.com/tus/tusd@v1.11.0/pkg/handler/unrouted_handler.go:303 +0x88

github.com/filebrowser/filebrowser/v2/http.tusHandler.ServeHTTP.func1({0x111ead0, 0x40000f2700}, 0x400024c000, 0x40000b1740)

	/src/http/tus.go:108 +0x13c

github.com/filebrowser/filebrowser/v2/http.withUser.func1({0x111ead0, 0x40000f2700}, 0x400008f7f8?, 0x40000b1740)

	/src/http/auth.go:90 +0x354

github.com/filebrowser/filebrowser/v2/http.tusHandler.ServeHTTP({0x4000127950, 0x4000152b40, 0x4000128dc0, {0x6f49d4, 0xb}, 0x4000200d50, {0x6f2db8, 0x8}, 0x40002049e0, 0x4000094240}, ...)

	/src/http/tus.go:119 +0x13c

github.com/filebrowser/filebrowser/v2/http.NewHandler.func1.1({0x111ead0, 0x40000f2700}, 0x40000c0d20?)

	/src/http/http.go:31 +0xe4

net/http.HandlerFunc.ServeHTTP(0x40000bdf00?, {0x111ead0?, 0x40000f2700?}, 0x0?)

	/usr/local/go/src/net/http/server.go:2122 +0x38

github.com/gorilla/mux.(*Router).ServeHTTP(0x40001226c0, {0x111ead0, 0x40000f2700}, 0x40000bde00)

	/go/pkg/mod/github.com/gorilla/mux@v1.8.0/mux.go:210 +0x19c

net/http.serverHandler.ServeHTTP({0x4000195410?}, {0x111ead0, 0x40000f2700}, 0x40000bde00)

	/usr/local/go/src/net/http/server.go:2936 +0x2c0

net/http.(*conn).serve(0x40001a73b0, {0x111f288, 0x4000224510})

	/usr/local/go/src/net/http/server.go:1995 +0x518

created by net/http.(*Server).Serve

	/usr/local/go/src/net/http/server.go:3089 +0x4e8

[tusd] 2023/06/26 12:25:54.085749 event="UploadCreated" id="3ee20e40687868fac9febcf4c13d5fff" size="482103921" url="https://media.senpai.one//api/tus/3ee20e40687868fac9febcf4c13d5fff" 

[tusd] 2023/06/26 12:25:54.085830 event="ResponseOutgoing" status="201" method="POST" path="/api/tus" requestId="" 

2023/06/26 12:25:54 //api/tus/f95afe57a0ee323cf543b7eeb7b6582e: 404 194.209.11.14 <nil>

[tusd] 2023/06/26 12:25:54.382131 event="UploadCreated" id="1318a8ac2321dac897a5e18994b89a8c" size="482103923" url="https://media.senpai.one//api/tus/1318a8ac2321dac897a5e18994b89a8c" 

[tusd] 2023/06/26 12:25:54.382203 event="ResponseOutgoing" status="201" method="POST" path="/api/tus" requestId="" 

2023/06/26 12:25:55 //api/tus/3ee20e40687868fac9febcf4c13d5fff: 404 194.209.11.14 <nil>

2023/06/26 12:25:56 //api/tus/1318a8ac2321dac897a5e18994b89a8c: 404 194.209.11.14 <nil>

2023/06/26 12:25:56 //api/tus/2b6ac98494a4b21f9484602afc8da56c: 404 194.209.11.14 <nil>

[tusd] 2023/06/26 12:25:56.398600 event="UploadCreated" id="a34a7cba2e5e30c3786cdfdc6087474d" size="482103921" url="https://media.senpai.one//api/tus/a34a7cba2e5e30c3786cdfdc6087474d" 

[tusd] 2023/06/26 12:25:56.398676 event="ResponseOutgoing" status="201" method="POST" path="/api/tus" requestId="" 

2023/06/26 12:25:57 //api/tus/a34a7cba2e5e30c3786cdfdc6087474d: 404 194.209.11.14 <nil>

2023/06/26 12:26:03 //api/tus/3ee20e40687868fac9febcf4c13d5fff: 404 194.209.11.14 <nil>

[tusd] 2023/06/26 12:26:04.008473 event="UploadCreated" id="b42b8649b49520e0cd1fa7ad926b636a" size="482103921" url="https://media.senpai.one//api/tus/b42b8649b49520e0cd1fa7ad926b636a" 

[tusd] 2023/06/26 12:26:04.008606 event="ResponseOutgoing" status="201" method="POST" path="/api/tus" requestId="" 

2023/06/26 12:26:05 //api/tus/b42b8649b49520e0cd1fa7ad926b636a: 404 194.209.11.14 <nil>
```

Dockerfile:
```
FROM alpine/git:latest as src

RUN git clone -b feature/tus-uploads https://github.com/TobiasGoerke/filebrowser.git /src

FROM node:16.20.0-alpine as nodejs-build

WORKDIR /frontend
COPY --from=src /src/frontend/package.json ./
RUN npm install
COPY --from=src /src/frontend ./
RUN npm run build


FROM golang:1.20.3 as builder
WORKDIR /src
COPY --from=src /src /src
RUN go mod download

COPY --from=nodejs-build /frontend ./frontend
RUN CGO_ENABLED=0 GOOS=linux GOARCH=arm64 go build


FROM alpine:latest
RUN apk --update add ca-certificates \
                     mailcap \
                     curl

HEALTHCHECK --start-period=2s --interval=5s --timeout=3s \
  CMD curl -f http://localhost/health || exit 1

VOLUME /srv
EXPOSE 80

COPY --from=builder /src/docker_config.json /.filebrowser.json
COPY --from=builder /src/filebrowser /filebrowser

ENTRYPOINT [ "/filebrowser" ]
```

Buld command:
```
docker build . --platform linux/arm64 -f Dockerfile -t kirari04/filebrowser:arm64 --push
```

Published & Pulled from here: `https://hub.docker.com/r/kirari04/filebrowser/tags`
I just tweaked the dockerfile a little so it can be run on an arm based server

TobiasGoerke:
> https://media.senpai.one//api/tus/9418294d1f06c4ff0df6195392b68185

Based on your logs there's some strangely formatted URLs:
* `Creating tus handler for user admin on path https://media.senpai.one//api/tus`
* `//api/tus/b42b8649b49520e0cd1fa7ad926b636a`
* ...

In all cases, there only should be a single slash. 
Thus, my guess is that the [`basePath` is not being computed](https://github.com/filebrowser/filebrowser/blob/5238bee941481e14b4223615efdc1fcfbbb3ccf1/http/tus.go#L77) correctly.

This happens, when no `baseUrl` is set. I've pushed a quick fix to [this branch](https://github.com/TobiasGoerke/filebrowser/tree/tus-uploads-base-path-fix). Does this work for you?

Kirari04:
Thx @TobiasGoerke 
now it works!
You might want to include this in the PR so that it works with no configuration out of the box.

TobiasGoerke:
> Thx @TobiasGoerke now it works! You might want to include this in the PR so that it works with no configuration out of the box.

First wanted to make sure the changes work for you. It's merged now and also works for me.

Thanks for testing and pointing this out. Much appreciated 🥇 

Please let me know in case you encounter any more issues with this feature.

Kirari04:
Hi @TobiasGoerke 
after I put the server behind the cloudflare proxy and tried to upload there seems to be an issue where the upload request would be with http instead if https

grafik
grafik

TobiasGoerke:
> Hi @TobiasGoerke after I put the server behind the cloudflare proxy and tried to upload there seems to be an issue where the upload request would be with http instead if https
> 
> grafik grafik

I've experienced this, too and it's most probably an issue with your session. Please clear your local storage and try again. Tus saves previously uploaded chunk addresses. So when you switch from http to https on the same host, this can cause the issue

Kirari04:
@TobiasGoerke  suggested solution worked only sometimes so I looked up for an other solution.
Inside the tus gh issues I found a recommendation to put the server behind the nginx proxy using the following config:
https://github.com/tus/tusd/blob/main/examples/nginx.conf

After deleting the local storage and other session related stuff inside the browser, then restarting the filebrowser container, it finally resolved my issue.

Related issue:
"tus client always sending request to http instead of https"
https://github.com/tus/tus-js-client/issues/58

CEbbinghaus:
Honestly this is a great point. I will build the image myself and use that as waiting on this functionality to be merged to the upstream seems pointless at this point :\

TobiasGoerke:
> @TobiasGoerke suggested solution worked only sometimes so I looked up for an other solution. Inside the tus gh issues I found a recommendation to put the server behind the nginx proxy using the following config: https://github.com/tus/tusd/blob/main/examples/nginx.conf
> 
> After deleting the local storage and other session related stuff inside the browser, then restarting the filebrowser container, it finally resolved my issue.
> 
> Related issue: "tus client always sending request to http instead of https" [tus/tus-js-client#58](https://github.com/tus/tus-js-client/issues/58)

Great to hear! When using HTTPS from the first second (as should be in production environments), the problem won't even appear. Also, as far as I can tell, the issue doesn't cause uploads to abort, just throws some network errors invisible to the user.

Unfortunately, I don't think adding nginx to the project is an option. 


> Honestly this is a great point. I will build the image myself and use that as waiting on this functionality to be merged to the upstream seems pointless at this point :\

I'm afraid you're right. I've already been thinking about closing this PR and creating a fork instead..

CEbbinghaus:
> I'm afraid you're right. I've already been thinking about closing this PR and creating a fork instead..

As this project is under a organization with the same name we should offer to help maintain the project rather than fork it. @o1egl I guess this is our call out to you. We would love to see this project succeed & flourish but at the moment it seems stagnant. I assume @TobiasGoerke would be happy to help maintain it & while I have yet to make a contribution beyond reviewing the code I am willing to continue to do exactly that.

@o1egl I understand that maintaining FOSS software is difficult and time consuming especially when its being maintained during non work hours. But equally we are offering help to maintain this project and would be willing to fork it just to keep contributions flowing.

Pikamander2:
Seconding the above; I've seen far too many FOSS projects die because they stop receiving updates. It would be great if Filebrowser could officially add another person to the project to help push out regular updates and ensure the project's longevity.

nicoderoin:
While we're waiting for this to be reviewed / officially implemented, has anyone had any luck deploying this onto TrueNAS Scale? This might not be the right place to ask, but I can't for the life of me how to deploy this Docker image as a TrueNAS App (I've never deployed a custom image like this to TrueNAS before) - until this feature gets merged into the main code + deployed to TrueCharts, any guidance/recommendations would be appreciated

TobiasGoerke:
> While we're waiting for this to be reviewed / officially implemented, has anyone had any luck deploying this onto TrueNAS Scale? This might not be the right place to ask, but I can't for the life of me how to deploy this Docker image as a TrueNAS App (I've never deployed a custom image like this to TrueNAS before) - until this feature gets merged into the main code + deployed to TrueCharts, any guidance/recommendations would be appreciated

I don't know about TrueNAS but may be able to help.
Are you seeing an error message? What architecture is TrueNAS running on? x64? arm64? Have you build the image yourself and on your device?

nicoderoin:
> 



> > While we're waiting for this to be reviewed / officially implemented, has anyone had any luck deploying this onto TrueNAS Scale? This might not be the right place to ask, but I can't for the life of me how to deploy this Docker image as a TrueNAS App (I've never deployed a custom image like this to TrueNAS before) - until this feature gets merged into the main code + deployed to TrueCharts, any guidance/recommendations would be appreciated
> 
> I don't know about TrueNAS but may be able to help. Are you seeing an error message? What architecture is TrueNAS running on? x64? arm64? Have you build the image yourself and on your device?

Hey, thanks for the quick response!

tldr; I got it working this morning

I don't know if it was my exhaustion-addled brain or a simple setting I missed in my initial config, but it's running brilliantly now. Huge appreciation to you for working on adding this feature - definitely beats having to pay $20+/month for the same amount of storage from a third party service c:

TobiasGoerke:
> > > While we're waiting for this to be reviewed / officially implemented, has anyone had any luck deploying this onto TrueNAS Scale? This might not be the right place to ask, but I can't for the life of me how to deploy this Docker image as a TrueNAS App (I've never deployed a custom image like this to TrueNAS before) - until this feature gets merged into the main code + deployed to TrueCharts, any guidance/recommendations would be appreciated
> > 
> > 
> > I don't know about TrueNAS but may be able to help. Are you seeing an error message? What architecture is TrueNAS running on? x64? arm64? Have you build the image yourself and on your device?
> 
> Hey, thanks for the quick response!
> 
> tldr; I got it working this morning
> 
> I don't know if it was my exhaustion-addled brain or a simple setting I missed in my initial config, but it's running brilliantly now. Huge appreciation to you for working on adding this feature - definitely beats having to pay $20+/month for the same amount of storage from a third party service c:

Thank you for the feedback. It‘s much appreciated and I am glad the feature works well for you

CEbbinghaus:
To update this discussion I have reached out to @o1egl over email asking him to consider adding @TobiasGoerke & Myself to the org to help maintain this project & keep contributions flowing. We will see what happens but I am extremely hesitant to fork the project because of no response. Could be that they are just especially busy or on holiday and with just an extra week we would have gotten what we wanted. Either way we will see what happens 🤞🏻

frontend/src/api/tus.js
```diff
@@ -0,0 +1,81 @@
+import * as tus from "tus-js-client";
+import { tusEndpoint } from "@/utils/constants";
+import store from "@/store";
+import { removePrefix } from "./utils";
+import { tusSettings } from "../utils/constants";
```
o1egl: Use @ for imports for consistency

frontend/src/i18n/en.json
```diff
@@ -185,6 +185,14 @@
     "commandRunnerHelp": "Here you can set commands that are executed in the named events. You must write one per line. The environment variables {0} and {1} will be available, being {0} relative to {1}. For more information about this feature and the available environment variables, please read the {2}.",
     "commandsUpdated": "Commands updated!",
     "createUserDir": "Auto create user home dir while adding new user",
+    "tusUploads": "tus.io - Resumable File Uploads",
```
o1egl: I don't think the information about the upload library is meaningful for users. Let's just call it chunked upload.

frontend/src/views/settings/Global.vue
```diff
@@ -101,6 +101,91 @@
...
+          <h3>{{ $t("settings.tusUploads") }}</h3>
+
+          <i18n path="settings.tusUploadsHelp" tag="p" class="small">
+            <a
+              class="link"
+              target="_blank"
+              href="https://tus.io/"
+              >documentation</a
+            >
+          </i18n>
```
o1egl: Why are we pointing to this doc?

frontend/src/views/settings/Global.vue
```diff
@@ -101,6 +101,91 @@
...
+            }}</label>
+            <input
+              class="input input--block"
+              type="number"
+              v-model.number="settings.tus.chunkSize"
+              id="tus-chunkSize"
+              v-bind:disabled="!settings.tus.enabled"
+              min="0"
+              step="1000000"
+            />
```
o1egl: I think this should be set in a more user-friendly manner (i.e. 500k, 20m, etc...)

frontend/src/views/settings/Global.vue
```diff
@@ -101,6 +101,91 @@
...
+              $t("settings.tusUploadsRetryBaseDelay")
+            }}</label>
+            <input
+              class="input input--block"
+              type="number"
+              v-model.number="settings.tus.retryBaseDelay"
+              id="tus-retryBaseDelay"
+              v-bind:disabled="!settings.tus.enabled || settings.tus.retryCount < 1"
+              min="0"
+              step="1000"
```
o1egl: Why 1000?

frontend/src/views/settings/Global.vue
```diff
@@ -101,6 +101,91 @@
...
+              $t("settings.tusUploadsRetryBackoff")
+            }}</label>
+            <input
+              class="input input--block"
+              type="number"
+              v-model.number="settings.tus.retryBackoff"
+              id="tus-retryBackoff"
+              v-bind:disabled="!settings.tus.enabled || settings.tus.retryCount < 1"
+              min="1"
+              step="0.5"
```
o1egl: Why 0.5?

o1egl: I wouldn't add all those parameters and implement exponential backoff as a default behavior.

http/tus.go
```diff
@@ -0,0 +1,242 @@
...
+
+	return tusHandler, nil
+}
+
+func (th tusHandler) getOrCreateTusHandler(d *data, r *http.Request) (*tusd.UnroutedHandler, error) {
+	// Acquire a worker slot from the pool and release it when done
+	th.workerPool <- struct{}{}
+	defer func() {
+		<-th.workerPool
+	}()
```
o1egl: Why are we limiting it?

http/tus.go
```diff
@@ -0,0 +1,242 @@
...
+
+func (th tusHandler) createTusHandler(d *data, basePath string) (*tusd.UnroutedHandler, error) {
+	uploadDir := filepath.Join(d.user.FullPath("/"), th.uploadDirName)
+	tusStore := filestore.FileStore{
+		Path: uploadDir,
+	}
+	composer := tusd.NewStoreComposer()
+	tusStore.UseIn(composer)
+	if composer.UsesConcater && composer.UsesLocker {
+		log.Printf("WARNING: tusd's concater extension is enabled. This is not recommended as it can cause data loss. ")
```
o1egl: There is no way to configure it. I don't think we need this message here.

http/tus.go
```diff
@@ -0,0 +1,242 @@
...
+
+func getMetadataField(metadata tusd.MetaData, field string) (string, error) {
+	if value, ok := metadata[field]; ok {
+		return value, nil
+	} else {
+		return "", fmt.Errorf("metadata field %s not found in upload request", field)
+	}
+}
+
+func (th tusHandler) handleTusFileUploaded(handler *tusd.UnroutedHandler, d *data, event tusd.HookEvent) error {
```
o1egl:
```suggestion
func (th tusHandler) handleFileUploaded(handler *tusd.UnroutedHandler, d *data, event tusd.HookEvent) error {
```

http/tus.go
```diff
@@ -0,0 +1,242 @@
...
+	if err != nil {
+		return err
+	}
+
+	if len(dir) == 0 {
+		// os.Remove won't remove non-empty folders in case of race condition
+		if err := os.Remove(uploadDir); err != nil {
+			return err
+		}
+	}
```
o1egl: This code can be executed after ` os.MkdirAll(uploadDir, os.ModePerm);` but before the actual upload is started

settings/tus.go
```diff
@@ -0,0 +1,19 @@
+package settings
+
+const DefaultTusChunkSize = 20 * 1000 * 1000 // 20MB
```
o1egl:
```suggestion
const DefaultTusChunkSize = 20 * 1024 * 1024 // 20MB
```

o1egl (commented):
When upgrading the filebrowser from the current version to this branch, chunked upload doesn't work. 

Screenshot 2023-07-17 at 10 17 27
Screenshot 2023-07-17 at 10 21 08

http/tus.go
```diff
@@ -0,0 +1,242 @@
...
+		BasePath:              basePath,
+		StoreComposer:         composer,
+		NotifyCompleteUploads: true,
+	})
+	if err != nil {
+		return nil, fmt.Errorf("unable to create handler: %s", err)
+	}
+
+	// Goroutine to handle completed uploads
+	go func() {
```
o1egl: I wouldn't say I like the idea of hanging forever goroutine per user. In my opinion, it should be implemented differently.

TobiasGoerke:
> When upgrading the filebrowser from the current version to this branch, chunked upload doesn't work.
> 
> Screenshot 2023-07-17 at 10 17 27 Screenshot 2023-07-17 at 10 21 08

Could you please share how to reproduce the issue? How about the log output?

o1egl:
> > When upgrading the filebrowser from the current version to this branch, chunked upload doesn't work.
> > Screenshot 2023-07-17 at 10 17 27 Screenshot 2023-07-17 at 10 21 08
> 
> Could you please share how to reproduce the issue? How about the log output?

When you update from the current filebrowser, tus config doesn't exist. So a struct with GO default values is used. I think t can be fixed similar to [this](https://github.com/filebrowser/filebrowser/blob/c2f1423c02e4736f4c243c3164dc671879e065f3/settings/storage.go#L33-L35)

TobiasGoerke:
> > > When upgrading the filebrowser from the current version to this branch, chunked upload doesn't work.
> > > Screenshot 2023-07-17 at 10 17 27 Screenshot 2023-07-17 at 10 21 08
> > 
> > 
> > Could you please share how to reproduce the issue? How about the log output?
> 
> When you update from the current filebrowser, tus config doesn't exist. So a struct with GO default values is used. I think t can be fixed similar to [this](https://github.com/filebrowser/filebrowser/blob/c2f1423c02e4736f4c243c3164dc671879e065f3/settings/storage.go#L33-L35)

I have taken all your comments into account. There is only a single goroutine now and the chunkSize input is human readable. I couldnt test the upgrade though as I‘m executing filebrowser in a container. Could you please check whether that is working now?

TobiasGoerke:
All linter errors are fixed locally. Could you please trigger the pipeline, @o1egl? Thanks.

http/tus.go
```diff
@@ -0,0 +1,271 @@
...
+	return tusHandler, nil
+}
+
+func (th TusHandler) getOrCreateTusdHandler(d *data, r *http.Request) (*tusd.UnroutedHandler, error) {
+	// Use a mutex to make sure only one tus handler is created for each user
+	th.mutex.Lock()
+	defer th.mutex.Unlock()
+
+	tusdHandler, ok := th.tusdHandlers[d.user.ID]
+	log.Printf("Getting tus handler for user %s with basePath %s\n", d.user.Username, d.user.FullPath("/"))
```
o1egl: If looks like a debug log. Should we remove it?

http/tus.go
```diff
@@ -0,0 +1,271 @@
...
+		if originHeader, ok := r.Header["Origin"]; ok && len(originHeader) > 0 {
+			origin = originHeader[0]
+		}
+		basePath, err := url.JoinPath(origin, th.server.BaseURL, th.apiPath)
+		if err != nil {
+			return nil, err
+		}
+
+		log.Printf("Creating tus handler for user %s on path %s\n", d.user.Username, basePath)
+		tusdHandler, err := th.createTusdHandler(d, basePath) //nolint:govet
```
o1egl: This code is a bit confusing and might be error-prone because it `tusdHandler` shadows the variable created in the surrounding scope. As a workaround, var err error can be defined or a function definition can be changed to  

`getOrCreateTusdHandler(d *data, r *http.Request) (_ *tusd.UnroutedHandler, err error)`

```suggestion
tusdHandler, err = th.createTusdHandler(d, basePath)
```

http/tus.go
```diff
@@ -0,0 +1,271 @@
...
+		return tusHandler, fmt.Errorf("couldn't get settings: %w", err)
+	}
+
+	// Create a goroutine that handles uploaded file events for all users
+	go tusHandler.handleFileUploadedEvents()
+
+	return tusHandler, nil
+}
+
+func (th TusHandler) getOrCreateTusdHandler(d *data, r *http.Request) (*tusd.UnroutedHandler, error) {
```
o1egl: Let's use pointer receivers

http/tus.go
```diff
@@ -0,0 +1,271 @@
...
+func NewTusHandler(store *storage.Storage, server *settings.Server, apiPath string) (TusHandler, error) {
+	tusHandler := TusHandler{}
+	tusHandler.store = store
+	tusHandler.server = server
+	tusHandler.tusdHandlers = make(map[uint]*tusd.UnroutedHandler)
+	tusHandler.notifyNewTusdHandler = make(chan struct{})
+	tusHandler.apiPath = apiPath
+	tusHandler.mutex = &sync.Mutex{}
+
+	var err error
```
o1egl: this var can be removed by changing the func return type to `(_ TusHandler, err error)`

http/tus.go
```diff
@@ -0,0 +1,271 @@
...
+	server               *settings.Server
+	settings             *settings.Settings
+	tusdHandlers         map[uint]*tusd.UnroutedHandler
+	notifyNewTusdHandler chan struct{}
+	apiPath              string
+	mutex                *sync.Mutex
+}
+
+func NewTusHandler(store *storage.Storage, server *settings.Server, apiPath string) (TusHandler, error) {
+	tusHandler := TusHandler{}
```
o1egl: Let's use pointer to reduce data coping 
```suggestion
tusHandler := &TusHandler{}
```

o1egl (commented):
I think trying to hack the default tus handler so that it stores upload files in the user's directory is smelly. I think we can either create a shared download directory or just support chunked uploads without concurrency. It can be implemented without creating tmp files at all. Also, incomplete/broken parallel downloads generate a lot of garbage that should be monitored/removed. So, I propose to just implement custome serial chunked  upload that just appends data to the final files.

TobiasGoerke:
> I think trying to hack the default tus handler so that it stores upload files in the user's directory is smelly. I think we can either create a shared download directory or just support chunked uploads without concurrency. It can be implemented without creating tmp files at all. Also, incomplete/broken parallel downloads generate a lot of garbage that should be monitored/removed. So, I propose to just implement custome serial chunked upload that just appends data to the final files.

Integrated your comments. However, I'm not sure about this one / whether we can get around creating temp files.

Your idea would involve creating a custom [DataStore](https://github.com/tus/tusd/blob/v1.11.0/pkg/handler/datastore.go#L86). However, doing so cannot access the filename of the file to be created until after a chunk has been uploaded: the client doesn't send metadata (such as the filename) at neither `NewUpload` nor `WriteChunk`. This is reflected by the [protocol](https://tus.io/protocols/resumable-upload#post), which doesn't require metadata like the filename to be sent at file creation or chunk upload.

o1egl:
> > I think trying to hack the default tus handler so that it stores upload files in the user's directory is smelly. I think we can either create a shared download directory or just support chunked uploads without concurrency. It can be implemented without creating tmp files at all. Also, incomplete/broken parallel downloads generate a lot of garbage that should be monitored/removed. So, I propose to just implement custome serial chunked upload that just appends data to the final files.
> 
> Integrated your comments. However, I'm not sure about this one / whether we can get around creating temp files.
> 
> Your idea would involve creating a custom [DataStore](https://github.com/tus/tusd/blob/v1.11.0/pkg/handler/datastore.go#L86). However, doing so cannot access the filename of the file to be created until after a chunk has been uploaded: the client doesn't send metadata (such as the filename) at neither `NewUpload` nor `WriteChunk`. This is reflected by the [protocol](https://tus.io/protocols/resumable-upload#post), which doesn't require metadata like the filename to be sent at file creation or chunk upload.

We don't need to use the default tus implementation. The protocol supports uploads using a custom upload url that can be a final link to a file

bauerhunter11:
@TobiasGoerke Thank you for your work on this! I cant seem to get this working 

ERROR: failed to solve: failed to compute cache key: failed to calculate checksum of ref b3855634-0713-4eb1-b007-b8173056090e::7o0ksym9gs1whqgh9pqp58jzz: "/src/filebrowser": not found

Getting this error when I try to build. Dockerhub version seems to be out of date and lacking an exposed port so that isnt working for me either.

TobiasGoerke:
@o1egl I've implemented a custom data store that directly writes uploaded chunks to a file.
It features both parallel and resumable uploads. We no longer need any goroutines or clean-ups.
Let me know what you think. I'm pretty satisfied with how this new mechanism is working.

Edit: parallel uploads don't work reliably, right now. There is a race condition which may cause the files to be assembled in the wrong order when parallel uploads are enabled. We might need to disable parallel uploads after all

Edit 2: disabled parallel uploads. They'd work, if the client library sent the initial POST requests synchronously. The actual data / chunks could then be uploaded asynchronously. For now, it's easier to just disable parallel uploads but note, that they'd work given this prerequisite. @o1egl the PR is ready from my side.

TobiasGoerke:
> @TobiasGoerke Thank you for your work on this! I cant seem to get this working
> 
> ERROR: failed to solve: failed to compute cache key: failed to calculate checksum of ref b3855634-0713-4eb1-b007-b8173056090e::7o0ksym9gs1whqgh9pqp58jzz: "/src/filebrowser": not found
> 
> Getting this error when I try to build. Dockerhub version seems to be out of date and lacking an exposed port so that isnt working for me either.

Which Dockerfile are you using? The one posted above should work. Also, I just updated the docker image `tobiasgoerke/filebrowser:tus` so that you can pull the image off of dockerhub.

bauerhunter11:
> > @TobiasGoerke Thank you for your work on this! I cant seem to get this working
> > ERROR: failed to solve: failed to compute cache key: failed to calculate checksum of ref b3855634-0713-4eb1-b007-b8173056090e::7o0ksym9gs1whqgh9pqp58jzz: "/src/filebrowser": not found
> > Getting this error when I try to build. Dockerhub version seems to be out of date and lacking an exposed port so that isnt working for me either.
> 
> Which Dockerfile are you using? The one posted above should work. Also, I just updated the docker image `tobiasgoerke/filebrowser:tus` so that you can pull the image off of dockerhub.

The error above occurred when using the dockerfile text in the first post on this thread. Repulled your docker image from tobiasgoerke/filebrowser:tus and can confirm it works (port exposed) and tested chunking functionality which also appears to be working. 

@TobiasGoerke - thanks again, hopefully this can finally get merged to master. You rock.

o1egl:
@TobiasGoerke I've made a few changes to this PR.

- Completely removed dependency on tusd
- No more uploads mapping
- Removed the enable setting for chunked uploads. I don't think anyone would want to turn it off.

nicoderoin:
Hey all, rebuilt my Docker image with the latest changes and ran into a couple of issues (that were resolved by rolling back the changes to commit **7a8061adec77482d57a571a90938a98d597f0b44**), but I'm running into the following issue with the latest:

When uploading files from my mobile device (iOS), video files make it through but image files (*.jpeg, *.png) do not, and fail to upload with the following error message:

`Upload failed: Error: tus: unable to resume upload (new upload cannot be created without an endpoint), originated from request (method: HEAD, url: /api/tus/admin/Photos/IMG_8738.png?override=false, response code: 405, response text: , request id: n/a)`

Connection details:
FileBrowser hosted on-prem and exposed to WAN via a CloudFlared tunnel
Attempting to upload from mobile device on the same network via https external link with VPN on

Again, rolling back to the earlier hash resolves the issue in my case

o1egl:
> Hey all, rebuilt my Docker image with the latest changes and ran into a couple of issues (that were resolved by rolling back the changes to commit **[7a8061a](https://github.com/filebrowser/filebrowser/commit/7a8061adec77482d57a571a90938a98d597f0b44)**), but I'm running into the following issue with the latest:
> 
> When uploading files from my mobile device (iOS), video files make it through but image files (*.jpeg, *.png) do not, and fail to upload with the following error message:
> 
> `Upload failed: Error: tus: unable to resume upload (new upload cannot be created without an endpoint), originated from request (method: HEAD, url: /api/tus/admin/Photos/IMG_8738.png?override=false, response code: 405, response text: , request id: n/a)`
> 
> Connection details: FileBrowser hosted on-prem and exposed to WAN via a CloudFlared tunnel Attempting to upload from mobile device on the same network via https external link with VPN on
> 
> Again, rolling back to the earlier hash resolves the issue in my case

I can't reproduce it

TobiasGoerke:
Setting a baseUrl now breaks uploads. Try e.g. `FB_BASEURL="/test/"`

nicoderoin:
> Setting a baseUrl now breaks uploads. Try e.g. `FB_BASEURL="/test/"`

Initially, I had my baseUrl set to a blank string `FB_BASEURL=""` -- I tried changing this to `FB_BASEURL="/test/"` then `FB_BASEURL="/api/tus/"` and also tried removing the variable entirely, none of which resolved the issue. Adding a value to `FB_BASEURL` like in your recommendation now gives a 404 since the upload URL turns into `https://{sitename}.net/{baseUrl}/{baseUrl}/api/tus/{folder}` (note the duplicated baseUrl). Removing the variable entirely returns the same 405 as having it set to a blank value

If it'll help, I can provide a link to my filebrowser instance

StanislavJochman:
@o1egl I have the same issue. Using Cloudflare tunnels and my config can be seen below. When uploading locally it works fine but when using my domain in throws error as above: `Upload failed: Error: tus: unable to resume upload (new upload cannot be created without an endpoint), originated from request (method: HEAD, url: /api/tus/admin/Photos/IMG_8738.png?override=false, response code: 405, response text: , request id: n/a)`





```docker
version: "3"

services:
  filebrowser:
    image: filebrowser/filebrowser:latest
    ports:
      - 83:80
    volumes:
      - /Config/Filebrowser/filebrowser.db:/database.db
      - /srv/dev-disk-by-uuid-01168884-146b-4cb4-9379-901047374498/Data2/:/srv/Data2
      - /srv/dev-disk-by-uuid-b2edbf29-291e-4009-babf-a0aeae9eddd4/Data/:/srv/Data
      
    environment:
      - FB_BASEURL=/


    restart: unless-stopped
```

Hope it helps.

tomaszkubacki:
Got same issue as @StanislavJochman - I am using Cloudflare as well, so maybe that's it
edit: And previous version (filebrowser/filebrowser:v2.23.0) works perfectly well - so that's related to this PR

Pikamander2:
I haven't tested the dev version but Cloudflare was our main reason for wanting this as well, so hopefully this new issue can be resolved.

CEbbinghaus:
🎉 Its merged 🎉

bauerhunter11:
> @o1egl I have the same issue. Using Cloudflare tunnels and my config can be seen below. When uploading locally it works fine but when using my domain in throws error as above: `Upload failed: Error: tus: unable to resume upload (new upload cannot be created without an endpoint), originated from request (method: HEAD, url: /api/tus/admin/Photos/IMG_8738.png?override=false, response code: 405, response text: , request id: n/a)`
> 
> ```
> version: "3"
> 
> services:
>   filebrowser:
>     image: filebrowser/filebrowser:latest
>     ports:
>       - 83:80
>     volumes:
>       - /Config/Filebrowser/filebrowser.db:/database.db
>       - /srv/dev-disk-by-uuid-01168884-146b-4cb4-9379-901047374498/Data2/:/srv/Data2
>       - /srv/dev-disk-by-uuid-b2edbf29-291e-4009-babf-a0aeae9eddd4/Data/:/srv/Data
>       
>     environment:
>       - FB_BASEURL=/
> 
> 
>     restart: unless-stopped
> ```
> 
> Hope it helps.

Can confirm same issue when using my cloudflare tunnel.

xd003:
This PR has caused issues with uploading files when using cloudflare tunnels, hopefully it will be addressed soon as many users seem to be affected

Related Issues
https://github.com/filebrowser/filebrowser/issues/2624
https://github.com/filebrowser/filebrowser/issues/2621
https://github.com/filebrowser/filebrowser/issues/2650
https://github.com/filebrowser/filebrowser/issues/2588
https://github.com/filebrowser/filebrowser/issues/2593

Pikamander2:
@TobiasGoerke @CEbbinghaus @o1egl 

Relevant solution/workaround here: https://github.com/filebrowser/filebrowser/issues/2588#issuecomment-1664335304

Any thoughts on how we can fix this automatically while keeping chunked uploads?

jniggemann:
> Any thoughts on how we can fix this automatically while keeping chunked uploads?

I think it would be a good stopgap measue to let the instance admin enable / disable tus/chunked uploads. Like in the picture in https://github.com/filebrowser/filebrowser/pull/2145#issuecomment-1408095905 ...

CEbbinghaus:
It seems that the HEAD request is only sent upon a interrupted upload [see more](https://tus.io/faq#how-does-tus-work) and as such it should be (hopefully) only affecting a portion of the uploads rather than all. To fix this [this workaround](https://tus.io/protocols/resumable-upload#x-http-method-override) can be used that takes advantage of the `X-HTTP-Method-Override` to make tus treat a GET request like a HEAD request which should hopefully solve the problem.

masterwishx:
> Relevant solution/workaround here: [#2588 (comment)](https://github.com/filebrowser/filebrowser/issues/2588#issuecomment-1664335304)

strange but this solution ((http.request.uri contains "/api/tus/") bypass cache ) 
is not fixed picture files uploads but all others now working in my case ....
image

alexbelgium:
Hi, sorry to raise this merged thread but we are encountering the same kind of issues with using Filebrowser on HomeAssistant. HomeAssistant relies on an http ingress proxy that seems to refuse HEAD requests. Would there be a way to avoid using HEAD requests, or to disable the feature (as proposed in https://github.com/filebrowser/filebrowser/pull/2145#issuecomment-1682004989) ? Thanks very much!

According to my github stats I have around 6k users of the addon so this would be greatly appreciated by a lot of people :-)

And thanks very much for having implemented this feature !

PR #2206, opened 2022-11-01
Makes breadcrumbs sticky

methatronc:
In a dense folder, having to go back up to change the directory is time consuming : I propose to make the breadcrumbs sticky for practicity. Tested on mobile/tablet/desktop with chromium browser ( edge ), safari and firefox.


methatronc:
only works on dark mode : no variable was set to sync bg colors, closing the pr.

PR #2209, opened 2022-11-04
prevent accessing resource of file if user has no download permission

9d4:
**Description**
I'm working on a school where I use this filebrowser as our local drive. But due to the preview feature, even though students can't download file, they can still preview it. So I made a change where if a user does not have download permission, it means he is not allowed to get resource (file, folder still accessible).

Thank you...


9d4:
hello

PR #2215, opened 2022-11-11
change default settings file

EDIflyer:
**Description**
other documentation uses `/database.db` so change from `/database/filebrowser.db` for consistency (had lots of file not found issues before when trying to use default config!)
(see PR in docs - https://github.com/filebrowser/docs/pull/47)

**Further comments**
Ran into lots of permission issues when setting up in docker, then realised filenames were different between config and example docker script.


EDIflyer:
Bump - still awaiting review

EDIflyer:
Bump

EDIflyer:
Bump to remove stale label

docker/root/defaults/settings.json
```diff
@@ -3,6 +3,6 @@
   "baseURL": "",
   "address": "",
   "log": "stdout",
-  "database": "/database/filebrowser.db",
```
o1egl, line 6: This settings file is related to the linuxserver based container and the path of the file is set correctly

EDIflyer: It should probably be changed elsewhere then? The documentation seemed to vary between `/database/filebrowser.db` and `/database.db`? For example what caught me out was the Alpine example at https://filebrowser.org/installation#docker uses `/database.db` but then the text below to the sample config file at https://github.com/filebrowser/filebrowser/blob/master/docker/root/defaults/settings.json is then `/database/filebrowser.db`? Perhaps it just needs some text added to make it clear you need to change the sample config file to match? (I'll confess when I first set things up I had a mismatch and it took me a while to work out what was going on!).  I think if you don't have a config file it seems to default to expect `/database.db`?

EDIflyer:
Adding comment due to `stale` label - As mentioned above I think still needs some tweaks to create consistency?

docker/root/defaults/settings.json
o1egl: No, the default docker image https://github.com/filebrowser/filebrowser/blob/master/Dockerfile uses the https://github.com/filebrowser/filebrowser/blob/master/docker_config.json for config. The file you are referring to is used in the linuxserver based image. But I agree that the docs need to be changed. Feel free to make a [PR](https://github.com/filebrowser/docs/pulls)

EDIflyer: OK I'll try to look hopefully this weekend (PS just synced my fork but it seemed to appear as ready to merge but not changed yet - will post when it is!)

EDIflyer:
Going to work on this (commenting to remove stale label)

docker/root/defaults/settings.json
EDIflyer: Sorry for the delay - got round to it in https://github.com/filebrowser/docs/pull/52 by tweaking the docs very slightly

o1egl:
@EDIflyer I merged your docs changes. So I think this PR can be closed.

PR #2243, opened 2022-11-30
feat: Expand Globs into filenames before passing to Exec

gd0st:
**Description**
Hello,

I ran into an issue with the command line interface on the filebrowser where it would not parse glob characters (`*`) for commands like `ls` or `grep`.

I have added a step in the arg parser that expands `*` properly.

This is my first time doing extensive development with GoLang, so please make any feed back or suggest changes as I am relatively inexperienced with the language.
I have tested these changes for my own use cases, so therefore further testing, or a testing set would be helpful here.

I need someone to test these changes against Windows and Mac aswell, because this was developed in a GNU/Linux environment.
Here are some of the screenshots of the old versus new behavior.

**Before**

image

**After**

image

image

I have not done any CI testing, but I will certainly do more after some feedback.

Happy Holidays!

-Yannick


gd0st:
Haven't received any response in two days. Would like to get some feedback on this please.

PR #2285, opened 2023-01-05
Fix build on FreeBSD and non-Linux platforms

yonas:
**Description**
`bash` is installed in `/usr/local/bin/bash` on FreeBSD. This patch should make the build work without modification for FreeBSD and Linux.

- [?] DO make sure you are requesting to **pull a topic/feature/bugfix branch** (right side). Don't request your master!
- [?] DO make sure you are making a pull request against the **master branch** (left side). Also you should start *your branch* off *our master*.


yonas:
Bump. Fix commit linting error.

PR #2326, opened 2023-02-01
file upload with paste image/screenshot from clipboard

jkdoshi:
This implements paste-to-upload for files that are in the clipboard. The most common use-case is;

- take a screenshot using any screenshot tool (I'm using Flameshot)
- copy the screenshot into the clipboard without saving it to a file (press Ctrl-C)
- Switch to filebrowser's file listing page
- press Ctrl-V to paste the clipboard contents
- the client uploads a new file named 'image.png' to the server (without ever saving it to local filesystem)

The above gives a very nice user-experience and saves lots of time and keystrokes.

The screen recording GIF here - https://i.imgur.com/AWLENUZ.gif

### DIFF
--- a/frontend/src/components/header/HeaderBar.vue
+++ b/frontend/src/components/header/HeaderBar.vue
@@ -1,6 +1,8 @@
 <template>
   <header>
-    <img v-if="showLogo !== undefined" :src="logoURL" />
+    <a :href="home">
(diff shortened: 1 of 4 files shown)


o1egl (commented):
Nice idea, but unfortunately, it doesn't work on macOS

jkdoshi:
Okay, let me test on a mac and  update the PR.

PR #2348, opened 2023-02-20
feat: add oidc auth method

marekful:
**Description**

**What**

Adds support for Open ID Connect using `github.com/coreos/go-oidc/v3` to allow user authentication with an external Identity Provider via OAuth2. Only Open ID Connect Discovery is supported, that is, an Issuer URL must be provided and other endpoints and parameters are retrieved from the IdP.

Adds the `oidc` option for `auth.method` and four new configuration settings: `oidc.clientID`, `oidc.clientSecret`, `oidc.issuer`, `oidc.rediredtURL`.

The IdP must send the `preferred_username` claim and authentication succeeds if a File Browser user with matching username exists. The internal JWT token based authorization and session management is used following OIDC authentication similar to the `json` authentication method.

The path in `redirectURL` must be set to `/api/login` in File Browser configuration and, because a `?redirect=...` may be appended to this (e.g. `?redirect=/files/`), the IdP must be configured to allow for this dynamic part in the redirect URL.

***

**How**

When `auth.method` is `oidc` and the `auth` cookie is not set, `http.handleWithStaticData` is intercepted and an "OAuth2 Authorization Code" flow is initiated by redirecting the user to the discovered authorisation endpoint.

The callback is set to `/api/login` and the `Auth` interface method receives the code in the callback request's query from the external IdP which it then processes.

If a user is found in File Browser internal database by the `preferred_username` claim extracted from the OAuth ID Token then a new JWT token is generated and saved in the `auth` cookie.

The fronted then uses this token in the next verify request and persist the resulting state.

**Further comments**

There is an increasing number of active, open source software for running an Identity Provider which can be self hosted, and an ever growing interest by tech minded people and companies for using such solutions. The added security of OIDC authentication may be desired should someone wish to publicly expose File Browser and the convenience of SSO is a bonus.


meredrica:
@o1egl couldn't this be merged? I'm sure a lot of people (like me) would like this. there's a lot of oauth providers but not many that provide forward auth

Nomz420:
@o1egl please look into this

PR #2349, opened 2023-02-21
fix: error while using fallback of dir move

### LINKED ISSUE #1936: HTTP 500 Error: Moving directory from one location to another
**Description**
Trying to move an empty dir from one location to another dir in another location

**Expected behaviour**
Dir to move to the other location

Getting 500 Errors

**How to reproduce?**

- Please create a directory A in the home folder. 
- Please create another directory B in the home folder.
- Try copying A into B

slhmy:
Fix #1936 

## What's actually wrong in this ISSUE

When a folder need to recover an existing one, `fs.Rename` will not work and `MoveFile` function will going a fallback.
However, the fallback usage is not correct and will cause error when it start to **copy & remove the folder** but **treated it in a single file way**.

## How I fix this isssue

By using `Copy` install of `CopyFile` & `RemoveAll` install of `Remove`. `CopyFile` & `Remove` are only effective when the moving object is file.

### DIFF
--- a/fileutils/file.go
+++ b/fileutils/file.go
@@ -17,12 +17,12 @@ func MoveFile(fs afero.Fs, src, dst string) error {
 		return nil
 	}
 	// fallback
-	err := CopyFile(fs, src, dst)
+	err := Copy(fs, src, dst)
 	if err != nil {
 		_ = fs.Remove(dst)
 		return err
 	}
-	if err := fs.Remove(src); err != nil {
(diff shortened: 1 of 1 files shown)


slhmy:
@o1egl sorry for calling, but I think this PR needs your approval.

azcn2503:
This should not have been closed. The change fixes a known issue.

markwoodward86:
Please review, this is not fixed.

PR #2370, opened 2023-03-13
Audit Logger

### LINKED ISSUE #2369: Audit Log
As an administrator, I want to be able to see who made changes to the files and when, so that any misuse can be tracked.

As administrator of the system I can't see which files were edited by which user.
If there was a misuse, I can't see which user was responsible.

An audit log tracks the following events:
- File created / uploaded
- File edited
- File moved / renamed
- File deleted

The audit logs are part of the standard server logs.

The audit log is managed in a separate file and can be accessed by an administrator via the UI.

SRTigers98:
**Description**

Implemented an audit logger for creating, updating, patching and deleting a resource via the REST API.
The current implementation uses the default logger of the builtin log package to log the audit log messages.

However the implementation can easily be changed to use another log destination e.g. a file.
This would be useful if the audit log should be written to a separate file to show it in the UI.

Fixes #2369

### DIFF
--- a/audit/audit.go
+++ b/audit/audit.go
@@ -4,10 +4,9 @@ import "log"
 
 var auditLogger *log.Logger
 
-func init() {
-	setupAuditLogger()
-}
-
-func setupAuditLogger() {
-	auditLogger = log.Default()
+func getAuditLogger() *log.Logger {
+	if auditLogger == nil {
+		auditLogger = log.Default()
+	}
+	return auditLogger
 }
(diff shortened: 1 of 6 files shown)


SRTigers98:
@o1egl Can you please review the PR?

christaikobo:
@SRTigers98 If you are still interested, there is a fork which is under active development, you can open a PR there. It has already solved several problems I had with filebrowser. https://github.com/gtsteffaniak/filebrowser/

PR #2403, opened 2023-04-07
Add Custom link Option at share panel

langren1353:
PR：when user share links they may want to custom the link of the url.
Changes：
- [x] frontend/src/components/prompts/Share.vue Share面板更新
- [x] frontend/src/api/share.js API地址变更
- [x] http/share.go httpData变更
- [x] share/share.go serverHanldle Logic
- [x] i18n Langs


frontend/src/components/prompts/Share.vue
```diff
@@ -167,23 +198,38 @@ export default {
...
+        let query = {
+          url: this.url,
+          password: this.password,
+          custom: this.custom,
+          customLink: this.customLink,
+        };
+
+        if (this.custom) {
+          if (!this.checkCustomLink()) {
+            alert("自定义链接，只支持英文和数字");
```
o1egl, line 219: Please use i18n for displayed messages

frontend/src/components/prompts/Share.vue
```diff
@@ -196,6 +242,18 @@ export default {
         this.$showError(e);
       }
     },
+    checkCustomLink() {
+      if (this.custom) {
+        return /[\w]/.test(this.customLink);
```
o1egl, line 247: I think the better validation would look like `^[a-z0-9]+(?:-[a-z0-9]+)*$`

frontend/src/components/prompts/Share.vue
```diff
@@ -196,6 +242,18 @@ export default {
...
+      }
+      return true;
+    },
+    customChange(tf) {
+      console.log("数据发生变化");
```
o1egl, line 252: Debug print?

frontend/src/components/prompts/Share.vue
```diff
@@ -196,6 +242,18 @@ export default {
         this.$showError(e);
       }
     },
+    checkCustomLink() {
+      if (this.custom) {
+        return /[\w]/.test(this.customLink);
+      }
+      return true;
+    },
+    customChange(tf) {
```
o1egl, line 251: What `tf` stands for? It's better to give this variable a proper name

frontend/src/i18n/he.json
```diff
@@ -228,6 +228,7 @@
     "setDateFormat": "הגדר פורמט תאריך",
     "settingsUpdated": "ההגדרות עודכנו!",
     "shareDuration": "משך השיתוף",
+    "shareCustomLink": "Aangepaste koppeling",
```
o1egl, line 231: It doesn't look correct

share/share.go
```diff
@@ -4,6 +4,9 @@ type CreateBody struct {
 	Password string `json:"password"`
 	Expires  string `json:"expires"`
 	Unit     string `json:"unit"`
+
+	Custom     bool `json:"custom"`
+	CustomLink string `json:"customLink"`
```
o1egl, line 9: Why do we need 2 variables for that? CustomURL would be enough

http/share.go
```diff
@@ -90,13 +90,19 @@ var sharePostHandler = withPermShare(func(w http.ResponseWriter, r *http.Request
...
+		str = body.CustomLink
+	} else {
+		bytes := make([]byte, 6) //nolint:gomnd
+		_, err := rand.Read(bytes)
+		if err != nil {
+			return http.StatusInternalServerError, err
+		}
+
+		str = base64.URLEncoding.EncodeToString(bytes)
 	}
```
o1egl, line 104:
```suggestion
	str := body.CustomLink
	if len(body.CustomLink) == 0 {
		bytes := make([]byte, 6) //nolint:gomnd
		_, err := rand.Read(bytes)
		if err != nil {
			return http.StatusInternalServerError, err
		}

		str = base64.URLEncoding.EncodeToString(bytes)
	}
```

o1egl (commented):
The backend must make an additional validation to be sure that there is no other record with the same url

PR #2506, opened 2023-06-17
feat: add support for sql-based databases

pkuwwt:
**Description**
Add sql support in addition to bolt, including sqlite3, postgres, mysql.

Most code are in `storage/sql`, except a few lines in `cmd/root.go` and `cmd/utils.go`.

**Usage**

```bash
filebrowser -d sqlite3://test.db
filebrowser -d 'postgres://user:password@192.168.1.100:5432/postgres?sslmode=disable'
```

**Reason**
Bolt is a relative non-mainstream choice for storage. When someone use filebrowser as a component of a project, he/she will need a centralized storage server, e.g. postgres, mysql, other than bolt.


JuanRodenas:
Hello!
When will this feature be merged?
Best regards

pkuwwt:
conflicts resolved @o1egl

sevmonster:
Can any effort be put into merging this? Filebrowser CLI cannot be used inside container or while it's running in general, because of the exclusive database lock. This is seriously inconvenient. I am forced at this point to either kill filebrowser every time I want to run a CLI command, or build this PR for use in production. This also affects automation and makes it difficult to programmatically add users (for example) based on other input. I would use the API but it is not documented and seems fiddly to use, #2551.

sevmonster:
I built this and merged with `master`. Works with Postgres, but I've found some issues.
1. The `github.com/mattn/go-sqlite3` library isn't being updated anymore, and the last release is from 2022. It didn't build successfully on my machine, and I had to update it to the latest Git commit `v0.0.0-20240202170028-6ee3e6746e6b` to build.
2. ~~`AuthMethod` of type `proxy` does not seem to work. It either isn't grabbing the header from the `auther` row, or is not being properly substituted in the SQL statement. I am investigating.
    The statement being run:~~
    ```sql
    SELECT id,username,password,scope,locale,lockpassword,viewmode,perm,commands,sorting,rules,hidedotfiles,dateformat,singleclick FROM "fb_users" WHERE username=''
    ```
    ~~As a consequence of this, I see the login page since it is shown when the user either doesn't exist or isn't authed properly, as noted in many other issues.~~
    Actually, this seems to be an upstream bug. #2995
3. `config set` attempts to insert already existing rows, like `Root` and `Socket`. I had to delete all the offending rows to allow it to re-insert them in order to use the command and add new values.

pkuwwt:
@sevmonster `config set` problem solved by insert-or-update

hacdias:
Hey! Thanks for all the effort put into this. I do agree that this could be useful, but it's quite a large change and effort to maintain. As such, I am parking this now, but leaving the PR open.

hacdias:

PR #1554, opened 2021-09-13
Hook authentication method

ramiresviana:
Implements an auther that allows to specify a command that will be executed when a user tries to logs in, provides credentials via environment variables and interprets the standard output to control the authentication. This is the simple way that i found for users to extend the authentication system, the usage is uncomplicated and the code should be easy to maintain. In addition to existing hooks, it makes FileBrowser able to fulfill more specific needs and to be integrated with different systems and technologies.

### Usage
```
filebrowser config set --auth.method=hook --auth.command="/path/to/command arg"
```

### Specification

- Output from the hook command must be on a valid key-value format separated by line, invalid parameters are ignored and authentication is blocked when no hook action is provided.
- Controlling the authentication is done by using `allow`, `block` and `pass` keywords. Each keyword changes the behavior of the login, it can grant or prevent the user authentication with `allow` and `block`, and the `pass` keyword is used for the database based authentication.
- Passwords are stored and updated on success login attempts, allowing it to be used with the database based authentication method.
- User settings can be provided to the hook output, allowing customized experience per user basis, default values are used for omitted fields. For existing users, the settings can be changed by providing the wanted values, omitted fields will remain unchanged.

### Hook command output fields

```ini
hook.action        = auth|block|pass
user.scope         = /path/to/
user.locale        = ar|de|en|es|fr|is|it|ja|ko|nl-be|pl|pt-br|pt|ro|ru|sv-se|zh-cn|zh-tw
user.viewMode      = list|mosaic
user.singleClick   = true|false
user.sorting.by    = name|size|modified
user.sorting.asc   = true|false
user.commands      = cmd1 cmd2...
user.hideDotfiles  = true|false
user.perm.admin    = true|false
user.perm.execute  = true|false
user.perm.create   = true|false
user.perm.rename   = true|false
user.perm.modify   = true|false
user.perm.delete   = true|false
user.perm.share    = true|false
user.perm.download = true|false
```

Output example, trailing whitespace is ignored:

```ini
hook.action        = auth
user.scope         = /home/
user.locale        = en
user.viewMode      = mosaic
user.singleClick   = false
user.sorting.by    = modified
user.sorting.asc   = false
user.commands      = ls rm mkdir touch
user.hideDotfiles  = false
user.perm.admin    = false
user.perm.execute  = true
user.perm.create   = true
user.perm.rename   = true
user.perm.modify   = true
user.perm.delete   = true
user.perm.share    = true
user.perm.download = true
```

### Hook command examples

[...]

PR #2509, opened 2023-06-19
Fix hook Method login rules were erased and rule permissions were add…

zhizhi10:
**Further comments**
1.Fix hook Method login rules were erased and rule permissions were added in the validation script.
2.You can add the following rule to the validation script, and the hook won't erase your existing rules when you don't use it.  
Hook method from https://github.com/filebrowser/filebrowser/pull/1554
`echo 'user.perm.rules     = [{"regex": false, "allow": true, "path": ".", "regexp": {"raw": ""}}, {"regex": false, "allow": false, "path": ".", "regexp": {"raw": ""}}]''
`


auth/hook.go
```diff
@@ -261,6 +264,7 @@ var validHookFields = []string{
 	"user.perm.delete",
 	"user.perm.share",
 	"user.perm.download",
+	"user.perm.rule",
```
o1egl:
```suggestion
	"user.perm.rules",
```

auth/hook.go
```diff
@@ -228,6 +230,7 @@ func (a *HookAuth) GetUser(d *users.User) *users.User {
 			Asc: a.Fields.GetBoolean("user.sorting.asc", d.Sorting.Asc),
 			By:  a.Fields.GetString("user.sorting.by", d.Sorting.By),
 		},
+		Rules:        a.Fields.GetRules("user.perm.rule", d.Rules),
```
o1egl:
```suggestion
		Rules:        a.Fields.GetRules("user.perm.rules", d.Rules),
```

auth/hook.go
```diff
@@ -300,3 +304,20 @@ func (hf *hookFields) GetArray(k string, dv []string) []string {
...
 	return dv
 }
+
+func (hf *hookFields) GetRules(k string, dv []rules.Rule) []rules.Rule {
+	val, ok := hf.Values[k]
+	if !ok {
+		return dv
+	}
+	var dvv []rules.Rule
+	for _, ruleString := range strings.Split(val, ";") {
```
o1egl: If we expect a rule to be encoded as a json, why not to simply return a json array instead of semicolon separated  objects?

auth/hook.go
```diff
@@ -300,3 +304,20 @@ func (hf *hookFields) GetArray(k string, dv []string) []string {
...
+		var rule rules.Rule
+		err := json.Unmarshal([]byte(ruleString), &rule)
+		if err != nil {
+			return dv
```
o1egl: Let's write this error to the log for better auditability

o1egl (changes requested):

zhizhi10 (commented):
I have modified the code according to your suggestion

auth/hook.go
```diff
@@ -300,3 +304,17 @@ func (hf *hookFields) GetArray(k string, dv []string) []string {
 	}
 	return dv
 }
+
+func (hf *hookFields) GetRules(k string, dv []rules.Rule) []rules.Rule {
+	val, ok := hf.Values[k]
+	if !ok {
+		return dv
+	}
+	var Rules []rules.Rule
```
o1egl: In go by convention all the variables should start from the lower case. You can rename it to `userRules` or something like that

o1egl (changes requested):
Please fix go lint errors

zhizhi10:
图片
Rename `Rules` to `userRules`
Fixed go lint errors

PR #2510, opened 2023-06-20
Add a healthcheck script that works with a dynamic port.

andrew-kennedy:
**Description**
This PR modifies the docker healthcheck command to run an external script that supports a custom port from the `.filebrowser.json` config file.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,10 +1,14 @@
 FROM alpine:latest
 RUN apk --update add ca-certificates \
                      mailcap \
-                     curl
+                     curl \
+                     jq
+
+COPY healthcheck.sh /healthcheck.sh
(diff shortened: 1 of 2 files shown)


o1egl:
What is the reason for changing the port inside a docker container?

andrew-kennedy:
> What is the reason for changing the port inside a docker container?

I have multiple containers within the same network namespace in podman so that they can all resolve each other via DNS, but this means they also share a port namespace and cannot all listen on the same ports. Port mapping to a different host port is orthogonal, and so I just change the ports my container software listens on should they collide instead of assuming port 80. For example, in my use case, port 80 in that namespace is being listened on by a reverse proxy.

PR #2550, opened 2023-07-15
feat: add Read/Write Permission

bear0412:
**Description**

Add Read/Write permission to filebrowser.
It works like ACL in linux

**Further comments**

None.

### DIFF
--- a/frontend/src/components/settings/Rules.vue
+++ b/frontend/src/components/settings/Rules.vue
@@ -4,20 +4,15 @@
       <input type="checkbox" v-model="rule.regex" /><label>Regex</label>
       <input type="checkbox" v-model="rule.allow" /><label>Allow</label>
 
-      <input
-        @keypress.enter.prevent
-        type="text"
-        v-if="rule.regex"
-        v-model="rule.regexp.raw"
-        :placeholder="$t('settings.insertRegex')"
-      />
-      <input
-        @keypress.enter.prevent
-        type="text"
-        v-else
-        v-model="rule.path"
-        :placeholder="$t('settings.insertPath')"
-      />
+      <input @keypress.enter.prevent type="text" style="margin-right: 8px;" v-if="rule.regex" v-model="rule.regexp.raw"
+        :placeholder="$t('settings.insertRegex')" />
+      <input @keypress.enter.prevent type="text" style="margin-right: 8px;" v-else v-model="rule.path"
+        :placeholder="$t('settings.insertPath')" />
+
+      <input type="checkbox" :checked="rule.perm && rule.perm.includes('read')"
+        @input="changePermOfRule('read', index)" /><label>Read</label>
+      <input type="checkbox" :checked="rule.perm && rule.perm.includes('write')"
+        @input="changePermOfRule('write', index)" /><label>Write</label>
 
       <button class="button button--red" @click="remove($event, index)">
         -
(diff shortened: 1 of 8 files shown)


http/share.go
```diff
@@ -69,6 +73,10 @@ var shareGetsHandler = withPermShare(func(w http.ResponseWriter, r *http.Request
 })
 
 var shareDeleteHandler = withPermShare(func(w http.ResponseWriter, r *http.Request, d *data) (int, error) {
+	if !d.CheckReadPerm(r.URL.Path) {
+		return http.StatusForbidden, nil
+	}
```
o1egl, line 78: I don't think read permission should be checked here

http/data.go
```diff
@@ -47,6 +48,64 @@ func (d *data) Check(path string) bool {
...
+	}
+
+	write := true
+	for _, rule := range d.settings.Rules {
+		if rule.Matches(path) {
+			if !rule.Allow {
+				write = false
+			} else {
+				write = strings.Contains(rule.Perm, "write")
+			}
```
o1egl:
```suggestion
			if !rule.Allow {
				write = false
				continue
			}
			write = strings.Contains(rule.Perm, "write")
```

http/data.go
```diff
@@ -47,6 +48,64 @@ func (d *data) Check(path string) bool {
...
+		}
+	}
+
+	for _, rule := range d.user.Rules {
+		if rule.Matches(path) {
+			if !rule.Allow {
+				write = false
+			} else {
+				write = strings.Contains(rule.Perm, "write")
+			}
```
o1egl:
```suggestion
			if !rule.Allow {
				write = false
				continue
			}
			write = strings.Contains(rule.Perm, "write")
```

o1egl (changes requested):
It fails with the error.

Screenshot 2023-07-17 at 12 10 13

rules/rules.go
```diff
@@ -17,6 +19,13 @@ type Rule struct {
...
 	Path   string  `json:"path"`
 	Regexp *Regexp `json:"regexp"`
+	Perm   string  `json:"perm"`
+}
+
+// Perm is a read/write permission.
+type Perms struct {
+	Read  bool `json:"read"`
+	Write bool `json:"write"`
 }
```
o1egl, line 29: Looks like this struct is not used

frontend/src/components/settings/Rules.vue
```diff
@@ -37,6 +50,20 @@ export default {
...
+          ? rule.perm.includes("read")
+          : !rule.perm.includes("read");
+      const isWrite =
+        ruleName === "write"
+          ? rule.perm.includes("write")
+          : !rule.perm.includes("write");
+      this.rules[index].perm = `${!isRead ? "read" : ""}${
+        !isRead && !isWrite ? "|" : ""
+      }${!isWrite ? "write" : ""}`;
+    },
```
o1egl, line 66: Why not to store it as an inner object?
```
perms: {read: true, write: true}
```

http/data.go
```diff
@@ -47,6 +48,64 @@ func (d *data) Check(path string) bool {
...
+	}
+
+	read := true
+	for _, rule := range d.settings.Rules {
+		if rule.Matches(path) {
+			if !rule.Allow {
+				read = false
+			} else {
+				read = strings.Contains(rule.Perm, "read")
+			}
```
o1egl, line 63: Let's follow the same code style as in CheckWritePerm

http/data.go
```diff
@@ -47,6 +48,64 @@ func (d *data) Check(path string) bool {
...
+		}
+	}
+
+	for _, rule := range d.user.Rules {
+		if rule.Matches(path) {
+			if !rule.Allow {
+				read = false
+			} else {
+				read = strings.Contains(rule.Perm, "read")
+			}
```
o1egl, line 73: Same here

PR #2579, opened 2023-07-30
Remove redundant calls to baseURL/url #2581

maeryo:
There is an issue where `${baseURL}` may be called twice when uploading through tus.

The scenario can be reproduced as follows:

1. Run `./filebrowser --baseUrl baseUrl`
2. When attempting to upload, a 404 error occurs as the POST request is invoked with `baseUrl/url/baseUrl/url`
3. However, there is no redundant call in the subsequent HEAD request

This seems to happen because, as the `const tusEndpoint` of `constants.js` is transferred to the `fetchURL` of `utils.js`, the predefined `baseURL` is added once more to the fetch

To resolve this, the `baseURL` is removed before calling `fetchURL` if it already exists.

For more details, please refer to the following issue. (#2581)

### DIFF
--- a/frontend/src/api/files.js
+++ b/frontend/src/api/files.js
@@ -2,7 +2,7 @@ import { createURL, fetchURL, removePrefix, removeTusEndpoint } from "./utils";
 import { baseURL } from "@/utils/constants";
 import store from "@/store";
 import { upload as postTus, useTus } from "./tus";
-import {tusEndpoint} from "../utils/constants";
+import { tusEndpoint } from "../utils/constants";
 
 export async function fetch(url) {
   url = removePrefix(url);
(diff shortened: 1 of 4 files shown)


o1egl (changes requested):
I think it is better to fix the code which calls this function.

mbbyn:
As far as I can tell, the root cause is that constants.js `tusEndpoint` is defined with `baseUrl` prefix, then `fetchUrl` is adding `baseUrl` once more.

I would propose the removal of the `{baseUrl}` from the constant if it is not used anywhere else, or introduce a new constant `bareTusEndpoint` that simply doesn't have the prefix and use that when passing to `fetchUrl`.

https://github.com/filebrowser/filebrowser/blob/efd41cc4c147b8d2d5e61fb2642df8d934f49362/frontend/src/utils/constants.js#L20

https://github.com/filebrowser/filebrowser/blob/efd41cc4c147b8d2d5e61fb2642df8d934f49362/frontend/src/api/utils.js#L6-L14

maeryo:
@o1egl 
I have made a new commit following your advice.
What do you think about this approach?

o1egl:
I've pushed a commit to this branch that does what I meant. I don't think we need to do the "magic" of stripping a prefix from the URL. We can simply construct the correct one 😃

maeryo:
Cool! Looks great. Much cleaner code!

Issue #2575, opened 2023-07-29
Is tus directory upload possible in 2.24.0?


maeryo:
There is no problem uploading using tus.
However, uploading the directory throws a 404 Not found.
This is possible only when a directory with the same name exists.

o1egl:
Which browser are you using? Can't reproduce it in chrome

maeryo:
@o1egl 
I solved it by directly modifying tusPostHandler() in tus_handlers.go file.

Old
```
		switch {
		case errors.Is(err, afero.ErrFileNotFound):
			if !d.user.Perm.Create || !d.Check(r.URL.Path) {
				return http.StatusForbidden, nil
			}
		case err != nil:
			return errToStatus(err), err
		}
```

New
```
switch {
		case errors.Is(err, afero.ErrFileNotFound):
			if !d.user.Perm.Create || !d.Check(r.URL.Path) {
				return http.StatusForbidden, nil
			}

			dirPath := filepath.Dir(r.URL.Path)
			if _, err := d.user.Fs.Stat(dirPath); os.IsNotExist(err) {
				if err := d.user.Fs.MkdirAll(dirPath, 0755); err != nil {
					return http.StatusInternalServerError, err
				}
			}
		case err != nil:
			return errToStatus(err), err
		}
```

When I modified it as above, I confirmed that it works without problems.
Thank you.

PR #2580, opened 2023-07-30
fix: add directory creation code (#2575)

maeryo:
**Description**
This is the code for the problem that 404 Not found occurs when uploading a directory using tus. (https://github.com/filebrowser/filebrowser/issues/2575)

@o1egl
please review this code.
thank you.

### DIFF
--- a/files/file.go
+++ b/files/file.go
@@ -24,6 +24,7 @@ import (
 )
 
 const PERM = 0664
+const DirPERM = 0755
 
 // FileInfo describes a file.
 type FileInfo struct {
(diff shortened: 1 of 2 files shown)


files/file.go
```diff
@@ -24,6 +24,7 @@ import (
 )
 
 const PERM = 0664
+const DirPERM = 0755
```
o1egl:
```suggestion
const DIR_PERM = 0755
```

o1egl (commented):
Great job! Small suggestion

Issue #2568, opened 2023-07-24
Cannot click anywhere as everything greyed out


bari86:
Hi,
When I click, for example New Folder, then a popup window will come up with everything else at background is greyed out. Then, if I click cancel, the background greyed out is still there making me cannot click anything else. I can only use Filebrowser again if I do a refresh. I tried on chrome, but on firefox it is fine. I am using Caddy for reverse proxy. I just map my subdomain to localhost:8080. I did ignore most of the settings in the Nginx section here as I am unsure how to put in caddy just now. Is this the cause or something else?

in caddy i just use this:
`file.abc.com {
    reverse_proxy 10.0.0.1:8080
}`

sample nginx code found in github:
`location /filebrowser {
    # prevents 502 bad gateway error
    proxy_buffers 8 32k;
    proxy_buffer_size 64k;

    client_max_body_size 75M;

    # redirect all HTTP traffic to localhost:8080;
    proxy_pass http://localhost:8080;
    proxy_set_header X-Real-IP $remote_addr;
    proxy_set_header Host $http_host;
    proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
    #proxy_set_header X-NginX-Proxy true;

    # enables WS support
    proxy_http_version 1.1;
    proxy_set_header Upgrade $http_upgrade;
    proxy_set_header Connection "upgrade";

    proxy_read_timeout 999999999;
}`

codsec:
Im having the same problem, but im just using in the simple way(windows) on localhost:8080, did u find a possible error or something?

bari86:
I tried using Firefox and it's fine. But if I use chrome it have issues.
Check on your end?

codsec:
> 

Im creating a app with electron to acess with this app, works greats, but chrome looks bug asf idk why

tws101:
Wanted to open an issue myself then read this,
CONFIRMED ISSUE WHEN USING FILEBROWSER ON CHROME

Edge and firefox have NO issue.

The Problem
On any copy, delete or move command after filebrowser completes the operation the screen goes GREY/DIM and no input can be performed.  

Work Around
Refresh, ctrl f5...  Screen brightens back up and input can be performed
or
dont use chrome



My setup Filebrowser runs on a remote truenas server, using the Truecharts version of the app, I remote in on port 10187

codsec:
> 

yeah is basically the same problem

o1egl:
Looks like an issue with a new version of chrome

[...]

PR #2582, opened 2023-07-30
fix: resolved CSS rendering issue in Chrome browser #2568

maeryo:
**Description**

https://github.com/filebrowser/filebrowser/issues/2568

A CSS rendering issue with modal overlays has been identified in recent versions of the Chrome browser.
The previous method of using the display property did not function as intended due to updates in Chrome's rendering engine.

Changes:

- Modified the show keyframe to use the visibility property instead of display.
- Set the .overlay class to use visibility: hidden and opacity: 0 as default values.

These changes ensure that the modal overlay functions correctly across different browser versions, including the latest Chrome update. The proposed solution leverages standard CSS properties and strengthens resistance against potential future rendering changes.

Please review the changes and let me know if you have any feedback.
Thank you!

### DIFF
--- a/frontend/src/css/dashboard.css
+++ b/frontend/src/css/dashboard.css
@@ -319,6 +319,8 @@ body.rtl .card .card-title>*:first-child {
   height: 100%;
   width: 100%;
   z-index: 9999;
+  visibility: hidden;
+  opacity: 0;
   animation: .1s show forwards;
 }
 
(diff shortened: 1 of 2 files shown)


jsdxyz:
hello maeryo, please can you show how to fix it ? the filebrowser is installed with docker

jsdxyz:
this error is only in google-chrome I tryed in Brave, firefox end others.

PR #2616, opened 2023-08-04
Update German translation

derech1e:
- Add missing strings.
- Fix some improper whitespaces.

With the new chunked uploads feature, the question of `continue` a upload instead of replacing it was added. For me this question is very irritating because I think `skip` would be a better term. In my language, the term `continue` is sometimes referenced as something other than to `skip` a file. Anyways the translation should be ready to merge.

image


o1egl:
By pressing "continue", any incomplete upload will be finalized. If you choose "replace", the file upload process will restart from the beginning. We can change the title and description to better explain the behavior

derech1e:
> By pressing "continue", any incomplete upload will be finalized. If you choose "replace", the file upload process will restart from the beginning. We can change the title and description to better explain the behavior

OK, I didn't know that.

PR #2674, opened 2023-08-27
fix: Refactor path resolution logic for project root

maeryo:
With the introduction of Vite in filebrowser, there have been incidents where debug, serve, and build became inoperable. Considering that there were no issues in the git action, it seems likely that this issue can occur under certain circumstances.

My development environment is node v18.16.1 on Windows, but I encounter the problem as shown in the attached screenshot.
I've made a few changes to address this. Please review.

image

### DIFF
--- a/frontend/vite.config.js
+++ b/frontend/vite.config.js
@@ -14,10 +14,15 @@ const plugins = [
   compression({ include: /\.js$/i, deleteOriginalAssets: true }),
 ];
 
+function getProjectRoot() {
+  const currentFile = new URL(import.meta.url).pathname;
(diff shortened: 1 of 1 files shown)


kloon15:
Thats weird, this is straight from what u get after `npm create vue@legacy`.
The problem with ur solution is demonstrated in the documentation:
image
 
Can u try the old way and report if that works for u?

```
import path from "node:path";
...
    "@/": `${path.resolve(__dirname, 'src')}/`,
```

maeryo:
@kloon15 

Your approach works very well.
It's simpler and more fitting than my code, so I will change to the code you've suggested.
Thank you.

PR #2677, opened 2023-08-28
feat: implement upload speed calculation and ETA estimation

maeryo:
I have closed the previous pull request(#2675) and reworked on it.
I resolved conflicts regarding the newly merged commits.
Also, I optimized the previously implemented code and addressed several issues.

---

There has been a significant user demand for transfer speed and ETA for a while.
This Pull Request addresses that need.

Due to the characteristics of the tus protocol, where transfers are made in chunk units, there are frequent starts and stops, leading to a large variance in speed. To address this issue, we have applied a moving average to the speed calculations.

Additionally, to reduce the strain on the threads, the calculations were set to be executed at intervals. Below is a capture of the implementation.

image


maeryo:
@o1egl 
I've rewritten the code and submitted a pull request.
Please review.

thank you! 😄

Issue #2554, opened 2023-07-18
Session Timeout for users


Tarsman013:
**Is your feature request related to a problem? Please describe.**
it is difficult to upload multiple, large files when after a period of not using the client, but still uploading, it logs the user out. This has caused multiple large transfers to fail

**Describe the solution you'd like**
the ability to set a session timeout for each user in minutes or hours (0 would be never)

**Describe alternatives you've considered**
none exist to my knowledge 

**Additional context**
n/a

brianfromm:
@Tarsman013 This has been addressed in PR #5546 (see my comment on your other issue #2553 for full details).

You can now configure the session timeout globally using `tokenExpirationTime`. For existing installations:

```sh
filebrowser config set --tokenExpirationTime 6h
```

**Note:** Setting `0` for "never expire" is not supported. The maximum you can set is something like `8760h` (1 year) if you need very long sessions.

Also, this is currently a global setting, not per-user. All users share the same session timeout value.

PR #2753, opened 2023-10-09
#2554 configure user session timeout

darddan:
**Description**
Session timeout is hardcoded to 2 hours. This PR makes that value configurable with a fallback to 2 hours. Related to #2554

There are two use-cases for me for this change. The first is that I wan't to be able to stay logged in for longer. The other one is if I'm downloading really large files and session expires in between then the upload is canceled at that point. 

IMO for the second use-case this PR shouldn't be enough and it needs another proper fix.


darddan:
Hi @o1egl, can I get a review/comment on the changes? I fixed the linting issues after the CI failed last time

http/auth.go
```diff
@@ -172,11 +174,13 @@ var signupHandler = func(w http.ResponseWriter, r *http.Request, d *data) (int,
...
 }
 
-var renewHandler = withUser(func(w http.ResponseWriter, r *http.Request, d *data) (int, error) {
-	return printToken(w, r, d, d.user)
-})
+func renewHandler(tokenExpireTime time.Duration) handleFunc {
+	return withUser(func(w http.ResponseWriter, r *http.Request, d *data) (int, error) {
+		return printToken(w, r, d, d.user, tokenExpireTime)
+	})
+}
```
o1egl: Please adjust it with the fix in master https://github.com/filebrowser/filebrowser/blob/master/http/auth.go#L175-L178

StanislavJochman:
Where can I set a custom user session timeout? Is it only through API?

akamkomputer:
Is it possible to set TokenExpirationTime in docker?

fritx:
Update: I've submitted a PR for this feature. https://github.com/filebrowser/filebrowser/pull/2845

---

cfg file should work @StanislavJochman @akamkomputer 

```sh
filebrowser -c your/path/to/filebrowser.json
```

```js
// filebrowser.json
{
  "port": 8080,
  "baseURL": "",
  "address": "127.0.0.1",
  "log": "stdout",
  "database": "filebrowser.db",
  "root": ".",
  "tokenExpirationTime": "12h"
}
```

StanislavJochman:
But can it be set using dockerfile? Or do I have to edit file manually?

fritx:
@StanislavJochman oops, my fault/.. Update: I've submitted a PR for this feature. https://github.com/filebrowser/filebrowser/pull/2845

```sh
# One of the below should work:
filebrowser --token-expiration-time=12h
# or
filebrowser -c fb.json
# fb.json
{
  "token-expiration-time": "12h"
}
```

UPDATE:  corrected: `filebrowser --token-expiration-time=12h`, as pointed out by baur.

baur:
implemented: 
```
filebrowser --token-expiration-time=12h
```
not **timeout**

k0gen:
> implemented:
> 
> ```
> filebrowser --token-expiration-time=12h
> ```
> 
> not **timeout**

What are the allowable values and range here? Can "0" be used to disable the timeout?

MattDHill:
@o1egl Can you confim how we can disable the expiration altogether such that sessions never expire?

PR #2754, opened 2023-10-10
add login test scenarios only

nabim777:
Here only writing the test and the login test scenario added


nabim777:
Sorry! I was practicing for test automation and I was supposed to make PR in the forked repo but mistakenly make PR in main Repo

Issue #2721, opened 2023-09-18
Upload events do not work.


mourisio:
**Description**
I use the latest version of filebrowser - 2.25.0. When I try to run commands from built-in command runner, using after upload or before upload events, it just does not work. But when I create new file via filebrowser interface it works. this is the script I am trying to run for test purpose - /bin/sh -c 'echo "File  is uploaded." >> /home/test.txt'

**Expected behaviour**
Run script after or before uploading a file.

**What is happening instead?**
Script does not work. it only works when I create a file via web interface.

**Additional context**
The error was not occurred in v2.23.0 or less. I have the same problem in higher versions (I tested all of them).

**How to reproduce?**
Go to command runner in web interface.
write script in before or after upload event tab.
upload a file.

kafeinnet:
Duplicate of #2626 I think.

Issue #2626, opened 2023-08-08
Command Runner 'After Upload' not executed after a file is uploaded


saulchristie:
**Description**
'After upload' Command Runner is not executed following a file upload.

**Expected behaviour**
'After upload' Command Runner to be executed following a file upload.

**What is happening instead?**
By way of user experience, nothing is happening. File uploads fine, but no command is executed.

**Additional context**
The error occurs in v2.24.2, running bare metal on Ubuntu. Defined 'after upload' command has not been modified since the functionality worked, and syntactically identical commands defined in other Command Runner areas are still executed as expected.

**How to reproduce?**
Define a command to be executed after a file upload within FileBrowser's 'Command Runner' config.
Upload a file.
Observe the defined command is not executed.

**Files**
N/A

kafeinnet:
I was just going to open an issue about that. I confirm this issue, I have the exact same problem with filebrowser in a docker environment. Reverting to 2.24.1 fix the problem.

With 2.24.1 I can see this log (here I uploaded a `lemmings.mp4` file):
```
2023/08/29 17:37:21 Using config file: /.filebrowser.json
2023/08/29 17:37:21 Listening on [::]:80
2023/08/29 17:38:22 [INFO] Blocking Command: "/opt/ffmpeg/convert.sh /srv/lemmings.mp4"
```

With 2.24.2 the last line never appears.

ghost:
Can confirm, I had to revert to 2.23.0 to get it to work again

ruihaut:
Issue still here on 2.27.0

AndresJ551:
Same, trying to curl "after upload" doesn't work, the same with "before upload".
Have no error message on logs.
Using the same command after delete or after copy works perfectly.

I'm using the latest tag from https://hub.docker.com/r/filebrowser/filebrowser

mi-bauer:
I looked into the codebase and for me it looks like this (please correct me if I'm wrong):
[...]

Issue #2600, opened 2023-08-03
Command Runner: trigger before_upload, after_upload not working


baur:
**Description**
v2.24.1

Command Runner: trigger before_upload, after_upload not working

**Expected behaviour**
run next command:
```
echo "FILE $FILE, SCOPE = $SCOPE, TRIGGER = $TRIGGER, USERNAME = $USERNAME, DESTINATION = $DESTINATION " >> command.log.txt
```

**What is happening instead?**
nothing is happened

**Additional context**
run command via PowerShell

after_delete
before_delete 

woks fine!

Nayruden:
I am seeing the same thing on v2.25.0 (Docker s6). I've tried everything I can think of, but it never seems to be called under any condition.

PR #2767, opened 2023-10-15
fix: before and after hook on upload event (Fixs #2721 #2626 #2600)

kang-kang:
From 2.24.0 upload changes to tus.io, but the hooks do not be implemented.
There is a request before upload start, but I've not found how to know the upload finished in backend, therefore I implement it in frontend.


o1egl:
Thank you for your contribution, but I think there must be a better way to solve that. Maybe we need to modify the client and send a flag on the last chunk of data.

PR #2847, opened 2023-11-26
Create filebrowser.service

lolpro11:
**Description**
Add Systemd service


o1egl:
I believe that should be included in the documentation, not here

lolpro11:
Fair enough. Could you edit the wiki and add a section "Running as a
systemd service"?

Issue #2530, opened 2023-07-04
Opening image looses scroll position in file list


svefro:
**Description**
when browsing a folder with a lot of images. opening a image and closing it causes the file list to scroll all the way to the top.

**Expected behaviour**
when closing the image i should return to where i was with the image i just viewed selected in the list.

**What is happening instead?**
the file list is empty and loaded again causing scrolling to be reset to the top.

**Additional context**
also it would be nice with a button on each thumbnail to open it in a new viewer or directly the url to the image.

**How to reproduce?**
open a folder with images. scroll down abit. open a image by doubble clicking. close the image with the X. observe empty list. scroll position is reset to the top

**Files**

svefro:
can someone remove the stale label? This has not been resolved.

svefro:
The issue is still here

sweetpalma:
Could agree. Any way to fix this?

svefro:
It is most likely easy for those that know vue.js and have the time and incentive to fix it. I do not know vue.js

svefro:
issue is still not fixed

Rod-fish:
I also have the same issue, and it's even worse for me. When I have thousands of images in my folder, once I find the image I need at the bottom and open it, then close it, it takes me back to the top and I have to search for that image again. This has been driving me crazy. Can we add a pagination feature? It would greatly alleviate this problem.

svefro:
yes it is a real show-stopper for using it to view images.

I do not know enough about vue to fix it. i'm sure it is a easy fix for those that know this project.

svefro:
issue persists as far as i know

sweetpalma:
@svefro I created a quick fix PR regarding for this issue, hopefully it will get merged: https://github.com/filebrowser/filebrowser/pull/2872

svefro:
Perfect. let's hope it gets merged and is included in the next release. :)

maromcik:
> @svefro I created a quick fix PR regarding for this issue, hopefully it will get merged: #2872

I cloned your fork, but the fix does not work for me. 
Steps to reproduce:
Open a folder with many pictures 
Scroll down or into the middle
Open a picture\
Close with escape
Filebrowser still reloads the folder and returns the page to the top.

Is there anything that can be done about it? I'd love this feature. Am I missing something?

sweetpalma:
@maromcik could not reproduce... Did you changed the branch from `master` to `restore-scroll-position`?

PR #2872, opened 2023-12-10
fix: restore scroll position after closing the preview

sweetpalma:
**Description**

Updated frontend portion of application to make it restore the scroll position after closing the preview. I think that this fix is quite important for those who plan to use this tool as a simple frontend for browsing image dumps remotely (including myself).

**Related issue** 

https://github.com/filebrowser/filebrowser/issues/2530

**Demonstration**

Demonstration


o1egl:
Thank you for creating this PR. I tested it, and it doen't work with text file views. Also, when I delete the file from the preview mode, it shows me an error. I think better solution would be to open all those preview views as popups instead of replacing the original view.

sweetpalma:
@o1egl alright, I'll take a look 👍

PR #2906, opened 2023-12-31
feat: add lazy loading folder size

suacrbah:
**Description**

**Further comments**
The original issue #2861  proposed to calculate size of the folder tree every time, this may be expensive when the directory structure is complex and huge, this PR propose a lazy loading method, in which the calculation only happens if you click the `show` text on the Info card.

New Info card view:

before

After clicking on `show`:

after

Also support multi folder aggregation:

multi


frontend/src/components/prompts/Info.vue
```diff
@@ -166,6 +175,26 @@ export default {
         this.$showError(e);
       }
     },
+    foldersize: async function (event) {
```
o1egl:
```suggestion
    folderSize: async function (event) {
```

frontend/src/api/files.js
```diff
@@ -42,6 +42,16 @@ async function resourceAction(url, method, content) {
...
 
+async function resourceSizeAction(url, method, content) {
+  url = removePrefix(url);
+  let opts = { method };
+  if (content) {
+    opts.body = content;
+  }
+  const res = await fetchURL(`/api/resources/size${url}`, opts);
+  return res;
+}
```
o1egl: I don't think we need 2 separate functions for that. It is always a get request with no body. So resourceSizeAction and foldersize can be merged and simplified

files/file.go
```diff
@@ -95,6 +95,37 @@ func NewFileInfo(opts FileOptions) (*FileInfo, error) {
 	return file, err
 }
 
+func NewFolderInfo(opts FileOptions) (*FileInfo, error) {
```
o1egl: I think it makes sense to add a flag to the FileOptions and use NewFileInfo instead. Similar to the Expand flag.

files/file.go
```diff
@@ -95,6 +95,37 @@ func NewFileInfo(opts FileOptions) (*FileInfo, error) {
...
+	}
+	size, err := GetFolderSize(file.RealPath())
+	if err != nil {
+		return nil, err
+	}
+	file.Size = size
+	return file, nil
+}
+
+func GetFolderSize(path string) (int64, error) {
```
o1egl: This function can be private

frontend/src/components/prompts/Info.vue
```diff
@@ -13,11 +13,20 @@
...
         <span id="content_length"></span> {{ humanSize }}
       </p>
 
+      <p v-if="dir || selected.length > 1">
+        <strong>Size: </strong
+        ><code
+          ><a @click="foldersize($event)">{{
+            $t("prompts.show")
+          }}</a></code
+        >
```
o1egl: Please fix code formating

o1egl (commented):
Thank you for creating this PR. I left a few comments. Adding a feature flag that turns it off might be a good idea because it might easily harm the system if used on big folders.

frontend/src/components/prompts/Info.vue
suacrbah: Thank you for the patient code review. The function name filesize() used in the same file was in my mind when naming these functions, should I change filesize() to fileSize() also to make them conformant?

suacrbah:
> Thank you for creating this PR. I left a few comments. Adding a feature flag that turns it off might be a good idea because it might easily harm the system if used on big folders.

I added a switch for turning the folder size off, however, even if the switch is on, folder size is lazy loaded, meaning that it calculates the selected folder size only if the user clicks the `show` button on the Info page.

maeryo:
@suacrbah 
cool!
Is it possible to implement the file list screen with lazy load?
If there are too many files in the directory, a timeout may occur.

suacrbah:
> @suacrbah 
> 
> cool!
> 
> Is it possible to implement the file list screen with lazy load?
> 
> If there are too many files in the directory, a timeout may occur.

Sry I'm not sure what you mean by "file list screen with lazy load", can you kindly explain it a little more?

maeryo:
@suacrbah 
What I meant to say was that the current structure of Filebrowser, which loads all files in a directory at once, can lead to HTTP timeouts when there are too many files. To address this issue, is it possible to introduce 'lazy loading'? This means loading only the necessary files sequentially as the user scrolls down. This approach could allow for an efficient display of files in directories with many files, without encountering timeouts.

suacrbah:
@maeryo 
Actually, I think that's a valid proposal, and can be implemented by using `Readdir(count int)` which reads exactly `count` directory entries in the given directory. However, that might be beyond the scope of this PR.

maeryo:
@suacrbah 
I completely agree that this request might be beyond the scope of this PR.
However, I'm glad to hear that there is a potential way to improve it based on my suggestion.
Thank you!

ThienMD:
This feature is very useful, I hope this PR will be merged soon

http/resource.go
```diff
@@ -19,6 +19,24 @@ import (
 	"github.com/filebrowser/filebrowser/v2/fileutils"
 )
 
+var resourceGetSizeHandler = withUser(func(w http.ResponseWriter, r *http.Request, d *data) (int, error) {
+	folder, err := files.NewFileInfo(files.FileOptions{
+		Fs:         d.user.Fs,
+		Path:       r.URL.Path,
+		Modify:     d.user.Perm.Modify,
+		Expand:     true,
+		FolderSize: true,
```
o1egl, line 28: This feature should be configurable and disabled by default. It might be harmful for public instances.

o1egl (changes requested):
Well done! Please fix linter errors, so we can merge this PR

suacrbah:
@o1egl 
Thank you for the approval, but I'm currently unable to access a computer at hand. I will change the code in about 5 days, sorry for the delay.

niubility000:
> @suacrbah What I meant to say was that the current structure of Filebrowser, which loads all files in a directory at once, can lead to HTTP timeouts when there are too many files. To address this issue, is it possible to introduce 'lazy loading'? This means loading only the necessary files sequentially as the user scrolls down. This approach could allow for an efficient display of files in directories with many files, without encountering timeouts.

We already have this function. --disable-type-detection-by-header, it was added like two or three years ago. We may set it as a default value.
Use it like this:
.\filebrowser.exe -d .\filebrowser.db --disable-preview-resize --disable-type-detection-by-header --cache-dir .\cache

Here's so many functions in filebrowser. I think we need to mention them in an obvious place. A lot of users don't know them.

maeryo:
> > @suacrbah What I meant to say was that the current structure of Filebrowser, which loads all files in a directory at once, can lead to HTTP timeouts when there are too many files. To address this issue, is it possible to introduce 'lazy loading'? This means loading only the necessary files sequentially as the user scrolls down. This approach could allow for an efficient display of files in directories with many files, without encountering timeouts.
> 
> We already have this function. --disable-type-detection-by-header, it was added like two or three years ago. We may set it as a default value. Use it like this: .\filebrowser.exe -d .\filebrowser.db --disable-preview-resize --disable-type-detection-by-header --cache-dir .\cache
> 
> Here's so many functions in filebrowser. I think we need to mention them in an obvious place. A lot of users don't know them.

Thank you for the suggestion.
However, even using the specified arguments does not change the outcome.
A timeout occurs when about 20,000 images, each approximately 2MB in size, are duplicated.
Naturally, the storage is on an SSD, and there is no lack of computing performance.

When I activated debug mode and checked the log, the only information available was as follows:

2024/02/19 06:03:42 /viewer/images/: 500 172.16.236.24 write tcp 10.233.76.235:8080->127.0.0.6:38251: write: broken pipe
2024/02/19 06:03:42 http: superfluous response.WriteHeader call from github.com/filebrowser/filebrowser/v2/http.NewHandler.NewHandler.func2.handle.func19 (data.go:76)

I have reviewed other people's issues and discussions but have not found any content confirming that the argument works correctly.

P.S. I have confirmed that the arguments were applied correctly by checking the db file with the cat command.

PR #2990, opened 2024-02-05
fix: editor discard prompt

Zouizoui78:
**Description**
Currently the editor changes discard prompt is bugged and appears in the following two situations where it should not:
- when the file is empty, like right after its creation
- after the file has been saved

This PR aims to solve both issues.


Zouizoui78:
I changed my commit a bit (forced pushed because of a mistake on my part) to make use of the logic that is already implemented by ace to know if the editor is dirty or not.

PR #3004, opened 2024-02-16
freezing the list in the backgroud while previewing a file

niubility000:
It is to fix #2188, #1870, and it works perfectly.

When you return from the preview, you will go back to the last position of the list.

This is a new PR instead of the old one #3002
the old PR can't pass the checks totally because of  the "type may not be empty [type-empty]" error. When we commit a changed file, we need write a type in the **commit message** box, something like "fix: ......". the default name "Update Files.vue" is totally wrong. It's a little bit misleading.

https://github.com/conventional-changelog/commitlint/?tab=readme-ov-file#config
Common types according to [commitlint-config-conventional (based on the Angular convention)]can be:

build
chore
ci
docs
feat
fix
perf
refactor
revert
style
test

PR #3017, opened 2024-02-24
fix:remain-in-the-same-position-in-list-after-rename-or-delete

### LINKED ISSUE #1870: File list scrolls up when renaming/deleting files.
**Description**
I have a huge library of files in a folder. Im trying to rename them systematically one by one. 
The problem is i always have to scroll down to the last file i renamed to continue.

**Expected behaviour**
I should be able to rename them one by one without the site jumping to the top of the list again.

The page scrolls up to the first item in the list.

**How to reproduce?**
Have a lot of files scroll down and rename one. Same thing happens when deleting a file in the middle.

niubility000:
To fix: #1870, and also to update #3004
The list will stay in the same position with updated items after you delete or rename a file or folder, even if you do it in preview.

### DIFF
--- a/frontend/src/components/prompts/Delete.vue
+++ b/frontend/src/components/prompts/Delete.vue
@@ -45,6 +45,7 @@ export default {
     submit: async function () {
       buttons.loading("delete");
 
+      window.sessionStorage.setItem('modified', "true");
       try {
         if (!this.isListing) {
           await api.remove(this.$route.path);
@@ -74,6 +75,9 @@ export default {
         this.$showError(e);
         if (this.isListing) this.$store.commit("setReload", true);
       }
+      Vue.nextTick(() => {
+        window.sessionStorage.setItem('modified', "false");
+      });
     },
   },
 };
--- a/frontend/src/components/prompts/Delete.vue
+++ b/frontend/src/components/prompts/Delete.vue
@@ -75,9 +75,6 @@ export default {
         this.$showError(e);
         if (this.isListing) this.$store.commit("setReload", true);
       }
-      this.$nextTick(() => {
-        window.sessionStorage.setItem("modified", "false");
-      });
     },
   },
 };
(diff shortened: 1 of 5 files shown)


niubility000:
To remain in the same position of list when rename an item, some projects use "v-model". It allow users to dbl click and modify the name directly in the list without renewing the whole page. But it need change a lot of things in filebrowser. So we can't use it.

And since the route doesn't change when we close the prompt "rename" or "delete", we can't use the "listFrozen" here. I just add a new variable "modified" to monitor the rename or delete process. It's simple  and works perfectly.

frontend/src/components/prompts/Delete.vue
```diff
@@ -45,6 +45,7 @@ export default {
     submit: async function () {
       buttons.loading("delete");
 
+      window.sessionStorage.setItem("modified", "true");
```
niubility000, line 48: This is to monitor the delete or rename behavior.

frontend/src/views/Files.vue
```diff
@@ -117,7 +116,7 @@ export default {
       this.$store.commit("closeHovers");
 
       // Set loading to true and reset the error.
-      if (window.sessionStorage.getItem('listFrozen') !=="true"){ 
+      if (window.sessionStorage.getItem("listFrozen") !=="true" && window.sessionStorage.getItem("modified") !=="true"){
```
niubility000, line 119: "modified" works in the same way as "listFrozen"

frontend/src/views/Files.vue
```diff
@@ -151,3 +150,9 @@ export default {
   },
 };
 </script>
+<style>
+  #listing{
+    height: calc(100vh - 8em);   
+    overflow-y: auto; 
+  }
+</style>
```
niubility000: This is important. It will correct the whole layout of the mainpage. the strollbar will only show in the listing box. and when you are in preview page, you can't see the underground strollbar in the rightside any more.

frontend/src/views/files/Editor.vue
```diff
@@ -1,5 +1,9 @@
 <template>
-  <div id="editor-container">
+  <div 
+    id="editor-container"
+    @touchmove.prevent.stop 
+    @wheel.prevent.stop
```
niubility000, line 5: This is a fix to #3004.

Issue #3008, opened 2024-02-18
Unable to access Filebrowser on the public internet


ghost:
**Description**

**Expected behaviour**
Im running a google cloud linux server, I am trying to run filebrowser using the native linux binary, and access the filebrowser webapp from the servers static public ip

**What is happening instead?**
I get this error: https://pixeldrain.com/u/FC1Mhsps

**Additional context**
In the google cloud console, I have allowed egress and ingress traffic on port 8081 (the port the webapp is running on), tcp and udp, I have disabled ufw (linux firewall), however I still cannot access it from its public ip (e.g: x.x.x.x:8081)
However, when I access localhost:8081 on the server with, I can access the webapp, meaning the webapp is running normally
I have another webapp running on port 8080, which I can access from the public internet perfectly fine

**How to reproduce?**
 Run the native file-browser binary on a server with a public static IP, try to access filebrowser with that public ip

niubility000:
try this: 
.\filebrowser -d filebrowser.db config set --address 0.0.0.0

--address 0.0.0.0 will make you to access to filebrowser from public internet or something like 192.168.1.101. The default value is 127.0.0.1. In some case you can's access to filebrowser from Public internet with this default value.

ghost:
> try this: .\filebrowser -d filebrowser.db config set --address 0.0.0.0
> 
> --address 0.0.0.0 will make you to access to filebrowser from public internet or something like 192.168.1.101. The default value is 127.0.0.1. In some case you can's access to filebrowser from Public internet with this default value.

Thank you very much, this works.

ai-mic-stack:
i am using file broswer it is opening in public ip from outside network but not bale to login giving inacivity error

Issue #2584, opened 2023-07-30
unable to access remotely


leoxu1007:
Successfully accessed from localhost:8080, but unable to access from local network.

o1egl:
You need to provide a correct IP address to the app. Alternatively, use `filebrowser -a 0.0.0.0`

Issue #1836, opened 2022-02-23
Using Filebrowser on an RCLON Mount results in "Something really went wrong." with Folders that have numerous Files in them.


NightHammer1000:
**Description**
Trying to use Filebrowser on a folder that is on a RClone Mount results in a "Something really went wrong." message when the folder has more than ~200 Files in it.

**Expected behaviour**
Files should be listed.

**What is happening instead?**
Nothing is shown at all except "Something really went wrong."

**Additional context**
I do belive this has to do with how Filebrowser lists the Content of that Folder. It runs in some kind of Timeout because it naturally takes longer to look into every single file on an Cloud Mount.
A normal "ls" or "ll" gives me the Result instantly because of my Caching Setup.
I watched what my system is doing with "Netdata" and noticed something.
Opening the folder inside Filebrowser results in IOWait. Which is normal for a Cloud Mount, but only when opening files from the Mount.
But everything needed from the File to be listed should be retrievable from my local cache just like "ls" or "ll" does.

If Filebrowser does indeed open the files to read more Info, a Light Mode for specific Folders, which only builds the List from "ls" or "ll" would be worth to think about in the future.

**How to reproduce?**
Put a Folder with many small files on an RClone Google Drive Mount. (Like a Rom-Set or something like that)
Start Filebrowser and try to access the Folder. 
It will fail.

o1egl:
If you have images there, filebrowser reads them to make thumbnails

NightHammer1000:
Nope. Just ZIPs inside there. Nothing more

The one Folder i tried it with has around 600 zip files inside of it.

In comparison, NGINX with Fancyindex displays the folder in a fraction of a second.

Fukitsu:
I had the same problem and read in another issue about enabling ``--disable-type-detection-by-header``. With it it displays the files without problem

NightHammer1000:
Indeed it does!

Nice find. Thanks!

alfablac:
btw, this is true for general files, one folder with 2k txt files loads instantly, a folder with 600 video files won't, even with all disable flags of filebrowser

Issue #1689, opened 2021-11-29
Performance and sharing files from folders containing large numbers of files (~4000)


menasamuel:
**Description**

Performance and sharing files from folders containing large numbers of files (~4000)

- Navigating to a folder containing a large number of files 
- find one specific file using search 
- share that one file only 

**Expected behaviour**

- displaying the list of files is fast
- searching for a file (instead of scrolling) is fast
- sharing that single file is easy to do

**What is happening instead?**

- displaying a list of files takes 1 minute 
- sharing from the search results page is not possible 
- attempting to share from the search results actually just attempts to preview the file
- dismissing the file preview starts listing the folder's contents again so takes another minute 

**Additional context**

- using docker container on synology nas 
- every other operation and feature is working perfectly and very performantly

**How to reproduce?**

- spin up docker container 
- mount to /srv a folder containing subfolders containing a large number of files (several thousand) 
- log in and navigate to folder containing large number of files 
- observe the slowness 
- try to share a single file halfway down the list (requires searching) 

**Files**

menasamuel:
Suggest the following:
- a local cache file which can be stored in the folder used for config and can be invalidated and rebuilt as needed by admins
- allowing sharing from the search results screen

If the two above are implemented, this would be super ideal!

niubility000:
1. Use "--disable-type-detection-by-header", 
for example:  .\filebrowser.exe -d .\filebrowser.db --disable-preview-resize --disable-type-detection-by-header --cache-dir .\cache
No matter how large your folder is, it will show up in 2 seconds.
2. This current version is hard to scroll down fast.  You can test my modified version:   https://gitee.com/niubilitypdj/filebrowser/releases/, it's using virtual listing. It can scroll down to anyplace in a second.    If necessary, I can make a PR when I have time. 
3. I think  we need to add a "share" button in the preview page.

menasamuel:
Thanks @niubility000 
I added those params to my docker run command and managed to improve the performance of listing files greatly. I also mounted the cache directory so I can see the cache being populated which is great.

If the share button in the preview page is implemented, then there's no need to scroll fast as it can be done via the search feature which feels more natural than scrolling to me. Perhaps this issue should be re-opened as a feature request. I will mark this bug as closed and create a new feature for the share button to the preview page.

xingoxu:
I'm sorry but is this still valid in v2.23.0？
[...]

PR #3023, opened 2024-02-27
fix: change two default values

niubility000:
There are two default values that cause some problems and always confuse new users again and again, 
1. The default IP address is 127.0.0.1. This only allow users to visit from 127.0.0.1:8080 or localhost:8080. They can't visit from 192.168.1.x:8080 or from public website. #3008, #2584
So, I change this default IP address to 0.0.0.0. Then we can visit from 127.0.0.1:8080 or localhost:8080, or 192.168.1.x:8080, or public website.

2.  The default value of "--disable-type-detection-by-header" is "false". The users may take 30 seconds to wait for the showing of a list with thousands of files, and sometimes may get an error code, especially in low-end server. #1836, #1689, 
So, I change this to "--enable-type-detection-by-header" "false".  By default we don't need it to run filebrowser. 

And also, if a filebrowser.db does not exist in current cmd root folder, it will iniatial a new one automatically in this cmd root folder without warning. It may confuse the users and they may use a wrong database.  After the modification, the abs path of database will show in the cmd window.

screenshot
BTW, we use localhost instead of "0.0.0.0", because "0.0.0.0:8080" looks so weird, and can't be visited directly.

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -54,7 +54,7 @@ func init() {
 }
 
 func addServerFlags(flags *pflag.FlagSet) {
-	flags.StringP("address", "a", "127.0.0.1", "address to listen on")
+	flags.StringP("address", "a", "0.0.0.0", "address to listen on")
 	flags.StringP("log", "l", "stdout", "log output")
 	flags.StringP("port", "p", "8080", "port to listen on")
 	flags.StringP("cert", "t", "", "tls certificate")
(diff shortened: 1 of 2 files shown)


cmd/root.go
```diff
@@ -181,7 +181,11 @@ user created with the credentials from options "username" and "password".`,
 
 		defer listener.Close()
 
-		log.Println("Listening on", listener.Addr().String())
+		if strings.Contains(adr, "0.0.0.0") {
+			log.Println("Listening on localhost:" + server.Port)
+		} else {
+			log.Println("Listening on", listener.Addr().String())
+		}
```
Nerahikada, line 188: 0.0.0.0 != localhost. This is a misleading expression and should remain as before.

Nerahikada:
This includes multiple changes. To facilitate review, it is recommended that you separate the PR for each change.

niubility000:
> This includes multiple changes. To facilitate review, it is recommended that you separate the PR for each change.

Sure, I will separate them to three simple PRs. But then I can commit them at the same time, since they are changing the same file. If one is merged, another one will be out of date. So I will commit them one after one.

cmd/root.go
niubility000: Ok, still show it as 127.0.0.1.

Nerahikada:
If a PR's files become outdated, you can manually merge those changes. I think it is better to open 3 PRs at the same time than to commit them after they have been merged.

PR #3026, opened 2024-02-27
fix: changing the default IP address to 0.0.0.0 root.go

niubility000:
This is an update to #3023

The default IP address is 127.0.0.1. It always cause some problems when new users running filebrowser. This "127.0.0.1" only allow users to visit from 127.0.0.1:8080 or localhost:8080. They can't visit from 192.168.1.x:8080 or from public website. https://github.com/filebrowser/filebrowser/issues/3008, https://github.com/filebrowser/filebrowser/issues/2584

So, I change this default IP address to 0.0.0.0. Then we can visit from 127.0.0.1:8080 or localhost:8080, or 192.168.1.x:8080, or public website.

Since 0.0.0.0:8080 is weird, and can't be visited directly. So in cmd window, we still show it as 127.0.0.1:8080.

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -54,7 +54,7 @@ func init() {
 }
 
 func addServerFlags(flags *pflag.FlagSet) {
-	flags.StringP("address", "a", "127.0.0.1", "address to listen on")
+	flags.StringP("address", "a", "0.0.0.0", "address to listen on")
 	flags.StringP("log", "l", "stdout", "log output")
 	flags.StringP("port", "p", "8080", "port to listen on")
 	flags.StringP("cert", "t", "", "tls certificate")
@@ -181,7 +181,11 @@ user created with the credentials from options "username" and "password".`,
 
 		defer listener.Close()
 
-		log.Println("Listening on", listener.Addr().String())
+		if strings.Contains(adr, "0.0.0.0") {
+			log.Println("Listening on 127.0.0.1:" + server.Port)
+		} else {
+			log.Println("Listening on", listener.Addr().String())
+		}
 		//nolint: gosec
 		if err := http.Serve(listener, handler); err != nil {
 			log.Fatal(err)
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -180,9 +180,8 @@ user created with the credentials from options "username" and "password".`,
 		checkErr(err)
 
 		defer listener.Close()
-
 		if strings.Contains(adr, "0.0.0.0") {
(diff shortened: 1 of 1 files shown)


Nerahikada:
I agree that `0.0.0.0:8080` looks strange to non-technical users, but this should not be changed. Instead, how about displaying a message like "You can access the FileBrowser at `http://<machine local IP>:8080/`"?

niubility000:
> I agree that `0.0.0.0:8080` looks strange to non-technical users, but this should not be changed. Instead, how about displaying a message like "You can access the FileBrowser at http:// < machine local IP > :8080/"?

Do we need to add a function to get the real local ip to replace machine local IP ? 
then in the cmd window, we will show something like:

listening on 0.0.0.0:8080...
You can access the FileBrowser at http://192.168.1.100:8080/

If the user change the ip, we still need to show this real local ip? This may confuse the user. 

I just think 0.0.0.0 includes every ip using in this server. So using localhost:8080, or 127.0.0.1:8080 is fine.

Nerahikada:
I haven't given much thought to the IP of the message to be displayed to non-technical users. Frankly, I think it could be anything.

But my point is that it is 100% wrong to show the message as listening on `127.0.0.1:8080` when it is actually listening on `0.0.0.0:8080`.

niubility000:
> I haven't given much thought to the IP of the message to be displayed to non-technical users. Frankly, I think it could be anything.
> 
> But my point is that it is 100% wrong to show the message as listening on `127.0.0.1:8080` when it is actually listening on `0.0.0.0:8080`.

How about this:
When using the default ip 0.0.0.0, show it like:


When using a ip the user modified, show it like before:

Nerahikada:
~~I think it is good. However, I have a feeling that such a message for beginners would be better written in the documentation.~~
I think a change to 0.0.0.0 is sufficient if only to reduce spam like #3008 and #2584. (I forgot they never read the documentation in the first place)

Nerahikada:
Well, it is much better to have the maintainer review the message to be logged than me.

niubility000:
We may change this "You can access the Filebrowser at" to some official expression. English is not my mother language.

o1egl:
This change might cause security issues for those using it behind the proxy. After the upgrade, they will suddenly expose filebrowser port to the internet.

niubility000:
> This change might cause security issues for those using it behind the proxy. After the upgrade, they will suddenly expose filebrowser port to the internet.

Sure. It's reasonable.

PR #3051, opened 2024-03-15
Feat: stay in the same position of the list when back from a sub folder

niubility000:
This is a new feature #3004, it will work like the Windows Explorer.


niubility000:
Feat: should be feat:
redo it.

PR #3052, opened 2024-03-15
feat: stay in the same position in the list when back from a subfolder

niubility000:
This is a new feature to #3004, it will work like the Windows Explorer.

### DIFF
--- a/frontend/src/components/files/ListingItem.vue
+++ b/frontend/src/components/files/ListingItem.vue
@@ -247,6 +247,10 @@ export default {
       this.addSelected(this.index);
     },
     open: function () {
+      window.sessionStorage.setItem(
+        this.$route.path,
+        JSON.stringify([document.documentElement.scrollTop, this.index])
+      );
       this.$router.push({ path: this.url });
     },
   },
--- a/frontend/src/components/prompts/Rename.vue
+++ b/frontend/src/components/prompts/Rename.vue
@@ -92,7 +92,7 @@ export default {
       try {
         await api.move([{ from: oldLink, to: newLink }]);
         if (!this.isListing) {
-          this.$router.push({ path: newLink });
+          this.$router.replace({ path: newLink });
           return;
         }
 
--- a/frontend/src/views/Files.vue
+++ b/frontend/src/views/Files.vue
@@ -47,6 +47,7 @@ export default {
     return {
       error: null,
       width: window.innerWidth,
+      parentDir: false,
     };
   },
   computed: {
@@ -69,24 +70,22 @@ export default {
   },
   watch: {
     $route: function (to, from) {
+      if (to.path.endsWith("/") && from.path.startsWith(to.path) && from.path.length > to.path.length ){
+        this.parentDir = "true";
+      } else {
+        this.parentDir = "false"; 
+        window.sessionStorage.removeItem(to.path);
+      } 
       if (from.path.endsWith("/")) { 
         if (to.path.endsWith("/"))   {
           window.sessionStorage.setItem("listFrozen", "false");
-          this.fetchData();
-          return;
         } else {
           window.sessionStorage.setItem("listFrozen", "true");
-          this.fetchData();
-          return;        
         }
       } else if (to.path.endsWith("/")) {
         this.$store.commit("updateRequest", {});
-        this.fetchData();
-        return;
-      } else {
-        this.fetchData();
-        return;
       } 
+      this.fetchData();
     }, 
     reload: function (value) {
       if (value === true) {
@@ -116,7 +115,9 @@ export default {
       this.$store.commit("closeHovers");
 
       // Set loading to true and reset the error.
-      if (window.sessionStorage.getItem("listFrozen") !=="true" && window.sessionStorage.getItem("modified") !=="true"){ 
+      if (window.sessionStorage.getItem("listFrozen") !=="true" 
+          && window.sessionStorage.getItem("modified") !=="true" 
+          && this.parentDir !== "true"){ 
         this.setLoading(true);
       }
       this.error = null;
--- a/frontend/src/views/Files.vue
+++ b/frontend/src/views/Files.vue
@@ -1,6 +1,6 @@
 <template>
   <div>
-    <header-bar v-if="error || req.type == null" showMenu showLogo />
+    <header-bar v-if="error || req.type == undefined" showMenu showLogo />
 
     <breadcrumbs base="/files" />
     <listing />
--- a/frontend/src/views/Files.vue
+++ b/frontend/src/views/Files.vue
@@ -70,23 +70,27 @@ export default {
   },
   watch: {
     $route: function (to, from) {
-      if (to.path.endsWith("/") && from.path.startsWith(to.path) && from.path.length > to.path.length ){
+      if (
+        to.path.endsWith("/") &&
+        from.path.startsWith(to.path) &&
+        from.path.length > to.path.length
+      ) {
         this.parentDir = "true";
       } else {
-        this.parentDir = "false"; 
+        this.parentDir = "false";
         window.sessionStorage.removeItem(to.path);
-      } 
-      if (from.path.endsWith("/")) { 
-        if (to.path.endsWith("/"))   {
+      }
+      if (from.path.endsWith("/")) {
+        if (to.path.endsWith("/")) {
           window.sessionStorage.setItem("listFrozen", "false");
         } else {
           window.sessionStorage.setItem("listFrozen", "true");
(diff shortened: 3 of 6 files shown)


frontend/src/components/files/ListingItem.vue
```diff
@@ -247,6 +247,10 @@ export default {
       this.addSelected(this.index);
     },
     open: function () {
+      window.sessionStorage.setItem(
+        this.$route.path,
+        JSON.stringify([document.documentElement.scrollTop, this.index])
+      );
```
niubility000, line 253: This is the only right place to save the scrolltop and index to window.sessionStorage before we leave the current list.

frontend/src/components/prompts/Rename.vue
```diff
@@ -92,7 +92,7 @@ export default {
       try {
         await api.move([{ from: oldLink, to: newLink }]);
         if (!this.isListing) {
-          this.$router.push({ path: newLink });
+          this.$router.replace({ path: newLink });
```
niubility000, line 95: It's a fix for the “back” button function. If we use push. the route link with the old name still in the route history. and when we back to it, an error will show. So we need to change it to replace.

frontend/src/views/Files.vue
```diff
@@ -69,24 +70,22 @@ export default {
   },
   watch: {
     $route: function (to, from) {
+      if (to.path.endsWith("/") && from.path.startsWith(to.path) && from.path.length > to.path.length ){
+        this.parentDir = "true";
+      } else {
+        this.parentDir = "false"; 
+        window.sessionStorage.removeItem(to.path);
+      }
```
niubility000: We only need to keep the position when we're back from a subfolder. This is for it.

frontend/src/views/files/Editor.vue
```diff
@@ -148,8 +147,7 @@ export default {
 
       this.$store.commit("updateRequest", {});
 
-      let uri = url.removeLastDir(this.$route.path) + "/";
-      this.$router.push({ path: uri });
+      history.back();
```
niubility000, line 146: this.$router.push({ path: uri }) will add a new route link to the route history. We should use history.back() here. Then when we click the "back" button in the browser, we will not revisit this .txt file. the windows explorer works in the same way.

frontend/src/views/files/Listing.vue
```diff
@@ -385,7 +385,7 @@ export default {
     req: function () {
       if (window.sessionStorage.getItem("listFrozen") !=="true" && window.sessionStorage.getItem("modified") !=="true"){
         // Reset the show value
-        this.showLimit = 50;
+        this.showLimit = this.req.numDirs + this.req.numFiles;
```
niubility000, line 391: showLimit here will not affect the real showLimit when the page mounts. But this showLimit here will reset the scrollHeight, and allow the browser to scoll down enough to get to the position we need.

frontend/src/views/files/Listing.vue
```diff
@@ -394,6 +394,11 @@ export default {
 
           // Fill and fit the window with listing items
           this.fillWindow(true);
+          if (window.sessionStorage.getItem(this.$route.path)) {
+            this.addSelected(JSON.parse(window.sessionStorage.getItem(this.$route.path))[1]);
+            document.documentElement.scrollTop = JSON.parse(window.sessionStorage.getItem(this.$route.path))[0];
+	    window.sessionStorage.removeItem(this.$route.path);
+          }
```
niubility000, line 409: Here is the only good place to set the scrolltop and selected item to the saved values.

frontend/src/views/files/Listing.vue
```diff
@@ -882,9 +887,9 @@ export default {
       const windowHeight = window.innerHeight;
 
       // Quantity of items needed to fill 2x of the window height
-      const showQuantity = Math.ceil(
-        (windowHeight + windowHeight * 2) / this.itemWeight
-      );
+      const showQuantity = window.sessionStorage.getItem(this.$route.path) ?
+        Math.ceil((JSON.parse(window.sessionStorage.getItem(this.$route.path))[0] + windowHeight * 2) / this.itemWeight) :
+        Math.ceil((windowHeight + windowHeight * 2) / this.itemWeight);
```
niubility000: This is important. Without it, the scrollTop fucntion can reach to the desired position in Chrome, with a obvious scrolling process. and in firefox and safari, the scrollTop function can't reach to the desired position because it runs very fast in firefox and can only invoke the scrollevent() once or twice. To solve this, some coders may save this page's whole vue instance to window.sessionstorage, it's too complicated.

With this modification, the scrolltop works perfectly. it can reach to the desired position immidiately without invoking the scrollEvent().

frontend/src/views/files/Preview.vue
```diff
@@ -346,8 +346,7 @@ export default {
     close() {
       this.$store.commit("updateRequest", {});
 
-      let uri = url.removeLastDir(this.$route.path) + "/";
-      this.$router.push({ path: uri });
+      history.back();
```
niubility000: We should use history.back() here. Then when we click the "back" button in the browser, we will not revisit all the files we previewed. The windows explorer works in the same way.

frontend/src/views/files/Listing.vue
```diff
@@ -394,6 +394,11 @@ export default {
 
           // Fill and fit the window with listing items
           this.fillWindow(true);
+          if (window.sessionStorage.getItem(this.$route.path)) {
+            document.documentElement.scrollTop = JSON.parse(window.sessionStorage.getItem(this.$route.path))[0];
+            if(!this.isMobile)  this.addSelected(JSON.parse(window.sessionStorage.getItem(this.$route.path))[1]);
```
niubility000: In iphone, when highlighting the item, we can see an obvious flash. It's a bug in iphone browser.  So just disable this highlight function in phone.

o1egl:
What is the actual difference in the ui behaviour?

niubility000:
> What is the actual difference in the ui behaviour?

Do you mean this one "In iphone, when highlighting the item, we can see an obvious flash. It's a bug in iphone browser. So just disable this highlight function in phone."

The safari or edge in iphone caches the old page incorrectly before leaving it because safari runs too fast.
For example, Folder1 has two subfolder11, subfolder12. We first click the subfolder11 to enter it, and then return back to folder1, the subfolder11 will be selected(**a little bit slower than the edge in PC**).  then, we click subfolder12 to enter it, and return back to folder1, we can see at first the subfolder11 is selected and then the blue box  jump to subfolder12. This is the different behaviour from the edge in PC. **Is this OK? We can see this obvious jump**. Maybe it looks weird for the users.

if we add two lines in listingItem.vue
    open: function () {
      **this.resetSelected();
      this.addSelected(this.index);**
      window.sessionStorage.setItem(
        this.$route.path,
        JSON.stringify([document.getElementById("listing").scrollTop, this.index])
      );
      this.$router.push({ path: this.url });
    },
When we return to subfolder12, we can see the folder11 is still selected at first. That means safari doesn't run this two lines:       **this.resetSelected();
      this.addSelected(this.index);**

If we add a settimeout  with 100ms
    open: function () {
      **this.resetSelected();
      this.addSelected(this.index);**
      window.sessionStorage.setItem(
        this.$route.path,
        JSON.stringify([document.getElementById("listing").scrollTop, this.index])
      );
      setTimeout(() => {
        this.$router.push({ path: this.url });
      }, 100);
    },
Then, the safari works ok (but still a little bit slow). 
So, that means the safari or edge in iphone runs too fast to execute this two lines **this.resetSelected();    this.addSelected(this.index);**. 

Another reason I don't wanna highlight the item in phone is that the operation bar shows up in the small screen and it looks a little bit messy.   It's not intended to show. The users just wanna browse the list when they return back from a subfolder. 

aa

If all these are acceptable, we can change it to:
    open: function () {
      **this.resetSelected();
      this.addSelected(this.index);**
      window.sessionStorage.setItem(
        this.$route.path,
        JSON.stringify([document.getElementById("listing").scrollTop, this.index])
      );
      setTimeout(() => {
        this.$router.push({ path: this.url });
      }, 100);
    },

niubility000:
To keep the screen clean and no weird behaviour(I have't seen this kind of jump in other projects),  just disable this highlight function in mobile phone.  

By the way, maybe someone doesnot read the updated #3004. To test the effect, we need to a list with more than 50 folders, then click a folder(a) in the bottom, and click back button in the browser to return to the list, we will see the position the folder(a) is  the same and the folder(a) is hightlighted (only in pc).

alanmilinovic:
This doesn't work anymore on latest version.

PR #3057, opened 2024-03-19
Added ability to share folder for view rendering

jsheely:
Myself and a number of issues have been raised requesting the ability to have html content that can be served and rendered instead of only downloaded.

This PR is starting that conversation and adds a functional but crude implementation of the designed feature

- A folder has a new share icon that allows for sharing a "view" folder. Any file in that folder is now viewable
- Adds new /view route to router to handle this. Simply removes the Content-Disposition header conditionally

A user could now share a directory that contains a HTML, Vue or React for website for example directly from within FileBrowser

**Current limitations**
- Only available directly on sharing a folder
- Share icon not added everywhere you might want to share a folder/file directly for rendering

Looking for feedback before completion.

### DIFF
--- a/frontend/src/api/share.js
+++ b/frontend/src/api/share.js
@@ -34,3 +34,6 @@ export async function create(url, password = "", expires = "", unit = "hours") {
 export function getShareURL(share) {
   return createURL("share/" + share.hash, {}, false);
 }
+export function getViewURL(share) {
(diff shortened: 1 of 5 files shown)


jsheely:
@o1egl Looking for feedback if this is something you're interested in bringing into the project. I can scaffold out the rest of the feature then.

PR #3059, opened 2024-03-20
feat: Add open in new tab link

obourgain:
**Description**
For each file in the file list, add a link to open it in a new tab.
This is useful when you have a folder with a lot of files, as otherwise when hitting the back button, you may have to scroll the file list again.

My current code doesn't display a pretty button or anything, but it works for me. Would you open to integrate a feature like that?

- [x] AVOID breaking the continuous integration build. / I ran `make test frontend`

**Further comments**

Screenshot of filebrowser serving itself with this PR:
image

### DIFF
--- a/frontend/src/components/files/ListingItem.vue
+++ b/frontend/src/components/files/ListingItem.vue
@@ -30,6 +30,10 @@
       <p class="modified">
         <time :datetime="modified">{{ humanTime() }}</time>
       </p>
+
+      <a class="" target=”_blank” @click="openInNewTab">
+        open in new tab
+      </a>
     </div>
   </div>
 </template>
@@ -248,6 +252,10 @@ export default {
     },
     open: function () {
       this.$router.push({ path: this.url });
+    },
+    openInNewTab: function () {
+      const routeUrl = this.$router.resolve({path: this.url});
+      window.open(routeUrl.href, '_blank');
     },
   },
 };


niubility000:
It's already fixed in #3004 and #3052 in a common way. you can test it here:https://github.com/niubility000/filebrowser/archive/refs/heads/feat-005-stay-in-the-same-position-of-the-list-when-back-from-a-sub-folder.zip
Your solution is so simple and smart. It will open a new tab everytime when you click the item's "open in new tab" in the list. But the UI with"open in new tab" in every item looks a little bit weird, and it doesn't work in the simple list mode, and in iphone the behaviour is a little bit weird.

obourgain:
I do agree that the layout is not good, especially in list mode.
Also, it could be an icon instead of text, e.g.  in this style [https://thenounproject.com/browse/icons/term/new-tab/](https://thenounproject.com/browse/icons/term/new-tab/) (I didn't check the licence for those, it's just for example).

I don't understand "it doesn't work in the simple list mode". It doesn't look pretty, but the link is there and opens in a new tab. The layout issue is fixable, and an icon could be prettier, but I first wanted to first know if you are interested in this feature or not.
image

What is weird with iPhone? I tried on my android, and it opens a new tab.

Another problem is the "Use single clicks to open files and directories". If this is enabled, clicking the "open in new tab" also triggers the "onClick" of the item.

tox2ik:
This is a terrible way to add this standard behavior (bear with me, i'm not criticizing _you_): 

- It clutters the UI and looks bad.
- Where else have you seen a link like this - ask yourself? 
- You can open in a new tab by holding ctrl and clicking a link in many other web-apps.
- You can open in a new tab with "right+click -> open in new tab" (right click = long press on android)

OR...
if you can't ctrl+click it's because the name of the folder is not an anchor, as it should be. So better fix that.

Right now there is a `div[role=button]` type of element that serves as the clickable part. This is "bad" (but useful because the whole area is clickable);

To make this semantically and functionally correct there should be an anchor around the name of the folder `<a href=/files/folder-33>folder33</a>` instead of the current `<p class=name>{{name}}<a/>`

At the very least, what is proposed in this PR, I think should be an opt-in, default-off feature via user-profile settings, but hopefully you agree that the `<a>{{name}}</a>` solution is the better way to solve this. I mean without introducing a new button that clutters the UI and without doing something that no other browser known to man has ever done.

So overall, I do see the utility in being able to open in a new tab, but the current solution is not good. I have proposed an alternative approach that is the standard way to solve this in HTML, and would be supported on mobile via "long press" or what have you...

Gdakgdak22 (approved):

o1egl:
Thank you for this proposal, but from the ui/ux perspective this implementation is not good. I agree with @tox2ik . I think it would be beter if we simply use the native right-click popup window
Screenshot 2024-04-25 at 01 34 07

kloon15:
U can just simply use middle click to open a link in new tab.

zhaozhixu:
> This is a terrible way to add this standard behavior (bear with me, i'm not criticizing _you_):
> 
>     * It clutters the UI and looks bad.
> 
>     * Where else have you seen a link like this - ask yourself?
> 
>     * You can open in a new tab by holding ctrl and clicking a link in many other web-apps.
> 
>     * You can open in a new tab with "right+click -> open in new tab" (right click = long press on android)
> 
> 
> OR... if you can't ctrl+click it's because the name of the folder is not an anchor, as it should be. So better fix that.
> 
> Right now there is a `div[role=button]` type of element that serves as the clickable part. This is "bad" (but useful because the whole area is clickable);
> 
> To make this semantically and functionally correct there should be an anchor around the name of the folder `<a href=/files/folder-33>folder33</a>` instead of the current `<p class=name>{{name}}<a/>`
> 
> At the very least, what is proposed in this PR, I think should be an opt-in, default-off feature via user-profile settings, but hopefully you agree that the `<a>{{name}}</a>` solution is the better way to solve this. I mean without introducing a new button that clutters the UI and without doing something that no other browser known to man has ever done.
> 
> So overall, I do see the utility in being able to open in a new tab, but the current solution is not good. I have proposed an alternative approach that is the standard way to solve this in HTML, and would be supported on mobile via "long press" or what have you...

I think the ctrl+click cannot be used as "open link in new tab" because it's already used as "select multi clicked files". 

cc @o1egl

PR #3064, opened 2024-03-23
feat: add audio list play in shared folder

niubility000:
I just find after closing the phone's screen, the browser can still keep playing the mp3 list. So I add this function in the shared folder.  Then not only the users but also the visitors can use filebrowser as an audio player. It's cool.
(I also formated the file using latest .eslintrc.json)

to play the current song one time, and click prev, next button to switch songs:

 to play the current song repeatly:

to play all the audio files in current folder repeatly:

### DIFF
--- a/frontend/src/views/Share.vue
+++ b/frontend/src/views/Share.vue
@@ -29,7 +29,7 @@
     <breadcrumbs :base="'/share/' + hash" />
 
     <div v-if="loading">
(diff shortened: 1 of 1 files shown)


o1egl:
Thanks you this suggestion, but it looks a bit strange from the ui perspective. If we want to make those type of changes, I'd go for a totally custom player ui instead of a mix

PR #3065, opened 2024-03-23
feat: remain in the same position when back from a subfolder

niubility000:
This is a update of #3052
When you return back from a subfolder, you will stay in the last position of the list. And the dir you entered before will be highlighted in PC. 
 

(I also formated the file using latest .eslintrc.json)

By the way, the commit message's format is so strict. The subject should start with fix:, and fix: can't be Fix:, and the subject should not end with a ".",  and the optional box should keep empty.  
Otherwise, we need to re-pull the whole request.

### DIFF
--- a/frontend/src/components/files/ListingItem.vue
+++ b/frontend/src/components/files/ListingItem.vue
@@ -247,6 +247,10 @@ export default {
       this.addSelected(this.index);
     },
     open: function () {
+      window.sessionStorage.setItem(
+        this.$route.path,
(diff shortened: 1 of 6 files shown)


alanmilinovic:
Using this for a while and it works perfectly.

alanmilinovic:
Can this be merged and pushed to the next release please?

o1egl:
There are some merge conflicts

alanmilinovic:
@niubility000 can you please resolve merge conflicts?

PR #3081, opened 2024-04-01
add button to go up to .. (parent)

tox2ik:
**Description**

This adds a button to the static set of buttons in the upper-right corner. When clicked, the browser navigates one level up the folder hierarchy. I think this is a crucial input/action to have in a browser.

I would prefer to have it closer to the logo (on the left side), but people probably put branding there, or whatever, so I decided to just stick it with the other buttons.

This is a new feature. No issues exist for this as far as I am aware.

- [?] AVOID breaking the continuous integration build.

```
$make build (…)
found 0 vulnerabilities (…)
✓ 188 modules transformed. (…)
✓ built in 20.20s 
```

**Further comments**

<small>I'm not sure if the last two check can be ticked off, yet. I'll update the PR when I am.</small>

...How do I check the last point? I guess you don't want to use up all the free time on the builder. Not sure how to proceed here.

I have a couple of questions:

- am I supposed to add the new "gettext" localization strings to all the other language file or just the one that I'm using? Is there some automation in place to help updating those files?

And some questions not related to this PR

- do you prefer people ask before adding features or is opening a PR the way to do it?

- how come there is no video preview in the browser? Is that something that you would find useful? (I imagine it could be done with ffmpeg or even directly in go)

- how come the "back" button on the mouse does not work to navigate away from photo preview? `<Esc>` works though, but I think the back button should close the photo preview and navigate back to the file listing just the way `<Esc>` does.

### DIFF
--- a/frontend/src/views/files/Listing.vue
+++ b/frontend/src/views/files/Listing.vue
@@ -46,7 +46,12 @@
             show="delete"
           />
         </template>
-
+        <action
+          v-if="headerButtons.gotoParent"
+          icon="arrow_upward"
+          :label="$t('buttons.gotoParent')"
+          @action="gotoParent"
+        />
         <action
           v-if="headerButtons.shell"
           icon="code"
@@ -367,6 +372,7 @@ export default {
     },
     headerButtons() {
       return {
+        gotoParent: true,
         upload: this.user.perm.create,
         download: this.user.perm.download,
         shell: this.user.perm.execute && enableExec,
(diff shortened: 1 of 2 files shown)


frontend/src/views/files/Listing.vue
```diff
@@ -367,6 +372,7 @@ export default {
     },
     headerButtons() {
       return {
+        gotoParent: true,
```
tox2ik, line 375: This could be a preference.
Not sure that everyone needs this button, since it hasn't been added already.

I'm willing to add this new preference (user profile), and show the new button only if the preference is enabled. Let me know if you think that's better.

The reason I'm asking about this is because it's probably wise to keep the default UI to a minimum.

PR #3087, opened 2024-04-01
fix: add jq raw output flag (#3085)

### LINKED ISSUE #3085: Healthcheck fails in Docker with default .filebrowser.json: curl: (3) URL rejected: Bad hostname
**Description**

This is a regression caused by #2938. Healthcheck fails due to [.filebrowser.json](https://github.com/filebrowser/filebrowser/blob/master/docker_config.json) having an empty address property. The healthcheck script used to default to localhost, now it defaults to empty string.

**Expected behaviour**

healthcheck script calls /health endpoint

curl fails, script exits with code 1

**Additional context**

I _think_ this can be fixed by setting `address` to "localhost" in [docker_config.json](https://github.com/filebrowser/filebrowser/blob/master/docker_config.json)

**How to reproduce?**

```
docker run --rm -p 8080:80 filebrowser/filebrowser:v2.28.0
```
and wait for healthcheck

### LINKED ISSUE #3092: Traefik proxy not proxying the container after latest update [v2.28.0]
**Description**

After the latest update, Traefik proxy is not proxying the container.

> Traefik is exposing it properly when i rollback to `v2.27.0`

**Expected behavior**
Traefik proxy should detect the container to forward it to a Domain.

When I check from Traefik dashboard, it's not listing the container.

**Additional context**
### Docker Compose
```yml
services:
  traefik:
    container_name: traefik
    image: traefik
    restart: unless-stopped
    networks:
      - proxy # rename this to your custom docker network
    ports:
      - target: 80
        published: 80
        protocol: tcp
        mode: host
      - target: 443
        published: 443
        protocol: tcp
        mode: host
      - 8080:8080 # Dashboard port
    volumes:
      - /var/run/docker.sock:/var/run/docker.sock:ro # ro = read-only access to the docker.sock
      - $DOCKERDIR/traefik/:/etc/traefik/
    environment:
      CF_DNS_API_TOKEN: $CLOUDFLARE_API_TOKEN
      TZ: $TZ
    extra_hosts:
      - "host.docker.internal:host-gateway"
    labels:
      traefik.http.routers.api.rule: Host(`traefik.$DOMAINNAME`) # Define the subdomain for the traefik dashboard.
      traefik.http.routers.api.entryPoints: https # Set the Traefik entry point.
      traefik.http.routers.api.service: api@internal # Enable Traefik API.
      traefik.http.routers.api.middlewares: auth@file
      traefik.enable: true # Enable Traefik reverse proxy for the Traefik dashboard.

  filebrowser:
    image: filebrowser/filebrowser   #v2.27.0
    container_name: filebrowser
    volumes:
      - "/:/srv"
      - $DOCKERDIR/filebrowser/filebrowser.db:/database.db
      - $DOCKERDIR/filebrowser/settings.json:/config/settings.json
    networks:
      - proxy
    ports:
      - "80:80"
    environment:
      - PUID=$PUID
      - PGID=$PGID
    restart: unless-stopped
    labels:
      traefik.enable: true
      traefik.http.routers.filebrowser.rule: Host(`file.$DOMAINNAME`)
      traefik.http.services.filebrowser.loadbalancer.server.port: 80
      traefik.http.routers.filebrowser.entryPoints: https
      traefik.http.routers.filebrowser.middlewares: auth@file
```

**How to reproduce?**
Use the latest version of filebrowser container with traefik proxy.

### LINKED ISSUE #3139: Filebrowser docker container in unhealthy status
**Description**
My FileBrowser Docker container works like a charm. But for several weeks already, the unhealthy flag is active but I don't neither see any error message in the docker logs, nor any misbehaviour.

**Expected behaviour**
Healthy docker container status flag

Unhealthy docker container status flag

**Additional context**
n/A

**How to reproduce?**
Directly after deployment and start

**Files**
log

CorySanin:
**Description**
Use jq's "raw output" flag so it doesn't introduce unwanted double quotes in its output.

Closes #3085, Closes #3092, Closes #3139

### DIFF
--- a/healthcheck.sh
+++ b/healthcheck.sh
@@ -1,5 +1,5 @@
 #!/bin/sh
 PORT=${FB_PORT:-$(jq .port /.filebrowser.json)}
-ADDRESS=${FB_ADDRESS:-$(jq .address /.filebrowser.json)}
+ADDRESS=${FB_ADDRESS:-$(jq -r .address /.filebrowser.json)}
 ADDRESS=${ADDRESS:-localhost}
 curl -f http://$ADDRESS:$PORT/health || exit 1
--- a/healthcheck.sh
+++ b/healthcheck.sh
@@ -1,5 +1,5 @@
 #!/bin/sh
-PORT=${FB_PORT:-$(jq .port /.filebrowser.json)}
+PORT=${FB_PORT:-$(jq -r .port /.filebrowser.json)}
 ADDRESS=${FB_ADDRESS:-$(jq -r .address /.filebrowser.json)}
 ADDRESS=${ADDRESS:-localhost}
 curl -f http://$ADDRESS:$PORT/health || exit 1


healthcheck.sh
```diff
@@ -1,5 +1,5 @@
 #!/bin/sh
 PORT=${FB_PORT:-$(jq .port /.filebrowser.json)}
```
Guiorgy: Would it make any sense to also add the `-r` flag to the port query, just incase the user defined the port in the config file as a string, for example:

```json
{
  "port": "80",
  "baseURL": "",
  "address": "",
  "log": "stdout",
  "database": "/database.db",
  "root": "/srv"
}
```

```shell
# jq .port filebrowser.json
"80"
# jq -r .port filebrowser.json
80
```

CorySanin: I see no harm in doing that. I'll commit that now.

kloon15:
Genuine question, why did we stop using the hardcoded `localhost/health` url? Do ppl actually run this in host mode?
I build my own using custom Dockerfile and just baffled about this.

Guiorgy:
@kloon15 Just look at the PR that did the change in the first place to see the reasoning: #2938

kloon15:
> @kloon15 Just look at the PR that did the change in the first place to see the reasoning: #2938

I dont think u should be changing IP address inside the container, just create a custom network for ur needs in docker.

Guiorgy:
@kloon15 I mean, why are you telling that to **me**? 😅
I just tried to answer your question. I didn't say it's a good reason...

Edit: Also, there will always be somebody that does something that "_should not be done_" for a reason that makes sense to them.

qdm12 (approved):
Thanks for the fix and sorry to have caused the issue!

Also my bad for not documenting more the reason of that change, more than just the error it gave me.

EDIT 2024.04.17: the reason was I needed FB to listen on 10.0.0.38 (so not localhost) so the healthcheck querying it on localhost wouldn't work. If the variable FB_ADDRESS is exposed, the healthcheck should adapt to it, so it needed a fix. Well, yes the fix missed on that jq corner case, my bad on that one.
This listening was needed I think for my host firewall to work correctly to restrict intra containers communication, or for reverse proxy reasons, not 100% sure why anymore but i was definitely needed.

saltydk:
@qdm12 Couldn't you just override the healthcheck in the deployment of your container since it is a runtime option?

qdm12:
It would be nice to merge this PR rather soon, it's only 2 lines 😉 Tagging as a reminder @o1egl thanks 👍 

@saltydk I did override the healthcheck script to test it out, and thought I would contribute with a PR, I didn't foresee the impact with jq not handling empty fields properly. When you say

> Couldn't you just override the healthcheck in the deployment of your container since it is a runtime option?

I could propose the exact same to you today, just override the healthcheck, but it's not really a solution, is it?

saltydk:
> It would be nice to merge this PR rather soon, it's only 2 lines 😉 Tagging as a reminder @o1egl thanks 👍
> 
> @saltydk I did override the healthcheck script to test it out, and thought I would contribute with a PR, I didn't foresee the impact with jq not handling empty fields properly. When you say
> 
> > Couldn't you just override the healthcheck in the deployment of your container since it is a runtime option?
> 
> I could propose the exact same to you today, just override the healthcheck, but it's not really a solution, is it?

A user runtime based override on something fairly niche makes more sense to me than changing the image but they merged it so it is what it is. Was just wondering is all.

Guiorgy:
@saltydk If the change impacts normal operation, then yes, but this was supposed to be a change that would benefit some without downsides to the rest. Being more accessible and easy to use for more people is always good. We are all humans, we make mistakes, no need to [make a mountain out of a molehill](https://en.wikipedia.org/wiki/Make_a_mountain_out_of_a_molehill).

o1egl:
Merged as part of the https://github.com/filebrowser/filebrowser/pull/3130

PR #3088, opened 2024-04-02
Focus inputs of promps: New file, New dir, Rename

tox2ik:
- feat: add compact listing mode
- fix: focus reanme and new dir,file inputs

**Description**

The current v-focus attribute was not doing anything.
I'm not sure if the on-mount-next-tick > setTimout is the standard way to do this, but it gets the job done.

From what I can tell, v-focus is not a standard feature and I don't see anything thats handling this custom directive.


tox2ik:
never mind... new version fixed it.

PR #3089, opened 2024-04-02
feat: add button to navigate up to .. (parent dir)

tox2ik:
**Description**

This adds a button to the static set of buttons in the upper-right corner. When clicked, the browser navigates one level up the folder hierarchy. I think this is a crucial input/action to have in a browser.

I would prefer to have it closer to the logo (on the left side), but people probably put branding there, or whatever, so I decided to just stick it with the other buttons.

This is a new feature. No issues exist for this as far as I am aware.

- [?] AVOID breaking the continuous integration build.

```
$make build (…)
found 0 vulnerabilities (…)
✓ 188 modules transformed. (…)
✓ built in 20.20s 
```

**Further comments**

<small>I'm not sure if the last two check can be ticked off, yet. I'll update the PR when I am.</small>

...How do I check the last point? I guess you don't want to use up all the free time on the builder. Not sure how to proceed here.

I have a couple of questions:

- am I supposed to add the new "gettext" localization strings to all the other language file or just the one that I'm using? Is there some automation in place to help updating those files?

### DIFF
--- a/frontend/src/i18n/en.json
+++ b/frontend/src/i18n/en.json
@@ -35,6 +35,7 @@
     "selectMultiple": "Select multiple",
     "share": "Share",
     "shell": "Toggle shell",
+    "gotoParent": "Open parent dir",
     "submit": "Submit",
     "switchView": "Switch view",
     "toggleSidebar": "Toggle sidebar",
--- a/frontend/src/views/files/FileListing.vue
+++ b/frontend/src/views/files/FileListing.vue
@@ -48,6 +48,12 @@
         </template>
 
         <action
+          v-if="headerButtons.gotoParent || true"
+          icon="arrow_upward"
+          :label="t('buttons.gotoParent')"
+          @action="gotoParent"
+        />
+        <action
           v-if="headerButtons.shell"
           icon="code"
           :label="t('buttons.shell')"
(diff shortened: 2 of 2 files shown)


o1egl:
Thank you for your message. Our UI already includes parent directory navigation. Can you please explain how this button will be more useful?
Screenshot 2024-04-24 at 23 40 00

tox2ik:
The breadcrumbs are not visible when you scroll down. This button is more useful because it's possible to navigate back/up to the parent without scrolling back to the top of the file listing.

tox2ik:
The UI element (breadcrumbs) you referred to is not visible when a folder has more than a handful of files. This image illustrates this behavior.

so much scrolling

As an alternative solution, the breadcrumbs div could stick to the static "header" with all the other buttons via static positioning in css, but that is a more complicated solution then just adding an "up" button. Alternatively the layout could be re-arranged to make the breadcrumbs appear on a separate line below the header.

File managers that do not have this "up" button do tend to have breadcrumbs, but then, it is always visible like in KDE Dolphin. On the other hand, some file managers do have this button like Windows Explorer. I'm not really in favor of any approach, but I do miss the ability to go to the parent folder from the "bottom" of a folder without scrolling back up or editing  the URL.

Are you concerned about the buttons section containing more than one arrow symbol? Do you want to avoid many buttons? Do you not care that the breadcrumbs disappear when scrolling?

Please share your thoughts on the proposal to make the breadcrumbs stick to a fixed position when scrolling or re-arranging the layout.
Thanks

PR #3079, opened 2024-03-29
Display Upload Progress as Percentage and File Size / Total File Size

alexjyong:
**Description**
image

EDIT: Closing this PR out for [https://github.com/filebrowser/filebrowser/pull/3111](3111), which takes advantage of Vue 3 and is a bit more simple in implementation. )

Uploads will now show progress as percentage and file size /total file size of all files being uploaded.

I noticed this was missing when I was uploading files to my server. Figured it would be useful for determining how much is left beyond just an ETA.

Didn't see any issues related to this however.

- [x] DO make sure you are requesting to **pull a topic/feature/bugfix branch** (right side). Don't request your master!
- [x] DO make sure you are making a pull request against the **master branch** (left side). Also you should start *your branch* off *our master*.
- [x] DO make sure that File Browser can be successfully built. See [builds](https://github.com/filebrowser/community/blob/master/builds.md) and [development](https://github.com/filebrowser/community/blob/master/development.md).
- [x] AVOID breaking the continuous integration build.

PR #3111, opened 2024-04-11
feat: Display Upload Progress as Percentage and File Size / Total File Size (Take Two)

alexjyong:
**Description**
preview of upload info panel

(Re-do of [PR 3079](https://github.com/filebrowser/filebrowser/pull/3079). Between a few glitches that I found combined with the migration to Vue 3 and all the conflicts it made, I figured it would be easier to just spin up a new PR.)

Uploads will now show progress as percentage and file size /total file size of all files being uploaded.

I noticed this was missing when I was uploading files to my server. Figured it would be useful for determining how much is left beyond just an ETA.

Didn't see any issues related to this however.


maeryo:
cool!
does this feature work properly when uploading multiple files or directories?

alexjyong:
@maeryo  Yep!
Multiple files

image

image


Multiple directories (i do find it odd that it doesn't show the directory name being uploaded as well)

image

frontend/src/stores/upload.ts
```diff
@@ -47,6 +47,41 @@ export const useUploadStore = defineStore("upload", {
...
     },
+    getProgressDecimal: (state) => {
+      if (state.progress.length === 0) {
+        return 0;
+      }
+
+      const totalSize = state.sizes.reduce((a, b) => a + b, 0);
+
+      // TODO: this looks ugly but it works with ts now
+      const sum = state.progress.reduce((acc, val) => +acc + +val) as number;
```
o1egl, line 67: Why `+` sign is used before `acc` and `val`?

frontend/src/stores/upload.ts
```diff
@@ -47,6 +47,41 @@ export const useUploadStore = defineStore("upload", {
...
+    },
+    getTotalSize: (state) => {
+      if (state.sizes.length === 0) {
+        return '0 Bytes';
+      }
+    
+      const totalSize = state.sizes.reduce((a, b) => a + b, 0);
+      const sizes = ['Bytes', 'KB', 'MB', 'GB', 'TB', 'PB', 'EB', 'ZB', 'YB'];
+      const i = Math.floor(Math.log(totalSize) / Math.log(1024));
+      return parseFloat((totalSize / Math.pow(1024, i)).toFixed(2)) + ' ' + sizes[i];
```
o1egl: This code segment is duplicated and can be extracted into a common function

frontend/src/stores/upload.ts
alexjyong: @kloon15 implemented the original logic and could probably explain better than I could, but that's a use of the [unary operator.](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Unary_plus)

Basically, the + sign in front of the acc and val turns them into numbers, and ensures that we are adding numbers and not concatenating strings.

frontend/src/stores/upload.ts
alexjyong: I'll take a look at that. Thanks!

frontend/src/stores/upload.ts
kloon15: Pretty much what @alexjyong said; the problem is that boolean is also assigned to the same variable, which im not sure why and how its used elsewhere and to not break anything while converting to ts, it was just kept as is hence the explicit casting to number.

kloon15:
Is it just me, or if ur uploading more than one file simultaneously, the download speed is wrong? 1 file displays 9mb/s while 5 files shows 1.5mb/s or something like that.
Looks like its divided by the number of current uploads.

alexjyong:
@kloon15  I'll try to reproduce that and if I can, I'll patch it up.  Good catch!

kloon15:
And if u could format ur code with `npm run format` that would be great.😃

frontend/src/stores/upload.ts
alexjyong: Went and implemented this! https://github.com/filebrowser/filebrowser/pull/3111/commits/d29dee3a5e1eb83fb139ffef774e9bb84a6af0ff

alexjyong:
@kloon15  Ahhhhhhhhh, I committed a few seconds too late, I just ran the `prettier` job. :P 
I'm having trouble reproducing the bug you saw. Maybe I'm doing something wrong?

kloon15:
> @kloon15 Ahhhhhhhhh, I committed a few seconds too late, I just ran the `prettier` job. :P I'm having trouble reproducing the bug you saw. Maybe I'm doing something wrong?

No worries, ill double check it tomorrow, maybe it was a one time thing, could have been a network issue or smth.

kloon15:
> @kloon15 Ahhhhhhhhh, I committed a few seconds too late, I just ran the `prettier` job. :P I'm having trouble reproducing the bug you saw. Maybe I'm doing something wrong?

I just tested it again, as soon as i add a second file, the indicated speed drops to about half of what it was before:
image
image

alexjyong:
@kloon15 Thanks! Although it _appears_ this existed before my branch though, since this seems to exist in master without my changes.

Screenshot 2024-04-26 at 2 00 47 PM
Screenshot 2024-04-26 at 2 00 58 PM


In master the speed did eventually go up, although a bit slower than before (which is expected, since I was uploading more files now.)

And I do see the issue in my branch as well. I think this already existed, although I wonder if it's a bug or the upload libraries getting things together? Because it eventually levels out to near what it was.

Screenshot 2024-04-26 at 2 14 01 PM

Screenshot 2024-04-26 at 2 14 16 PM

@kloon15  did you see the same thing with the speed leveling out as well?

kloon15:
> @kloon15 Thanks! Although it _appears_ this existed before my branch though, since this seems to exist in master without my changes.
> Screenshot 2024-04-26 at 2 00 47 PM Screenshot 2024-04-26 at 2 00 58 PM
> 
> In master the speed did eventually go up, although a bit slower than before (which is expected, since I was uploading more files now.)
> 
> And I do see the issue in my branch as well. I think this already existed, although I wonder if it's a bug or the upload libraries getting things together? Because it eventually levels out to near what it was.
> Screenshot 2024-04-26 at 2 14 01 PM Screenshot 2024-04-26 at 2 14 16 PM
> 
> @kloon15 did you see the same thing with the speed leveling out as well?

No it has nothing to do w ur branch or modifications, just brought it up since ur adding/reworking to it and that maybe smth is wrong w it.
Yea the speed started out low and then went up, thats when i took both screenshots, it never leveled out to what its supposed to be. Im gonna also mention that the estimated time is also wrong for me.
Heres a comparison with uploading 5 files with filebrowser and winscp, just compare and its actually the exact same speed for a single file and the time is added up:
image
image

o1egl:
I have observed that when I choose to upload a file that already exists, the "replace" modal window appears briefly and then disappears immediately. However, this behavior is not replicated when I use the drag-and-drop feature. I'm going to merge this PR since this issue is not related to it. Nevertheless, it would be great if one of you (@kloon15, @alexjyong) could take a closer look into this matter.

alexjyong:
@o1egl  I'll check it out if I get some time 🙂 . So it's not forgotten, I've made an issue [here](https://github.com/filebrowser/filebrowser/issues/3173) tracking it.

alexjyong:
@o1egl  haven't forgotten about this. been busy with professional and personal obligations that have come up. Sorry!

krigeta:
Hey, may somebody share the code of how to use the filebrowser on Modal? like a basic script of it?

PR #3122, opened 2024-04-16
Update healthcheck.sh adding -r flag to jq

blinkitogaming:
**Description**

I simply added the -r flag to the jq execution at healthcheck.sh


blinkitogaming:
I saw there's already a pending PR correcting this issue.

PR #3183, opened 2024-05-01
feat: Enhance MIME Type Detection for Additional File Extensions

gdy666:
By the way, before enhancing MIME type detection for additional file extensions, it's surprising that even the MIME type for .flac files cannot be recognized in the Linux arm64 environment.

音乐类型

QQ截图20240501194015


baur:
can you fix this #2603 also 

image

gdy666:
> can you fix this #2603 also
> 
> image

xlsx Already included

gdy666:
> can you fix this #2603 also
> 
> image



> can you fix this #2603 also
> 
> image

The problem in your screenshot has nothing to do with the problem involved in this PR. Your problem involves front-end css style processing. Let me think about whether there is a good way to deal with it for you.

gdy666:
> can you fix this #2603 also
> 
> image

https://github.com/filebrowser/filebrowser/pull/3187

gdy666:
> can you fix this #2603 also
> 
> image
@baur 
You can refer to this
https://github.com/filebrowser/filebrowser/pull/3187

files/mime.go
```diff
@@ -0,0 +1,608 @@
...
+	".sldx":      "application/vnd.openxmlformats-officedocument.presentationml.slide",
+	".sldm":      "application/vnd.ms-powerpoint.slide.macroEnabled.12",
+	".thmx":      "application/vnd.ms-officetheme",
+	".onetoc":    "application/onenote",
+	".onetoc2":   "application/onenote",
+	".onetmp":    "application/onenote",
+	".onepkg":    "application/onenote",
+	".xpi":       "application/x-xpinstall",
+	".wasm":      "application/wasm",
+	//---
```
o1egl: What is that for?

files/mime.go
```diff
@@ -0,0 +1,608 @@
...
+	".opus": "video/ogg",
+	".m4v":  "video/mp4",
+	".mkv":  "video/x-matroska",
+	".caf":  "audio/x-caf",
+	".m3u8": "application/x-mpegURL",
+	".mpd":  "application/dash+xml",
+	".webp": "image/webp",
+}
+
+func init() {
```
o1egl, line 603: Add nolint comment to suppress linter error

files/mime.go
gdy666: sorry,It is only used to distinguish the part from the github.com/kataras/iris. It would be more appropriate to remove this line.

PR #3187, opened 2024-05-02
fix: The file type icon in the file list is sensitive to the case of …

gdy666:
Fix: The file type icon in the file list is sensitive to the case of the suffix name

fix before:
before

fix after:
after


frontend/src/components/files/ListingItem.vue
```diff
@@ -270,4 +271,14 @@ const click = (event: Event | KeyboardEvent) => {
 const open = () => {
   router.push({ path: props.url });
 };
+
+const getExtension = (fileName: string): string => {
+  const lastDotIndex = fileName.lastIndexOf('.');
+  if (lastDotIndex === -1) {
+    return '';
```
o1egl: I think in that case we can return filename itself to support files like Dockerfile, Makefile etc

o1egl (approved):
Small suggestion

frontend/src/components/files/ListingItem.vue
gdy666: Yes, you are right

PR #3204, opened 2024-05-07
feat: make uploadsLimit into  user setting

gdy666:
QQ截图20240507222301

### DIFF
--- a/frontend/src/components/prompts/Upload.vue
+++ b/frontend/src/components/prompts/Upload.vue
@@ -37,11 +37,13 @@ import { useI18n } from "vue-i18n";
 import { useRoute } from "vue-router";
 import { useFileStore } from "@/stores/file";
 import { useLayoutStore } from "@/stores/layout";
+import { useAuthStore } from "@/stores/auth";
 
 import * as upload from "@/utils/upload";
 
 const { t } = useI18n();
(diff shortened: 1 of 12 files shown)


o1egl:
This PR adds only a client-side naive approach, which can be easily overridden. The proper approach requires server-side implementation as well. Why do we need it in general?

gdy666:
> This PR adds only a client-side naive approach, which can be easily overridden. The proper approach requires server-side implementation as well. Why do we need it in general?

The intention of this PR is solely to allow the client to control the number of concurrent file uploads, as in real-world applications, when clients upload files concurrently, upload speed may significantly decrease compared to uploading a single file due to factors such as concurrent disk reads or other performance issues.

gdy666:
> This PR adds only a client-side naive approach, which can be easily overridden. The proper approach requires server-side implementation as well. Why do we need it in general?

By the way, when uploading concurrently, what is displayed here is just the average upload speed. Shouldn't it display the total upload speed instead?

Caian:
I like this idea very much. I usually update large files to my server, so it is better to have a few files ready versus a lot of files on the fly.

Being a quality-of-life feature for a particular user it is not a problem that it is not enforced on the server-side.

Issue #2518, opened 2023-06-26
share optional password autocomplete issue


edwardchenchen:
**Is your feature request related to a problem? Please describe.**

![image](https://github.com/filebrowser/filebrowser/assets/6397192/7a4360f8-7eb9-41b2-b520-e76ad297b154)
currently when create a share, browser will autofill the password with login password

**Describe the solution you'd like**

I think autocomplete can be disabled on this field

**Describe alternatives you've considered**

**Additional context**

edwardchenchen:
unstale

edwardchenchen:
unstale

edwardchenchen:
unstale

alphonse-terrier:
unstale

edwardchenchen:
unstale

Issue #2035, opened 2022-07-11
Prevent share password from autofilling


Lombra:
**Is your feature request related to a problem? Please describe.**
Opening the share dialog autofills the share password using my user account password.

**Describe the solution you'd like**
Do not autofill this input.

**Additional context**
Using Firefox 102.

Lombra:
While obviously less disruptive, one might want to give the password change form the same treatment.

UbioZur:
This is probably your password manager (directly firefox or any other third party ones you may be using) that does the autofilling not filebrowser.

Try to disable the autofill from your password manager and see if it does indeed stop filling the password field automatically.

If that works, then it's a matter of deciding either you want to keep the convenience of autofilling your passwords, but have the inconvenience under filebrowser.
Or have the inconvenience of having to manually get your password manager to fill the fields, but not have the problem under filebrowser (and possibly many other websites)

Lombra:
I'm aware that it is my browser that's doing it, but I thought it might be possible to prevent this behaviour by way of web design. There is, for example, an attribute `autocomplete="new-password"` that seems relevant, though I can't speak to its functionality.

Towdium:
Same here, I think adding `autocomplete="new-password"` does the work. Password manager think the input password is the one to login because it is a form input with type "password", but actually it is not. That makes the browser work in a confusing way, and fill my account password as share password.

PR #3207, opened 2024-05-07
fix: fixed issue with password managers autofill

coolchevy:
**Description**

**Further comments**

Closing issues https://github.com/filebrowser/filebrowser/issues/2518 https://github.com/filebrowser/filebrowser/issues/2035

### DIFF
--- a/frontend/src/components/prompts/Share.vue
+++ b/frontend/src/components/prompts/Share.vue
@@ -107,8 +107,8 @@
         </div>
         <p>{{ $t("prompts.optionalPassword") }}</p>
         <input
-          class="input input--block"
-          type="password"
+          class="input input--block input--secure"
+          type="text"
           v-model.trim="password"
           tabindex="3"
         />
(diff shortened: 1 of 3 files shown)


o1egl:
Some people prefer to disable password manager support on certain fields, while others prefer to have it. The CSS solution provided is[ non-standard](https://developer.mozilla.org/en-US/docs/Web/CSS/-webkit-text-security) and may not be reliable. Personally, I use 1Password and have no issues with autofill. If there is an issue with a specific password management, we might consider adding some ignore data attributes

coolchevy:
@o1egl you need to test how Password Manger works now. It more looks like security issue for beginner PC users when this users input a password for sharing links using their personal password for file browser login.
I can't find an idea how is password manager could be used with sharing files, because if user will accept to save password -  password manager will create new user/password for user login. 
I'm tried to do it with autocomplete="off", but it is not working .

PR #3236, opened 2024-05-19
fix: Localhost should not be set to loopback.

novaTopFlex:
**Description**

- [X] DO make sure you are requesting to **pull a topic/feature/bugfix branch** (right side). Don't request your master!
- [X] DO make sure you are making a pull request against the **master branch** (left side). Also you should start *your branch* off *our master*.

**Further comments**


kloon15:
Do not change the IP in `vite.config.ts`, thats hardcoded for access to the running local server.

Issue #3109, opened 2024-04-11
Disable file preview, just download on file open


JonathanG-DC:
**Is your feature request related to a problem? Please describe.**
I'd like to set a user (or globally) so that the file preview function is completely disabled, instead the file is downloaded when clicked.

**Describe the solution you'd like**
An additional permission for users allowing the use of the preview file function.
If enabled the system acts as normal.
If disabled when the user clicks the file it is automatically downloaded rather than opening the preview.

**Describe alternatives you've considered**
A global flag for disabling the preview function would also work.

**Additional context**
N/A

JonathanG-DC:
Bump

PR #2954, opened 2024-01-22
feat: onlyoffice Integration (cont...)

arcastro:
**Description**

This is a continuation of the pull request here:
* https://github.com/filebrowser/filebrowser/pull/1420

In this pull request, I have:
* Rebased @drosoCode's contributions against the latest master branch
* Addressed the pending comments on that pull request:
  * OnlyOffice's secret key is no longer exposed to the client. Instead all signing occurs server-side.
  * The document key is now generated as a sha256 hash. 
  * "Magic" numbers are assigned meaningful names.
* Additionally, I made a few other improvements:
  * Fixed a bug in the generation and caching of the document key in order to properly support co-editing.
  * Set OnlyOffice dark/light theme according to filebrowser settings.
  * Added mobile support for the OnlyOffice editor.

**Links**
* https://github.com/filebrowser/filebrowser/pull/1420
* #1255 
* #1272 
* #2923

PR #3276, opened 2024-05-31
feat: open item in new window when pressing ctrl/meta key

andresbono:
**Description**

Open item in a new window or tab when pressing <kbd>CTRL</kbd> / <kbd>META</kbd> key.

When doing a selection of items, you can still press <kbd>CTRL</kbd> / <kbd>META</kbd> to (de)select single items.

---

**Further comments**

- ⚠️ I'm not sure if some users can consider this a regression as they may be used to start selecting items pressing  <kbd>CTRL</kbd> / <kbd>META</kbd>. By merging this PR, you would enter "selection mode" with <kbd>SHIFT</kbd> only.

### DIFF
--- a/frontend/src/components/files/ListingItem.vue
+++ b/frontend/src/components/files/ListingItem.vue
@@ -211,12 +211,14 @@ const drop = async (event: Event) => {
 const itemClick = (event: Event | KeyboardEvent) => {
   if (
     singleClick.value &&
-    !(event as KeyboardEvent).ctrlKey &&
-    !(event as KeyboardEvent).metaKey &&
+    !(
+      fileStore.selectedCount !== 0 &&
+      ((event as KeyboardEvent).ctrlKey || (event as KeyboardEvent).metaKey)
+    ) &&
     !(event as KeyboardEvent).shiftKey &&
     !fileStore.multiple
   )
-    open();
+    open(event);
   else click(event);
 };
 
@@ -230,7 +232,7 @@ const click = (event: Event | KeyboardEvent) => {
 
   touches.value++;
   if (touches.value > 1) {
-    open();
+    open(event);
   }
 
   if (fileStore.selected.indexOf(props.index) !== -1) {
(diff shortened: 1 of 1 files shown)


Silloky:
This would be a really good addition.
But would it show the toolbar at the top ?

Just looking ahead with the onlyoffice #2954 integration, where the editor is a bit small with the toolbar, plus it would be nice to keep the editor separate from the main tab.

Silloky:
bump to keep this open

Silloky:
bump

andresbono:
> But would it show the toolbar at the top ?
> 
> Just looking ahead with the onlyoffice #2954 integration, where the editor is a bit small with the toolbar, plus it would be nice to keep the editor separate from the main tab.

Yes, the the toolbar will be present at the top, no change in that behavior. I would say the ability to open items in a new window (this PR) and being able to conditionally show the toolbar at the top should be treated independently, isn't it?

Silloky:
I feel opening in a new tab means separating from the initial Filebrowser.
I don't think the toolbar should be shown when in a new tab : for example, for a PDF, I think it should be the standard PDF viewer, without the added Filebrowser options such as Remove.

> being able to conditionally show the toolbar at the top should be treated independently, isn't it?
I'm not talking about removing the toolbar in the standard interface, just in the new tabs, thus the link with your PR.

andresbono:
Could make sense, yes. I would make it optional, via some configuration option. See #3109. Filtering by file extension would be nice too.

I still don't see why coupling both improvements, and to be honest, I don't have the bandwidth to make the scope of this PR any bigger 🫤.

IMHO, as this PR already delivers some (tiny) enhancement already, I would merge it as-is. Further improvements can be built on top of it.

Silloky:
Sure !
Once this is merged, I'll try implementing what I'm saying.

o1egl:
It doesn't work in chrome on mac. At the moment command key is used to select multiple files. Maybe a workaround would be adding open in a new tab option to this PR https://github.com/filebrowser/filebrowser/pull/3343

andresbono:
Yeah, that was my concern described in "Further comments", some users may see this feature as a regression. Feel free to close the PR if it is not a good fit :) Thanks!

Issue #3242, opened 2024-05-20
save keybind doesnt work if capslock is on.


bathtubfulloftoast:
**Description**
whenever i try to save the file it brings up the save popup dialogue given by firefox

**Expected behaviour**
file is saved properly

**What is happening instead?**
firefox tells me to save the webpage

**Additional context**
i am using firefox
also it only happens whenever caps lock is on

**How to reproduce?**
go on firefox
turn on caps lock
press ctrl + s

Issue #3237, opened 2024-05-19
PDF preview on Android


pierluigizagaria:
**Description**
PDF cannot be opened Android devices.

**Expected behaviour**
View the pdf content like on web desktop.

**What is happening instead?**

<img src="https://github.com/filebrowser/filebrowser/assets/57801386/2a8bb367-77e4-45b9-b9f1-0415d450b871" height="400" />

**Additional context**
Problem already saw on other issue but all closed even tho the problem still persists.

mradalbert:
Duplicate of: https://github.com/filebrowser/filebrowser/issues/1201

pierluigizagaria:
Why is the #1201 closed? The issue still persists and no workaround are found

mradalbert:
Filevrowser doesn't render pdfs and relies on browser to do so. It is browser issue IMO.

pierluigizagaria:
Filebrowser has the preview feature for different kinds of file extensions that does not depend on the device browser. Could we manage pdf in the same way? So previews are always working regardless the browser?
Can we implement that If the browser does not support the file, automatically download it so chrome opens it directly?

Mubelotix:
Still an issue. This stale bot is infuriating

Issue #3230, opened 2024-05-16
Ability to create temporary users


p3t3rp3pp3r:
**Is your feature request related to a problem? Please describe.**
It would be great to add an option that permits to auto delete the user after a specific time (in hours), this feature can provide an integrated solution everytime we need to grant temporary access to filebrowser

qwertysmosh:
Please add this simple function if possible. Hope the devs find time to check on this, thank you so much!

Issue #3213, opened 2024-05-08
Return to same place in directory listing when closing a viewer window


a545254328:
For example, if you open a folder containing many photos, scroll down with your mouse, double-click to view one of the photos, and then close the photo, the web page will automatically jump back to the first photo. I need to scroll down again with my mouse to see the photo at the bottom

https://github.com/filebrowser/filebrowser/assets/51851008/d0be87b2-0679-416c-8308-e9088a45acc7

decaby7e:
I am encountering the same issue.

I believe this is a bug, as #3004 was supposed to have fixed this.

decaby7e:
This issue is still occuring.

Here is a demonstration (I'm running the latest version, v2.30.0): https://github.com/filebrowser/filebrowser/assets/27081624/8f986dc5-d5da-4385-8ae9-2e014881d35c

jniggemann:
Thank you for opening an issue and contributing to the filebrowser project!
I can confirm this issue on the latest version and it's not limited to images.

PR #3337, opened 2024-06-26
ci: Change stale days from 30 to 90

Cretezy:
**Description**

30 days before stale issues are closed is too short for a slow-moving project like Filebrowser. for example:

- https://github.com/filebrowser/filebrowser/issues/3007
- https://github.com/filebrowser/filebrowser/issues/3213
- https://github.com/filebrowser/filebrowser/issues/3230
- https://github.com/filebrowser/filebrowser/issues/3237
- https://github.com/filebrowser/filebrowser/issues/3242

These are just ~30s of looking. [They are >700 issues closed because of staleness.](https://github.com/filebrowser/filebrowser/issues?q=sort%3Aupdated-desc+is%3Aissue+is%3Aclosed+%22This+issue+was+closed+because+it+has+been+stalled+for+5+days+with+no+activity.%22)

I would even encourage to increase this to 180 or 365, or even remove completely. Issues should ideally be closed by a human when they are completed.

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -13,12 +13,12 @@ jobs:
     steps:
       - uses: actions/stale@v5
         with:
-          stale-pr-message: 'This PR is stale because it has been open 30 days with no activity. Remove stale label or comment or this will be closed in 5 days.'
-          close-pr-message: 'This PR was closed because it has been stalled for 5 days with no activity.'
-          stale-issue-message: 'This issue is stale because it has been open 30 days with no activity. Remove stale label or comment or this will be closed in 5 days.'
-          close-issue-message: 'This issue was closed because it has been stalled for 5 days with no activity.'
-          days-before-stale: 30
-          days-before-close: 5
+          stale-pr-message: 'This PR is stale because it has been open 90 days with no activity. Remove stale label or comment or this will be closed in 7 days.'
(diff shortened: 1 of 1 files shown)


o1egl:
I disagree with the statement. If there is no response from the community within 30 days, it implies that the proposal is either not relevant or lacks interest in implementation.

Cretezy:
A bug present now will still be there in 30 days, so why should the issue be closed?

The main purpose of these stale issue closers is to keep the number of open issues down, but this comes at the trade-off of closing more valid issues than invalid.

jniggemann:
I like this proposal very much and have experienced the underlying problem myself. Just today I got a "auto close" message for #3312, which is a reproducible bug with immediately noticeable user impact. 

IMHO the situation is difficult to address, because it can easily be taken for badmouthing the project or its developers, so let me explicitly say that it's neither when I say:

I expect that there are quite some bugs that someone took the time to dutifully report but which then were buried by autoclose.

I understand that the issue tracker gets a lot of tickets and only very few are qualified bug reports, but saying
> If there is no response from the community within 30 days, it implies that the proposal is either not relevant or lacks interest in implementation.

is a fallacy, because 31% of all PRs are closed without being merged. So there may well be community feedback / contribution, but that also gets buried the same way.

jniggemann:
Commenting to keep this open, please consider this change. See my comment above, more than 30% of all PRs are auto-closed...

PR #3494, opened 2024-09-16
style: Fix alignment file margin right

### LINKED ISSUE #3385: Not aligned
2.30 versions 
In list mode, the contents of the size and last modified columns are not aligned with the table header

IMG_2024-08-03_10-07-24

SilverSong3:
Fixes #3385 

**Description**
Fix for #3385 to align the size and modified titles above their respective data. The elements should now be lined up to rest on the same vertical alignment line.

An example of the fixed CSS style change is shown in the picture below:

360562202-bffcf567-4423-43c1-ad58-515ac169ac7d

**Important Note**
Because this is one of my first pull requests, I trust that though some unconventional methods were used the style did apply the fix and confirm the code patches the issue. It was tested using Inspect with Firefox and confirmed with the Style Editor that confirmed the first commit would not solve the issue because it would apply the -3em to the .item .header as well as the .item p element. Thankful to find [this stackoverflow answer](https://stackoverflow.com/a/16202009) to deselect the header for the CSS selector.

**Further comments**
Found 2 lines containing the same #listing.list .item.header .name and deleted them both to remove the offset the title had from the initial commit.


pam-param-pam:
image
im pretty sure this somehow introduced this problem

AsadUmar:
Can we please revert this PR? its creating more issues than its solving.

PR #3537, opened 2024-10-15
fix: Prompts disappearing on copy / move / upload

### LINKED ISSUE #3325: Overwrite dialogue dissapears as of 2.29.0
In 2.29.0 and 2.30.0 the file overwrite dialogue does not stay on the screen.  It appears and vanishes before the user has a chance to interact with it.

Downgrading to 2.28.0 makes it stick again.

### LINKED ISSUE #3519: 🐞 Upload Continue/Replace Dialog auto close with in second without click
**Description**
After failed upload file re-select for upload then show  continue/replace dialog that only display for less than 1 second then immediately closed without click. 
* After downgrade to v2.28.0 from 2.31.1 then working.
*  Not working with versions:  2.31.1,  2.31.0,  2.30.0,  2.29.0

Screenshot_20241002_074658_Chrome
* take more than 7 times for take screenshot because of instantly closed. 

**Expected behaviour**
Dialog should continue show untill select option.

Please solve this problem. 
Thank you,

Rmiller5466:
**Description**
Fixes the bug mentioned in #3519 and #3325 .  

I believe the root of the issues with disappearing modals is in how the `close()` was being handled in Prompts.Vue.  My latest commit removes the `closeModal` reactive var and instead opts for storing the close function directly on the prompt itself.  This now allows for the modals to "stack", and then when `closeHovers` is called, it properly dismisses each one, oldest to newest.  Switching the `closeHovers` array operation from `pop` to `shift` is what ensures all prompts are closed out.  I also removed the `closeHovers` from Upload.vue and FileListing.vue as they would prematurely dismiss any replace/rename prompt if present.  This closeHovers is now in `/utils/upload.ts`, which ensures any prompts are closed only after replace/rename  

closes #3519 
closes #3325

### DIFF
--- a/frontend/src/components/prompts/Prompts.vue
+++ b/frontend/src/components/prompts/Prompts.vue
@@ -30,8 +30,6 @@ const layoutStore = useLayoutStore();
 
 const { currentPromptName } = storeToRefs(layoutStore);
 
-const closeModal = ref<() => Promise<string>>();
-
 const components = new Map<string, any>([
   ["info", Info],
   ["help", Help],
@@ -52,11 +50,6 @@ const components = new Map<string, any>([
 ]);
 
 watch(currentPromptName, (newValue) => {
-  if (closeModal.value) {
-    closeModal.value();
-    closeModal.value = undefined;
-  }
-
   const modal = components.get(newValue!);
   if (!modal) return;
 
@@ -67,7 +60,7 @@ watch(currentPromptName, (newValue) => {
     },
   });
 
-  closeModal.value = close;
+  layoutStore.setCloseOnPrompt(close, newValue!);
   open();
 });
 
--- a/frontend/src/components/prompts/Prompts.vue
+++ b/frontend/src/components/prompts/Prompts.vue
@@ -3,7 +3,7 @@
 </template>
 
 <script setup lang="ts">
-import { ref, watch } from "vue";
+import { watch } from "vue";
 import { ModalsContainer, useModal } from "vue-final-modal";
 import { storeToRefs } from "pinia";
 import { useLayoutStore } from "@/stores/layout";
--- a/frontend/src/stores/layout.ts
+++ b/frontend/src/stores/layout.ts
@@ -29,12 +29,11 @@ export const useLayoutStore = defineStore("layout", {
     toggleShell() {
       this.showShell = !this.showShell;
     },
-    setCloseOnPrompt(closeFunction: any, onPrompt: string) {
-      let prompt = this.prompts.find((prompt) => prompt.prompt === onPrompt);
+    setCloseOnPrompt(closeFunction: () => Promise<string>, onPrompt: string) {
+      const prompt = this.prompts.find((prompt) => prompt.prompt === onPrompt);
       if (prompt) {
         prompt.close = closeFunction;
       }
-      return null;
     },
     showHover(value: PopupProps | string) {
       if (typeof value !== "object") {
--- a/frontend/src/stores/layout.ts
+++ b/frontend/src/stores/layout.ts
@@ -75,7 +75,7 @@ export const useLayoutStore = defineStore("layout", {
       });
     },
     closeHovers() {
-      this.prompts.shift()?.close?.();
+      this.prompts.pop()?.close?.();
     },
     // easily reset state using `$reset`
     clearLayout() {
--- a/frontend/src/stores/layout.ts
+++ b/frontend/src/stores/layout.ts
@@ -29,6 +29,13 @@ export const useLayoutStore = defineStore("layout", {
     toggleShell() {
       this.showShell = !this.showShell;
     },
+    setCloseOnPrompt(closeFunction: any, onPrompt: string) {
+      let prompt = this.prompts.find((prompt) => prompt.prompt === onPrompt);
+      if (prompt) {
+        prompt.close = closeFunction;
+      }
+      return null;
+    },
     showHover(value: PopupProps | string) {
       if (typeof value !== "object") {
         this.prompts.push({
@@ -36,6 +43,7 @@ export const useLayoutStore = defineStore("layout", {
           confirm: null,
           action: undefined,
           props: null,
+          close: null,
         });
         return;
       }
@@ -45,6 +53,7 @@ export const useLayoutStore = defineStore("layout", {
         confirm: value?.confirm,
         action: value?.action,
         props: value?.props,
+        close: value?.close,
       });
     },
     showError() {
@@ -53,6 +62,7 @@ export const useLayoutStore = defineStore("layout", {
         confirm: null,
         action: undefined,
         props: null,
+        close: null,
       });
     },
     showSuccess() {
@@ -61,10 +71,11 @@ export const useLayoutStore = defineStore("layout", {
         confirm: null,
         action: undefined,
         props: null,
+        close: null,
       });
     },
     closeHovers() {
-      this.prompts.shift();
+      this.prompts.shift()?.close?.();
     },
     // easily reset state using `$reset`
     clearLayout() {
--- a/frontend/src/stores/layout.ts
+++ b/frontend/src/stores/layout.ts
@@ -64,7 +64,7 @@ export const useLayoutStore = defineStore("layout", {
       });
     },
     closeHovers() {
-      this.prompts.pop();
+      this.prompts.shift();
     },
     // easily reset state using `$reset`
     clearLayout() {
--- a/frontend/src/stores/layout.ts
+++ b/frontend/src/stores/layout.ts
@@ -75,7 +75,7 @@ export const useLayoutStore = defineStore("layout", {
       });
     },
     closeHovers() {
-      this.prompts.pop()?.close?.();
+      this.prompts.shift()?.close?.();
     },
     // easily reset state using `$reset`
     clearLayout() {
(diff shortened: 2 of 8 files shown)


Rmiller5466:
Sorry for closing and reopening - thought I wouldn't have the time to look into this a bit further for a while, though ended up having a spare hour to dig around a bit more.  Wasn't completely happy with my quick fix, and then noticed a few more scenarios where the modals weren't behaving as expected.

killua99:
this might fix #3325

killua99:
To rmiller, this PR doesn't show a real skip button in the modal where one is copy or moving files, right ?

The skip functionally isn't available ? I just get the option cancel, rename, replace

Rmiller5466:
> To rmiller, this PR doesn't show a real skip button in the modal where one is copy or moving files, right ?
> 
> The skip functionally isn't available ? I just get the option cancel, rename, replace

I'll look into this this weekend.  Was there previously a skip option for copying / moving files?  I only recall seeing the skip option on uploads, though could be misremembering.  

 I think there a few additional outstanding modifications I made that aren't pushed here yet either, though those are just edge cases, not key UI / functionality differences.

Rmiller5466:
Turns out there was just one additional close hovers that I had missed, no other changes.  I think this PR just needs reviewed.  

@killua99 Could you elaborate more on the skip functionality you mentioned?  The only thing similar to this I observed was the "Continue" option when uploading a file that already exists.  I believe the copy/move is an "all or nothing" rather than one by one that the upload does

killua99:
@Rmiller5466 imagine you have a folder A and located in the path  `current/path/A` but there's a new folder and we need to move files that are inside folder A to the new path `new/path/A` inside that folder there was a file that the new path share in name but no content. When one use the feature copy or move the folder A from current to new path the modal ask you if you want to skip it or continue with the options, the options are to cancel, rename, replace. But what I'm looking is the skip this particular file but continue with the operation of moving the rest of the bulk.

Similar like normal desktop file manager does. They ask you to skip any action with that file (or apply the action to the entire bulk with the answer you gave, skip, rename, replace)

Currently the modal just cancel the entire bulk operation

Rmiller5466:
@killua99 Okay yes, I understand what you are saying.  Was this an existing functionality that's been lost, or are you asking for this as a new feature? If the latter, I'd say create a new issue for that, as I'd like to keep the PR as focusing on the bugfix of the prompts disappearing.

killua99:
Absolutely, keep this PR as it. Seems like is a new feature like it wasn't implemented before.

Going to search issue history and open a new one if needed.

o1egl:
@kloon15 would you be able to review this PR?

kloon15:
> @kloon15 would you be able to review this PR?

Sure thing.

kloon15:
> @kloon15 would you be able to review this PR?

I tested and it fixes the issue we had and i didnt notice any regression. 👍

Rmiller5466:
Is there anything else needed on my end?  Or are we just waiting for an official approval at this point?

kloon15:
Nope we just need Oleg to merge the PR.

Ryan ***@***.***> ezt írta (időpont: 2024. dec. 11., Sze
3:48):

frontend/src/stores/layout.ts
```diff
@@ -61,10 +71,11 @@ export const useLayoutStore = defineStore("layout", {
         confirm: null,
         action: undefined,
         props: null,
+        close: null,
       });
     },
     closeHovers() {
```
kloon15, line 76: We really should rename this to closeModal, the plural sense is misleading since it only closes the last modal.

frontend/src/stores/layout.ts
```diff
@@ -61,10 +71,11 @@ export const useLayoutStore = defineStore("layout", {
...
-      this.prompts.pop();
+      this.prompts.shift()?.close?.();
```
kloon15, line 77: This was correct before, its pop because we r pushing onto a list, so active modal = last element.

frontend/src/components/prompts/Prompts.vue
```diff
@@ -52,11 +50,6 @@ const components = new Map<string, any>([
 ]);
 
 watch(currentPromptName, (newValue) => {
-  if (closeModal.value) {
```
kloon15, line 55: We need to track if the modal is already open, so we dont open it twice if we came from another modal:
```
const activePrompt = layoutStore.prompts.at(-1);
  if (activePrompt && activePrompt.close != undefined) {
    // modal is already open, so do nothing
    return;
  }
```
Something like this but this is not 100% fix, it breaks the replacement modal on upload.

kloon15 (changes requested):
https://github.com/filebrowser/filebrowser/blob/25372edb5c0e616e82b76b5f523633af57d347e0/frontend/src/components/prompts/BaseModal.vue#L6
We need to change this to a function that only pops it from the list, so we dont try to double close.
```
removeModal() {
      this.prompts.pop();
    },
```
This goes to https://github.com/filebrowser/filebrowser/blob/master/frontend/src/stores/layout.ts

kloon15:
Thats my 2 cents, hope it helps to figure out the quirks.

Rmiller5466:
@kloon15 Would you be able to give some examples of the actions where you observed this PR not functioning?  I just built locally and manually put in the changes from #3635, and was not able to reproduce the issues you had observed.  The only scenario where I saw a non-desired functionality was after creating a new folder via the prompt it navigates you there and you would have to re-initiate the copy/move, though that is a separate issue that has more to do with the functionality of that button than the issue being addressed in this MR.

frontend/src/stores/layout.ts
Rmiller5466: This PR was based around the assumption that when closeHovers() was called. it was supposed to close all currently open modals and not just the last modal.  Not sure exactly how it works, though everywhere it is actually used the expectation seems to be "close any open modals" and would have potentially unwanted side effects / more extensive changes if this is only supposed to close the latest modal.

frontend/src/stores/layout.ts
Rmiller5466: Shift is what is allowing closeHovers to close out all open modals.  Without shift, I would run into cases where I would get a hanging copy modal, etc, that I couldn't interact with.

frontend/src/stores/layout.ts
kloon15: True, but there was never an actual use-case for more than 1 active modals being used when it was designed, but then again, it would still only close the last one.

kloon15:
> @kloon15 Would you be able to give some examples of the actions where you observed this PR not functioning? I just built locally and manually put in the changes from #3635, and was not able to reproduce the issues you had observed. The only scenario where I saw a non-desired functionality was after creating a new folder via the prompt it navigates you there and you would have to re-initiate the copy/move, though that is a separate issue that has more to do with the functionality of that button than the issue being addressed in this MR.

I'll double check it, im using Firefox as a sidenote.

frontend/src/stores/layout.ts
kloon15: Okay so we opened 2 modals (lets say move then new folder):
> prompts: Array["move"] -> Array["move", "newDir"]

So if u call shift() u essentially remove "move" from the array and close it.
If i got this correctly, this somehow closes all the modals right?
Maybe what im trying to say is outside the scope of this PR.

frontend/src/stores/layout.ts
```diff
@@ -29,13 +29,21 @@ export const useLayoutStore = defineStore("layout", {
     toggleShell() {
       this.showShell = !this.showShell;
     },
+    setCloseOnPrompt(closeFunction: any, onPrompt: string) {
+      let prompt = this.prompts.find((prompt) => prompt.prompt === onPrompt);
+      if (prompt) {
+        prompt.close = closeFunction;
+      }
+      return null;
```
kloon15: the return is unnecessary, it should be a void function that doesnt return anything.

frontend/src/stores/layout.ts
```diff
@@ -29,13 +29,21 @@ export const useLayoutStore = defineStore("layout", {
     toggleShell() {
       this.showShell = !this.showShell;
     },
+    setCloseOnPrompt(closeFunction: any, onPrompt: string) {
+      let prompt = this.prompts.find((prompt) => prompt.prompt === onPrompt);
```
kloon15: This should be const.

frontend/src/stores/layout.ts
```diff
@@ -29,13 +29,21 @@ export const useLayoutStore = defineStore("layout", {
     toggleShell() {
       this.showShell = !this.showShell;
     },
+    setCloseOnPrompt(closeFunction: any, onPrompt: string) {
```
kloon15: replace any with `() => Promise<string>`

frontend/src/types/layout.d.ts
```diff
@@ -3,6 +3,7 @@ interface PopupProps {
   confirm?: any;
   action?: PopupAction;
   props?: any;
+  close?: any;
```
kloon15: replace any with `() => Promise<string> | null`

kloon15:
Requested a few changes for eslint.

frontend/src/stores/layout.ts
kloon15: We should also clear out the prompts array then if thats the case.

Rmiller5466:
Just wanted to give this an update - been super busy with the holiday season and haven't had time to revisit the requested changes from above.  Hoping to get back into this in the coming week or so

PR #3543, opened 2024-10-18
fix: Use plain text for entering the sharing password

huangyxi:
**Description**

In the file `Share` panel, use plain text for the sharing password to prevent browsers from prompting to bind the file-sharing passwords to the domain.

**Further comments**


huangyxi:
Hello

o1egl:
From a security point of view, it isn't a good idea to make them plain text.

huangyxi:
Thank you for sharing your professional expertise. I'm not a security expert, but through my investigation, I do have identified a potential security risk in storing them as plain text, as described in https://www.halborn.com/disclosures/demonic-vulnerability . However, allowing browsers to prompt users to save the sharing password could lead to unintended alterations of the account password, which presents another security issue. I think the trade-off between usability and security should be discussed here, as while this change improves usability by avoiding browser prompts, it may expose the password to increased vulnerability. IMO, the risks associated with shoulder surfing or browser caching of the sharing password are tolerable in this case.

kloon15:
We could use this instead: https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/autocomplete#one-time-code
@huangyxi Could u update ur PR to this?

huangyxi:
> We could use this instead: https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/autocomplete#one-time-code @huangyxi Could u update ur PR to this?

Sure!

huangyxi:
Hello

huangyxi:
Hello, is it going to be merged into the master branch?

huangyxi:
hello

huangyxi:
Please check now.

huangyxi:
Please check.

hacdias:
For the reasons mentioned above, we're not merging this at the moment. The field **is** a password, so it will use that input type. If there is a way to tell password managers "look, this isn't a password in the general sense that you expect", then I can add that.

Changing it to type text with the one time password is a no-go.

Issue #3303, opened 2024-06-12
[docker] 2.30.0 On upload, the replace/rename modal self-dismisses before the user can click


working-name:
**Description**
If you try to upload a file that already exists, the modal self-dismisses before you get to pick rename or replace. The upload does not happen either.

This is with delete perm set to false, the rest is fairly vanilla options.

**Expected behaviour**
Modal should stay up until the user picks something.

**What is happening instead?**
No errors in the console, no network requests gone unanswered or answered with anything but a status 200. Uploads work otherwise, so not an nginx proxy manager issue.

Note: there are no network requests made since this is a pure UX javascript bug.

**Additional context**
```javascript
"perm": { ... "delete": false ... }
```

FileBrowser: 2.30.0
Docker Compose version v2.24.2
Docker CE 25.0.1
Debian 12
Client browser: brave, but seems irrelevant happens in others as well.

```yaml
version: "3.9"
services:
  holygrail:
    image: filebrowser/filebrowser
    user: 1001:1001
    deploy:
      resources:
        limits:
          cpus: "2"
          memory: "1g"
    ports:
      - "172.17.0.1:18202:80"
    restart: unless-stopped
    volumes:
      - "./drive:/srv"
      - "./filebrowser.db:/database.db"
      - "./settings.json:/config/filebrowser.json"
```

**How to reproduce?**
Just run a docker version of this and try to upload the same file you just did before to check on the modal. I don't know if pertinent but "delete" perm is set to false.

**Files**

https://github.com/filebrowser/filebrowser/assets/16505052/9236edf7-0941-4e5e-87ba-cbfba333ff6d

working-name:
Shut up bot, not stale. Just absent devs.

working-name:
Nope.

tibouh:
Same behavior here.
Deleting the file via filebrowser and uploading it again works.
Edit: more info in #3173 (issue appeared after #3111).

working-name:
I have delete perms off, so I didn't figure that part out. Thank you.

working-name:
nope.

tibouh:
nope

Daniel-dev22:
Not stale

Daniel-dev22:
Not stale

Daniel-dev22:
Not stale

working-name:
hopefully handled here https://github.com/filebrowser/filebrowser/pull/3537

tibouh:
[...]

PR #3597, opened 2024-11-15
fix: Revert using modal hook (#2689)

working-name:
`@closed` hook gets called for all actions, whether user initiated or not. This is the source of auto-closing modals without user interaction reported in #3303 and #3519. While testing its effects, I also discovered other little issues but I have no clue how to Vue so I'll leave it to the big bois.

1. Moving a file to the same folder prompts a second modal. If you hit rename it doesn't ask you for the name, it just appends `(1)` and then complains that it cannot find the file if you try move again.
2. Creating a new file with the same name of an existing one does not issue a toast notification. That's because `files.ts` api does not return a properly formatted `StatusError` [here](https://github.com/filebrowser/filebrowser/blob/master/frontend/src/api/files.ts#L139).

**Description**

> Cannot check all the builds, but /frontend builds ok with `npm run build`.

**Further comments**
I just played around with the thing in my browser and tried to understand where the calls to `closeHovers()` were coming from. Finally figured out it was the BaseModal.vue, because I had already started modifying the function to check for passed events and such. Too complicated for 0 reason. 

There are remaining issues now that you can actually interact with the UI, but this change should not get in the way. Just call the function when appropriate, i.e. after confirming a rename or whatever. `closeHovers()` should not run as a hook in my opinion, because it gets called by the internal vue dom handler on top of the modal's own actual hook code. Which is why the second prompt disappears before the user is even able to read what it asks.

I don't mind if you don't go this route, it's just food for thought.

### DIFF
--- a/frontend/src/components/prompts/BaseModal.vue
+++ b/frontend/src/components/prompts/BaseModal.vue
@@ -3,7 +3,6 @@
     class="vfm-modal"
     overlay-transition="vfm-fade"
     content-transition="vfm-fade"
-    @closed="layoutStore.closeHovers"
     :focus-trap="{
       initialFocus: '#focus-prompt',
       fallbackFocus: 'div.vfm__content',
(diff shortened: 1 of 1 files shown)


working-name:
OMG the stale bot is stalking me! 😳

working-name:
https://github.com/filebrowser/filebrowser/pull/3537

PR #3635, opened 2024-12-14
fix: Use layoutStore instead of this.$store

### LINKED ISSUE #3621: Can't create folder to move selected files
**Description**
I selected several files and after that I want to move it into new folder. But when I'm clicking create folder, nothing happens.

**Expected behaviour**
It would be nice, that new folder creates

no logs

ProfDoof:
**Description**
Looks like a this.$store was left behind after the switch to Pinia (or that's what I'm assuming right now). This change should fix the new folder button in the Move prompt. Should fix #3621. Not currently able to build locally as I am away from my dev machine so this entire PR was done in the Github web client.

### DIFF
--- a/frontend/src/components/prompts/FileList.vue
+++ b/frontend/src/components/prompts/FileList.vue
@@ -28,6 +28,7 @@
 import { mapState } from "pinia";
 import { useAuthStore } from "@/stores/auth";
 import { useFileStore } from "@/stores/file";
+import { useLayoutStore } from "@/stores/layout";
 
 import url from "@/utils/url";
 import { files } from "@/api";
@@ -49,6 +50,7 @@ export default {
   computed: {
     ...mapState(useAuthStore, ["user"]),
     ...mapState(useFileStore, ["req"]),
+    ...mapState(useLayoutStore, ["showHover"]),
     nav() {
       return decodeURIComponent(this.current);
     },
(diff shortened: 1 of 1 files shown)


Rmiller5466:
Pulled down and built, this seems to fix the mentioned issue.  Really need #3537 in first, as soon as the new folder button is clicked all the prompts disappear.  But hey, clicking the button _does_ bring up the new folder prompt now!

kloon15:
Yup it slipped our attention during the migration.

> Pulled down and built, this seems to fix the mentioned issue. Really need #3537 in first, as soon as the new folder button is clicked all the prompts disappear. But hey, clicking the button _does_ bring up the new folder prompt now!

Im testing with ur PR included, and unfortunately both prompts disappear if i create/cancel new folder. Im afraid more work is needed in #3537 to properly handle more than 1 open modal.

Rmiller5466:
@kloon15 thanks for checking that, I'll investigate it more tonight if able.  I coulda sworn the disappearing was fixed with more than 1, thought I tested that specific concern via the rename/replace prompt.  Will update once I have the time.

kloon15:
@ProfDoof If u have time, could u look into updating the directory list if a new folder is created?
Also idk how it worked before, but it should create folder at the directory tree we r browsing, not where the file we wanna move is located.

PR #3666, opened 2025-01-10
fix: Fix user creation on proxy auth

ahobsonsayers:
**Description**

 There was an error I made in #3569 where - regardless of the fact we were creating a user if they user did not exists - the error was still be returned at the end of the function as we were never actually setting it to nil. This mean the PR was not working as expect as raised by @melaniestnls in the comments of  #3569.

This PR fixes this issue by ensuring we are returning a nil error, while still returning any error that occurs that is **not** a user does not exist error

**Further comments**
N/A


auth/proxy.go
```diff
@@ -22,7 +22,12 @@ type ProxyAuth struct {
 func (a ProxyAuth) Auth(r *http.Request, usr users.Store, setting *settings.Settings, srv *settings.Server) (*users.User, error) {
 	username := r.Header.Get(a.Header)
 	user, err := usr.Get(srv.Root, username)
-	if errors.Is(err, fbErrors.ErrNotExist) {
+	if err != nil {
+		// Return the error unless it is a user does not exist error
+		if !errors.Is(err, fbErrors.ErrNotExist) {
+			return nil, err
+		}
+
```
o1egl: This code is hard to read. I think it would be better to extract user creation into the createUser function
```go
if errors.Is(err, fbErrors.ErrNotExist) {
    return createUser()
}
return user, err
```

Issue #3670, opened 2025-01-14
After_upload command runner does not work.


BeziCZ:
**Description**
Some command runners do not work. I want to automatically unzip the archive after it is uploaded. However, it does not work on versions later than 2.23.0. 

**Expected behaviour**
After a file is uploaded, the file gets unzipped, and the archive's contents are available.

**What is happening instead?**
The archive is uploaded but not unzipped.

**Additional context**
I know that downgrading to v2.23.0 would solve that issue; however, not in my case. I use Cloudflare, and there is an upload limit. From what I could find in the history of this repo, chunked upload is not yet supported in v2.23.0 and was introduced in v2.24.0. Thus, I have two options: the first is to be able to upload large files with no automatic unzipping, or the second is to have a working command runner but not the possibility of uploading large files.

**How to reproduce?**
Reproduction is quite simple. Deploy Filebrowser docker image, set up the rule, and then try to see if it works.

jantari:
duplicate of #2993

Issue #2993, opened 2024-02-07
Upload Event Command Runner Not Triggered


cesboa:
**Description**
Some "Command runner" does not work. Like "After Upload" or "Before Upload", while "After Rename" and "After Delete Work"
Filebrowser 2.27.0 running via brew on latest macOS

**Expected behaviour**
All "Command runners" to work

**What is happening instead?**
Nothing happens

**How to reproduce?**
Add a command to eg. "After Upload" and upload something. The command is not executed.
See attached.
Only After Delete and After Rename makes a directory, the other two does not make a directory. stdout gives info on the ones who works, nothing in the other two.

![Screenshot 2024-02-07 at 14 20 32](https://github.com/filebrowser/filebrowser/assets/1929823/47e1c624-2ead-477a-ac23-9c3a4ae50e42)

DaaGH:
Yes, I have the same thing. Docker-Alpine. v.2.24(s6)-2.27(s6). 2.23(s6) - ok.

jantari:
I am having the same issue. My usecase is that all files should be (re)named to their sha256 hash automatically after upload, so I configured a "After Upload" command runner with the following:

```bash
mv -- "$FILE" "$(dirname -- "$FILE")/$(sha256sum -- "$FILE" | cut -d ' ' -f 1)"
```

this works when I run it inside the container via `docker exec -it filebrowser /bin/sh`, but when configured as a command runner there is nothing in the logs and the file isn't renamed, just nothing happens.

jrukavina:
Same problem here, specifically for "After Upload" and "Before Upload" events.
Version 2.27.0

brett-james-rocketlab:
Getting this problem as well. I got it to happen for before / after save, but never before / after upload.

DgRosa:
I have the same problem, just trying to raise further awareness of this issue.

RomanSonnik:
Is there any fixes for that?

brett-james-rocketlab:
Yes, the problem came when chunked file support was added. So roll back to use I think it is 2.23 

.

Unfortunately brew doesn't have tsgged releases support for this repo, so I used docker instead and solved the problem.

RussellLuo:
Just encountered the same issue! Based on the answer from @brett-james-rocketlab (thanks 🙏), I got it to work with the following command:

```bash
docker run -d --name=filebrowser \
    -v /path/to/root:/srv \
    -v /path/filebrowser.db:/database.db \
[...]

PR #3676, opened 2025-01-16
fix: Upload event command runners

BeziCZ:
**Description**
This PR contains changes which add to the TUS uploading that when merged broke command runner triggered on events BEFORE_UPLOAD and AFTER_UPLOAD. This required creating two new hook methods for runner, one for before event and one for after event. These methods were needed because of how the TUS chunks the processed file. 

This PR solves following issues
[#3670]
[#2993]

### DIFF
--- a/http/tus_handlers.go
+++ b/http/tus_handlers.go
@@ -131,6 +131,13 @@ func tusPatchHandler() handleFunc {
 			)
 		}
 
+		if uploadOffset == 0 {
+			err = d.RunBeforeHook("upload", r.URL.Path, "", d.user)
+			if err != nil {
+				return errToStatus(err), err
+			}
+		}
+
 		openFile, err := d.user.Fs.OpenFile(r.URL.Path, os.O_WRONLY|os.O_APPEND, files.PermFile)
 		if err != nil {
 			return http.StatusInternalServerError, fmt.Errorf("could not open file: %w", err)
@@ -149,6 +156,14 @@ func tusPatchHandler() handleFunc {
 		}
 
 		w.Header().Set("Upload-Offset", strconv.FormatInt(uploadOffset+bytesWritten, 10))
+
+		if bytesWritten < int64(d.Settings.Tus.ChunkSize) {
+			fmt.Printf("Final chunk detected (size: %d). Running after hook\n", bytesWritten)
+			err = d.RunAfterHook("upload", r.URL.Path, "", d.user)
+			if err != nil {
+				return errToStatus(err), err
+			}
+		}
 
 		return http.StatusNoContent, nil
 	})
--- a/http/tus_handlers.go
+++ b/http/tus_handlers.go
@@ -158,7 +158,6 @@ func tusPatchHandler() handleFunc {
 		w.Header().Set("Upload-Offset", strconv.FormatInt(uploadOffset+bytesWritten, 10))
 
 		if bytesWritten < int64(d.Settings.Tus.ChunkSize) {
-			fmt.Printf("Final chunk detected (size: %d). Running after hook\n", bytesWritten)
 			err = d.RunAfterHook("upload", r.URL.Path, "", d.user)
 			if err != nil {
 				return errToStatus(err), err
--- a/http/tus_handlers.go
+++ b/http/tus_handlers.go
@@ -157,7 +157,7 @@ func tusPatchHandler() handleFunc {
 
 		w.Header().Set("Upload-Offset", strconv.FormatInt(uploadOffset+bytesWritten, 10))
 
-		if bytesWritten < int64(d.Settings.Tus.ChunkSize) {
+		if uint64(bytesWritten) < (d.Settings.Tus.ChunkSize) {
 			err = d.RunAfterHook("upload", r.URL.Path, "", d.user)
 			if err != nil {
 				return errToStatus(err), err
--- a/runner/runner.go
+++ b/runner/runner.go
@@ -17,6 +17,39 @@ type Runner struct {
 	*settings.Settings
 }
 
+// RunBeforeHooks and RunAfterHooks is used for the chunked upload since it cannot use the standard RunHook
+func (r *Runner) RunBeforeHook(evt, path, dst string, user *users.User) error {
+	path = user.FullPath(path)
+	dst = user.FullPath(dst)
+	if r.Enabled {
+		if val, ok := r.Commands["before_"+evt]; ok {
+			for _, command := range val {
+				err := r.exec(command, "before_"+evt, path, dst, user)
+				if err != nil {
+					return err
+				}
+			}
+		}
+	}
+	return nil
+}
+
+func (r *Runner) RunAfterHook(evt, path, dst string, user *users.User) error {
+	path = user.FullPath(path)
+	dst = user.FullPath(dst)
+	if r.Enabled {
+		if val, ok := r.Commands["after_"+evt]; ok {
+			for _, command := range val {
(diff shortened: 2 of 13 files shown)


http/tus_handlers.go
```diff
@@ -150,6 +157,14 @@ func tusPatchHandler() handleFunc {
 
 		w.Header().Set("Upload-Offset", strconv.FormatInt(uploadOffset+bytesWritten, 10))
 
+		if bytesWritten < int64(d.Settings.Tus.ChunkSize) {
+			fmt.Printf("Final chunk detected (size: %d). Running after hook\n", bytesWritten)
```
o1egl: Remove debug prints

runner/runner.go
```diff
@@ -17,6 +17,39 @@ type Runner struct {
 	*settings.Settings
 }
 
+// RunBeforeHooks and RunAfterHooks is used for the chunked upload since it cannot use the standard RunHook
```
o1egl: I believe it's sufficient for you to explain the function within your comments. There's no need to indicate where it is utilized.

http/tus_handlers.go
BeziCZ: Removed in added commits

runner/runner.go
BeziCZ: Rewritten to exclude this information

BeziCZ:
@o1egl Added another fix to this PR. Now it also resolves the problem of running the commands with shell.
Up to now when in settings `Execute on shell` was set to `/bin/sh -c` you got the response `/bin/sh -c: no such file or directory`

This was fixed by splitting the command to `/bin/sh` and `-c` the same way calling binaries is processed.

BeziCZ:
Another feature added, which I found broken. Upon entering the shell using `filebrowser config set --shell="sh -c"` the result was that into the db it was entered as `sh-c` without the space, which is obviously not functioning, As such I changed the command parsing to be separated by `,`

pjw91:
The bug, setting "Execute on shell" from UI (/settings/global) not working, stems from this commit: 5100e587d73831ecdb5e3bd35a78fef96ad248a4

In frontend/src/views/settings/Global.vue, you will see

```diff
-              v-model="settings.shell"
+              v-model="shellValue"
-      settings.shell = settings.shell.join(" ");
-        shell: this.settings.shell.trim().split(" ")
+  newSettings.shell = shellValue.value.split("\n");
+    shellValue.value = newSettings.shell.join("\n");
```

The `shell` field in `PUT /api/settings` request is typed as '[]string', and
* was split **by space** from the value in input[text] html tag, eg. "sh -c" was splitted into ["sh", "-c"]
* is split **by newline** from the value in input[text] html tag, eg. "sh -c" is splitted into ["sh -c"]

However, we could never type in a newline in an html input[text] tag

The parsing was done in multiple places (javascript/vue when user gives the value in web, and golang when taken from command line). This complexity results in the bug.

This MR proposed all parsing stuff in golang, is a good approach in my opinion.
I guess we should also change the api from '[]string' to 'string'.

jantari:
eagerly awaiting these fixes :)

BeziCZ:
> The bug, setting "Execute on shell" from UI (/settings/global) not working, stems from this commit: [5100e58](https://github.com/filebrowser/filebrowser/commit/5100e587d73831ecdb5e3bd35a78fef96ad248a4)
> 
> In frontend/src/views/settings/Global.vue, you will see
> 
> ```diff
> -              v-model="settings.shell"
> +              v-model="shellValue"
> -      settings.shell = settings.shell.join(" ");
> -        shell: this.settings.shell.trim().split(" ")
> +  newSettings.shell = shellValue.value.split("\n");
> +    shellValue.value = newSettings.shell.join("\n");
> ```
> 
> The `shell` field in `PUT /api/settings` request is typed as '[]string', and
> 
>     * was split **by space** from the value in input[text] html tag, eg. "sh -c" was splitted into ["sh", "-c"]
> 
>     * is split **by newline** from the value in input[text] html tag, eg. "sh -c" is splitted into ["sh -c"]
> 
> 
> However, we could never type in a newline in an html input[text] tag
> 
> The parsing was done in multiple places (javascript/vue when user gives the value in web, and golang when taken from command line). This complexity results in the bug.
> 
> This MR proposed all parsing stuff in golang, is a good approach in my opinion. I guess we should also change the api from '[]string' to 'string'.

Thanks for the feedback. I made some changes according to that. 

1. I removed processing on the frontend which IMO is just bad design. 
2. I changed from `[]string` to just `string` on frontend, because it is not needed.
3. IMO having the command on backend as a slice is just as useless as on the frontend. Since it would require additional logic and decision making to figure out what shell command should be executed with what command I really do not see any point in it. Thus I switched from `[]string` to `string` in backend as well. 
4. I removed `convertCmdStrtoCmdArray` function since it is not needed anymore after the type change.
5. Conversion of bytesWritten in the comparison to uint64 to make CI happy

pjw91:
Will changing settings.Setting Shell to string breaks filebrowser.db compatibility?

BeziCZ:
> Will changing settings.Setting Shell to string breaks filebrowser.db compatibility?

From what I found when testing it doesn't.

pjw91:
The old.config export from old.db:
```{
    "settings": {
        "key": "...",
        "signup": false,
        "createUserDir": false,
        "userHomeBasePath": "/users",
        "defaults": {
            "scope": ".",
            "locale": "en",
            "viewMode": "mosaic",
            "singleClick": false,
            "sorting": {
                "by": "",
                "asc": false
            },
            "perm": {
                "admin": false,
                "execute": true,
                "create": true,
                "rename": true,
                "modify": true,
                "delete": true,
                "share": true,
                "download": true
            },
            "commands": [],
            "hideDotfiles": false,
            "dateFormat": false
        },
        "authMethod": "noauth",
        "branding": {
            "name": "",
            "disableExternal": false,
            "disableUsedPercentage": false,
            "files": "",
            "theme": "",
            "color": ""
        },
        "tus": {
            "chunkSize": 10485760,
            "retryCount": 5
        },
        "commands": {
            "after_copy": [],
            "after_delete": [],
            "after_rename": [],
            "after_save": [],
            "after_upload": [],
            "before_copy": [],
            "before_delete": [
                "[ -h \"$FILE\" ] \u0026\u0026 unlink \"$(realpath \"$FILE\")\" || :"
            ],
            "before_rename": [],
            "before_save": [],
            "before_upload": []
        },
        "shell": [
            "sh",
            "-c"
        ],
        "rules": []
    },
    "server": {
        "root": "mnt",
        "baseURL": "",
        "socket": "",
        "tlsKey": "",
        "tlsCert": "",
        "port": "5900",
        "address": "127.0.0.1",
        "log": "stdout",
        "enableThumbnails": false,
        "resizePreview": false,
        "enableExec": false,
        "typeDetectionByHeader": false,
        "authHook": "",
        "tokenExpirationTime": ""
    },
    "auther": {}
}
```
I've set `shell` and `before_delete`.

And loading the old.db:
```
./filebrowser.new -d old.db -r mnt -p 6543
2025/02/18 22:32:14 Using database: /dev/shm/old.db
2025/02/18 22:32:14 No config file used
2025/02/18 22:32:14 Listening on 127.0.0.1:6543
2025/02/18 22:32:19 ERROR: couldn't get settings: json: cannot unmarshal array into Go struct field Settings.shell of type string
```

BeziCZ:
Not stale, just have a lot of work... Christ with this bot

jantari:
Ermahgerd.....

BeziCZ:
Not stale

PR #3742, opened 2025-02-15
fix: set videojs locale

### LINKED ISSUE #3721: Video controls on iPadOS mobile Safari
Hello,
I am on iPadOS 18.3, but this issue is present since at least last year.

When I use mobile Safari and I select a video file, the controls for playback and full screen/picture in picture are not displayed I only get the play button Once the video is loaded but when I tap it, I can only reload the page to stop the playback or go back in the browser to return to the directory listing.

The last working version that I am using now is 2.27.0.

The server runs on latest Alpine, binary file is  from the release.

Is there anything I can do to use the old player in new versions? A switch in the option page would be greater.

Thank you 

**Expected behaviour**
I should be able to playm pause, rewind, fast forward, slide to a different position, get the time since start and to the end, end of the video, enable and disable full screen and picture in picture, close the playback

After I press the play button, the video start playing then I cannot control the playback at all

**Additional context**
Works correctly up to version 2.2 7.0 included

**How to reproduce?**
Just play an MP4 video file in mobile Safari on iPadOS 18.3

similicious:
**Description**
fixes #3721 

Sets the `code` argument of `addLanguage` to a proper IANA language code instead of the static string `videoPlayerLocal`. See [videojs docs](https://videojs.com/guides/languages/#adding-languages-via-the-api).

This fixes the exception reported in #3721 causing the video controls to not show.


giulianozor:
Any ETÀ for merging this ?

giulianozor:
Is this project still maintained?

PR #3863, opened 2025-04-14
fix: Graceful shutdown

adrienkohlbecker:
**Description**

Hello,

I’ve noticed that the process always exits with a non-zero code because of the use of `log.Fatal` in the error block for `http.ListenAndServe` (which always returns a non-nil error). In my installation, it causes systemd to believe that the process failed even during routine restarts, which triggers alerts in my monitoring.

Additionally, because the `listener` is closed before the server, sometimes there is an additional error: 

```
2025/04/14 15:09:39 Using database: PATH/filebrowser.db
2025/04/14 15:09:39 No config file used
2025/04/14 15:09:39 Listening on 127.0.0.1:8080
^C2025/04/14 15:09:41 Caught signal interrupt: shutting down.
2025/04/14 15:09:41 accept tcp 127.0.0.1:8080: use of closed network connection
```

Using the code from https://dev.to/mokiat/proper-http-shutdown-in-go-3fji, I’ve fixed the graceful shutdown mechanism for the HTTP server to ensure a non-zero exit code.

Let me know what you think!

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -77,8 +77,12 @@ linters:
         - "1"
         - "2"
         - "3"
+        - "0666"
+        - "0700"
+        - "0700"
       ignored-functions:
         - strings.SplitN
+        - make
     nolintlint:
       allow-unused: false # report any unused nolint directives
       require-explanation: false # require an explanation for nolint directives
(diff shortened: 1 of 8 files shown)


adrienkohlbecker:
Hi @o1egl, 

This is not stale, I don't think I'm allowed to remove the label and to re-open the PR. 
Could this bot be too eager?

Thanks

hacdias (approved):

Issue #870, opened 2019-10-07
Feature request: remove logout on proxy auth


fdefilippo:
When filebrowser is configured to authenticate users via environment variable (reverse proxy) the logout link should be removed or hidden or other action. 

thank you.

ovidiucp:
I'd actually like to have the logout button redirect to a custom URL, that could be user specified.

I'm using oauth2_proxy with nginx. The former uses the /oauth2/sign_in path to log out the current user. I'd like to setup the Logout button in filebrowser to point to that URL.

ovidiucp:
It turns out logging out of oauth2_proxy cannot be done in filebrowser, since the cookie is an HttpOnly cookie. The Logout button was confusing some of my users, so I took a stab at this.

PR #3884, opened 2025-04-30
feat: configurable logout page URL for proxy/hook auth

jake-dog:
**Update 2025-05-13:** I've been running this for the past few weeks and added two commits to improve it.  The integration with oauth2-proxy is seamless!

**Description**

* add configurable logout page for use with proxy/hook auth
* allow token renewal after expiration when safe to do so with proxy auth and a custom logout page
* improve security by [validating alg as recommended](https://pkg.go.dev/github.com/golang-jwt/jwt/v4#section-readme)

### Configurable `logoutPage`

The logout button is hidden when proxy auth is used (re: #934) because, at the time, [_logging out of oauth2_proxy cannot be done in filebrowser_](https://github.com/filebrowser/filebrowser/issues/870#issuecomment-561437539).

oauth2_proxy has since been updated to call the upstream auth provider's logout method, re: oauth2-proxy/oauth2-proxy#1876.  filebrowser requires changes before this feature can be used:

1. add a configurable `logoutPage` URL to filebrowser settings
2. unhide the logout button when `logoutPage` is not set default (ie. `/login`)

Example:

* set a custom `logoutPage` for filebrowser

      ./filebrowser config set --logoutpage "/oauth2/sign_out?rd=https%3A%2F%2Ffilebrowser"

* run oauth2-proxy with custom logout URL (for keycloak)

      --backend-logout-url='https://keycloak/realms/myrealm/protocol/openid-connect/logout?id_token_hint={id_token}'

### Token renewal after expiration

Users with proxy auth and a custom logout page are likely to encounter friction with filebrowser's default token renewal mechanism.  The lifetime of session/renewal tokens issued by an auth proxy will conflict with filebrowser's JWT, prematurely logging out the user.  To mitigate this issue the filebrowser token may be renewed after expiration if, and only if, the following conidtions are met:

1. proxy auth is used with a custom logoutpage
2. the filebrowser issued JWT is valid in every way except the expiration date

### Validate token 'alg'

The golang-jwt/jwt/v4 library strongly encourages that the implementer validate the token 'alg'.  From the documentation:

> _The caller is strongly encouraged to set the WithValidMethods option to validate the 'alg' claim in the token matches the expected algorithm._

In light of the fact that this request makes it possible to renew expired tokens under specific circumstances, it makes sense to also ensure all JWT security guidance is strictly adhered to.


cmd/config.go
```diff
@@ -35,6 +35,7 @@ func addConfigFlags(flags *pflag.FlagSet) {
 	flags.Bool("create-user-dir", false, "generate user's home directory automatically")
 	flags.Uint("minimum-password-length", settings.DefaultMinimumPasswordLength, "minimum password length for new users")
 	flags.String("shell", "", "shell command to which other commands should be appended")
+	flags.String("logout-page", "", "url of custom logout page")
```
hacdias: I'm wondering if this should not be part of auth. So `auth.logout-page`. What do you think?

hacdias (commented):
@jake-dog this is nice. I just updated your PR with the latest changes from `main`. Could you see my comment above?

Also, could you validate if it works correctly with my updates?

cmd/config.go
jake-dog: "Logout" is a term that only makes sense in the context of authentication, so I agree it should be moved to `auth.logout-page`.  I will work on this change, and test the updated branch.

hacdias: Great. I will come back here once it's been updated.

Also, please use `auth.logoutPage` for consistency with the rest of the flags. I've updated the few kebab-case flags to lowerCamelCase in a backwards compatible manner.

jake-dog:
During testing I discovered that the idle timeout mechanism from #5439 conflicts with proxy auth's external token lifetime.  As a temporary fix I've disabled the idle timeout logout mechanism when proxy auth and custom logout is used (eebfdbc), however a future improvement could be allow the idle timeout to be set to an arbitrary user configured value, or disabled entirely (set zero or negative).

Furthermore I noticed that the JWT was being accidentally validated twice per request due to #5535.  The first token validation occurs in `request.ParseFromRequest()`, and the redundant validation occurs in `(*jwt.Validator).Validate()`.

https://github.com/filebrowser/filebrowser/blob/79980bcf52a9540aae89474fb470d33c66d97887/http/auth.go#L71
https://github.com/filebrowser/filebrowser/blob/79980bcf52a9540aae89474fb470d33c66d97887/http/auth.go#L76

Filebrowser can regain a tiny bit of performance by omitting the redundant validation (8e5c0b1).

<details>
  <summary>Breadcrumbs for how JWTv5 ParseFromRequest also validates</summary>
  
  1. [`request.ParseFromRequest()` calls `jwt.ParseWithClaims`](https://github.com/golang-jwt/jwt/blob/e9547a11aa603c52a31e315cdcaa71d0f25b921a/request/request.go#L39)
  2. [`jwt.ParseWithClaims` calls `(*jwt.Validator).Validate()`](https://github.com/golang-jwt/jwt/blob/e9547a11aa603c52a31e315cdcaa71d0f25b921a/parser.go#L123)
</details>

cmd/config.go
jake-dog: I believe this is resolved in d3a9f34, though it appears I now have a new merge conflict to resolve.

cmd/config.go
```diff
@@ -201,18 +196,17 @@ func printSettings(ser *settings.Server, set *settings.Settings, auther auth.Aut
 	fmt.Fprintf(w, "Sign up:\t%t\n", set.Signup)
 	fmt.Fprintf(w, "Hide Login Button:\t%t\n", set.HideLoginButton)
 	fmt.Fprintf(w, "Create User Dir:\t%t\n", set.CreateUserDir)
+	fmt.Fprintf(w, "LogoutPage:\t%s\n", set.LogoutPage)
```
hacdias:
```suggestion
	fmt.Fprintf(w, "Logout Page:\t%s\n", set.LogoutPage)
```

frontend/src/utils/constants.ts
```diff
@@ -10,6 +10,7 @@ const version: string = window.FileBrowser.Version;
 const logoURL = `${staticURL}/img/logo.svg`;
 const noAuth: boolean = window.FileBrowser.NoAuth;
 const authMethod = window.FileBrowser.AuthMethod;
+const logoutPage: string = window.FileBrowser.LogoutPage || "/login";
```
hacdias: Is the `|| "/login"` needed, considering that we're already setting the default on the backend in case it is not set? I'd rather have defaults in just one place instead of having them in multiple places.

http/auth.go
```diff
@@ -61,20 +63,32 @@ func (e extractor) ExtractToken(r *http.Request) (string, error) {
...
 }
 
+func renewableErr(err error, d *data) bool {
+	if d.settings.AuthMethod != fbAuth.MethodProxyAuth || err == nil {
+		return false
+	}
+
+	if d.settings.LogoutPage == "" || d.settings.LogoutPage == settings.DefaultLogoutPage {
+		return false
+	}
```
hacdias, line 73: Considering we set `d.settings.LogoutPage` to `settings.DefaultLogoutPage` in `storage.go` when fetching the settings, do we need to compare it with an empty string here? Will it not always be set then?

hacdias (commented):
Just a few questions!

frontend/src/utils/constants.ts
jake-dog: Agreed.  I incorrectly set it here because I couldn't figure out how the value was propagated, but I see it now (via golang template to inject inline javascript).  I'll remove the redundant assignment.

http/auth.go
jake-dog: After reviewing the code I believe in all cases `d.settings.LogoutPage` is initialized to `settings.DefaultLogoutPage` and not an empty string.  I will remove the ineffectual comparison.

hacdias (approved):

ELERON23:
Hi,
Is there documentation on how to set this up?
Thank you

jake-dog:
> Hi, Is there documentation on how to set this up? Thank you

I'm afraid not, but there is reasoning behind that.  There are many possible permutations of frontend authn proxy and the `auth.logoutPage` flag provides only a tiny abstract integration.  It would be very hard to document these options in a useful way without documenting other software with its own documentation.

PR #3896, opened 2025-05-10
fix: recursively remove associated shares when folder/file is deleted

stavros-tsioulis:
**Description**
When deleting a folder or a file, remove all shares associated with the file, or, if it is a folder, every file and folder beneath it.

**Further comments**
This is my first time writing go code so please let me know if I did something wrong!

### DIFF
--- a/storage/bolt/share.go
+++ b/storage/bolt/share.go
@@ -76,10 +76,10 @@ func (s shareBackend) Delete(hash string) error {
 	return err
 }
 
-func (s shareBackend) DeleteWithPath(path string) error {
-	err := s.db.Select(q.Eq("Path", path)).Delete(&share.Link{})
-	if errors.Is(err, storm.ErrNotFound) {
-		return nil
+func (s shareBackend) DeleteWithPath(pathPrefix string) {
+	var links []share.Link
+	s.db.Prefix("Path", pathPrefix, &links)
+	for _, link :=range links {
+		s.db.DeleteStruct(&share.Link{Hash: link.Hash})
 	}
-	return err
 }
--- a/storage/bolt/share.go
+++ b/storage/bolt/share.go
@@ -76,10 +76,16 @@ func (s shareBackend) Delete(hash string) error {
 	return err
 }
 
-func (s shareBackend) DeleteWithPathPrefix(pathPrefix string) {
+func (s shareBackend) DeleteWithPathPrefix(pathPrefix string) error {
 	var links []share.Link
 	s.db.Prefix("Path", pathPrefix, &links)
-	for _, link := range links {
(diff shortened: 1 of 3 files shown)


storage/bolt/share.go
```diff
@@ -75,3 +75,11 @@ func (s shareBackend) Delete(hash string) error {
...
 	return err
 }
+
+func (s shareBackend) DeleteWithPath(path string) error {
+	err := s.db.Select(q.Eq("Path", path)).Delete(&share.Link{})
+	if errors.Is(err, storm.ErrNotFound) {
+		return nil
+	}
+	return err
+}
```
hacdias: Thanks for the PR @stavros-tsioulis!

I was thinking about this piece of code.

I think there might be a more efficient way to do this. I see the `path` is an index in storm. Isn't it possible to use the `db.Prefix` to delete every matching share with the path as prefix? If that'd work, we wouldn't need to enumerate all children files and directories.

stavros-tsioulis: Hello.

Indeed, I missed that feature of Storm when reading its documentation... I've updated it now. Let me know if I should adjust anything else :)

storage/bolt/share.go
```diff
@@ -75,3 +75,11 @@ func (s shareBackend) Delete(hash string) error {
...
 	return err
 }
+
+func (s shareBackend) DeleteWithPathPrefix(pathPrefix string) {
+	var links []share.Link
+	s.db.Prefix("Path", pathPrefix, &links)
+	for _, link := range links {
+		s.db.DeleteStruct(&share.Link{Hash: link.Hash})
+	}
+}
```
hacdias: We should check for errors. Maybe something like this:

```suggestion
func (s shareBackend) DeleteWithPathPrefix(pathPrefix string) error {
	var links []share.Link
	if err := s.db.Prefix("Path", pathPrefix, &links); err != nil {
		return err
	}
	var err error
	for _, link := range links {
		err = errors.Join(err, s.db.DeleteStruct(&share.Link{Hash: link.Hash}))
	}
	return err
}
```

And also propagate the error back to the delete handler. There, I wouldn't return a 500, but instead just log it as a warning.

storage/bolt/share.go
hacdias: Thank you :) I made another comment. I also rebased the PR to be up to date with `master`.

storage/bolt/share.go
stavros-tsioulis: Right. Joining errors is something I considered but didn't know how to handle them, and decided to not do anything initially since it failing is not fatal in regards to the file deletion handler. Sorry for this back and forth though... Rebased and updated!

hacdias (approved):
Thanks a lot :D

PR #3899, opened 2025-05-11
fix: Fix upload command runners

BeziCZ:
**Description**
Followup to PR #3676 which gotten closed since I was quite busy and didn't have time to finish the changes. This PR contains changes which add to the TUS uploading that when merged broke command runner triggered on events BEFORE_UPLOAD and AFTER_UPLOAD. This required creating two new hook methods for runner, one for before event and one for after event. These methods were needed because of how the TUS chunks the processed file. Apart from these changes, there is also a change in `Setting` struct from `settings` package The attribute `Shell` is changed from `[]string` to `string`, this has broken backwards compatibility with importing old databases, however in the last commit I solve this by creating a function that unmarshalls json when importing the configuration.

This PR solves following issues
#3670 
#2993


sir-shalahuddin:
when after upload command runner fixed :(

BeziCZ:
> when after upload command runner fixed :(

When/if someone looks at this

i0ntempest:
This patch now has conflicts because of upstream changes. The two conflicting can be removed as the upstream code now includes them (or something similar).

http/tus_handlers.go
```diff
@@ -150,6 +157,13 @@ func tusPatchHandler() handleFunc {
 
 		w.Header().Set("Upload-Offset", strconv.FormatInt(uploadOffset+bytesWritten, 10))
 
+		if uint64(bytesWritten) < (d.Settings.Tus.ChunkSize) {
+			err = d.RunAfterHook("upload", r.URL.Path, "", d.user)
+			if err != nil {
+				return errToStatus(err), err
+			}
+		}
+
```
hacdias, line 166: What happens if the file size is exactly a multiple of the chunk size? I think then the after hook won't run.

hacdias (changes requested):
I'd like to see this merged, but it needs to be rebased on top of the latest changes, which also fixes the issues around the `Shell` setting. 

The after event needs to be triggered in a different way, however, due to the issue  I mentioned in the other comment.

hacdias:
Superseded by #5270

PR #3927, opened 2025-05-24
fix: abort ongoing requests when changing pages

manx98:
**Description**
In situations where the network is slow, users switch directories without canceling their previous requests for directory listings. This oversight results in ongoing invalid requests, which ultimately contributes to the occurrence of this issue.
Upon further examination of the code, I discovered that several request functions were still executing outdated requests. Consequently, I use the AbortController to promptly cancel these obsolete requests.


frontend/src/views/Files.vue
```diff
@@ -61,9 +61,7 @@ const route = useRoute();
 
 const { t } = useI18n({});
 
-const clean = (path: string) => {
-  return path.endsWith("/") ? path.slice(0, -1) : path;
-};
+let fetchDataController = new AbortController();
```
hacdias, line 65: Why is this removed?

manx98: This function was removed as it's no longer called by any code.

hacdias: Well, that I understood, still doesn't explain why it is no longer used. Anyways, I double checked and it seems to be because it's no longer necessary since we're not expecting to get old responses back.

hacdias (approved):
Thank you for this!

Issue #2850, opened 2023-11-28
Use a different key binding for search


pouriamoosavi:
**Is your feature request related to a problem? Please describe.**
Yes, I want to be able to use my own browser's search with the Ctrl+F key binding.

**Describe the solution you'd like**
Use another key binding to open the file browser search panel. For example, Ctrl+K. This key binding is used for searching in many documents, including your own document webpage (https://filebrowser.org/).

**Describe alternatives you've considered**
Add an option to disable the file browser search key binding completely. Be able to open it only by clicking on the search bar on the top.

Thanks

pouriamoosavi:
Completed?

bschwind:
Big +1 to this, overriding the default Ctrl-F broswer keybinding is maddening.

It's made _worse_ by the fact that I see text on the screen I want to search for, I press Ctrl-F, then suddenly the search UI pops up. I type the string I was going to type...and it finds nothing.

I search for the same string with the native browser search and it finds it perfectly.

Don't override the default browser keybinding.

WolfgangDpunkt:
I have the same problem. The ctrl+f browser search works perfectly to quickly jump to an item in the list, but by having this key binding hijacked from filebrowser I need to use the internal search that never works.
Please let uns config this!

contributor:
There is a trick to invoke browser native Ctrl+F search - turn CapsLock on before pressing Ctrl+F

contributor:
PR to fix this: https://github.com/filebrowser/filebrowser/pull/4638

PR #4638, opened 2025-05-25
fix: search uses ctrl+shift+f instead of hijacking browser's ctrl+f

contributor:
fixes issue https://github.com/filebrowser/filebrowser/issues/2850

Since there is no configuration for key bindings, it should be ok just to replace Ctrl+F with Ctrl+Shift+F in code.
The latter doesn't conflict with native browser hot keys


hacdias (approved):
I do agree that hijacking the browser's native search function is not the best in this case. If we were searching within some open file, or only a specific editor, it'd make sense.

PR #4716, opened 2025-05-26
feat: support streaming response for search results

manx98:
**Description**
1. Use streaming real-time return responses to enhance the prompt return of search results and avoid timeout issues caused by overly lengthy searches.
2. Optimize the search process on the server side and promptly terminate the search for invalid requests to continue execution.
3. Support the active termination of search requests to avoid browser lag caused by a large number of returned results.
show

### DIFF
--- a/frontend/src/api/files.ts
+++ b/frontend/src/api/files.ts
@@ -2,14 +2,22 @@ import { useAuthStore } from "@/stores/auth";
 import { useLayoutStore } from "@/stores/layout";
 import { baseURL } from "@/utils/constants";
 import { upload as postTus, useTus } from "./tus";
-import { createURL, fetchURL, removePrefix } from "./utils";
+import { createURL, fetchURL, removePrefix, StatusError } from "./utils";
 
-export async function fetch(url: string) {
+export async function fetch(url: string, signal?: AbortSignal) {
   url = removePrefix(url);
+  const res = await fetchURL(`/api/resources${url}`, { signal });
 
-  const res = await fetchURL(`/api/resources${url}`, {});
-
-  const data = (await res.json()) as Resource;
+  let data: Resource;
+  try {
+    data = (await res.json()) as Resource;
+  } catch (e) {
+    // Check if the error is an intentional cancellation
+    if (e instanceof Error && e.name === "AbortError") {
+      throw new StatusError("000 No connection", 0, true);
+    }
+    throw e;
+  }
   data.url = `/files${url}`;
 
   if (data.isDir) {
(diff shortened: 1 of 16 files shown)


ramiresviana:
This looks really clever—it required no drastic changes on either the backend or frontend, yet it has a big impact for users searching large folders. It seems reasonable to merge this since it enhance/fixes an existing feature. The only concern we need to address first is browser compatibility, as Safari doesn’t appear to support streams for the response body.

manx98:
> This looks really clever—it required no drastic changes on either the backend or frontend, yet it has a big impact for users searching large folders. It seems reasonable to merge this since it enhance/fixes an existing feature. The only concern we need to address first is browser compatibility, as Safari doesn’t appear to support streams for the response body.

I have tested it, and this function works normally on Safari. Although I don’t have more devices for further testing, I believe the Fetch API should be supported by all major browsers.
[Browser Compatibility Document](https://developer.mozilla.org/en-US/docs/Web/API/TextDecoderStream#browser_compatibility)

ramiresviana:
The incompatibility comes from the `response.body` property—see the *body is a readable byte stream* note in the [browser compatibility table](https://developer.mozilla.org/en-US/docs/Web/API/Response/body#browser_compatibility). The solution is to check whether `body` is a `ReadableStream`, and if not, fall back to waiting and processing the full response.

manx98:
> The incompatibility comes from the `response.body` property—see the _body is a readable byte stream_ note in the [browser compatibility table](https://developer.mozilla.org/en-US/docs/Web/API/Response/body#browser_compatibility). The solution is to check whether `body` is a `ReadableStream`, and if not, fall back to waiting and processing the full response.

I believe the compatibility mentioned here refers to the read/getReader methods owned by the body property, rather than the pipeThrough method I am using. Additionally, I can actually access [the target/content] normally when using the Safari browser.

[ReadableStream/pipeThrough browser_compatibility](https://developer.mozilla.org/en-US/docs/Web/API/ReadableStream/pipeThrough#browser_compatibility)

manx98:
> The incompatibility comes from the `response.body` property—see the _body is a readable byte stream_ note in the [browser compatibility table](https://developer.mozilla.org/en-US/docs/Web/API/Response/body#browser_compatibility). The solution is to check whether `body` is a `ReadableStream`, and if not, fall back to waiting and processing the full response.

The compatibility issue you mentioned is not the same as the pipeThrough method I used. Below is a complete compatibility list for ReadableStream features. Moreover, I have already tested it on a physical Safari device, and the API calls are functioning normally. The only drawback is that the TextDecoderStream used requires a relatively recent browser version.

[ReadableStream browser compatibility](https://developer.mozilla.org/en-US/docs/Web/API/ReadableStream#browser_compatibility)

[TextDecoderStream browser compatibility](https://developer.mozilla.org/en-US/docs/Web/API/TextDecoderStream#browser_compatibility)

hacdias:
From what I understand, Safari does not support a readable **byte** stream, so this code should work properly:

https://github.com/mdn/browser-compat-data/issues/22831
https://bugs.webkit.org/show_bug.cgi?id=250549

I think it's a good idea to get this in. If you have time to update the PR, I will take a look at it.

PR #4878, opened 2025-05-29
feat: Implements copy-on-write for TUS uploads

hippolin:
**Description**
This pull request enhances the `tusPatchHandler` by introducing a copy-on-write mechanism. When handling file uploads, a temporary file is created in the `/tmp` directory. This approach ensures that existing file content is preserved during uploads, minimizing the risk of data loss. This change addresses an issue encountered when using `awslabs/mountpoint-s3`, where uploads resulted in an "Operation not permitted" error. By implementing the copy-on-write strategy, we can handle file uploads more reliably.

Key changes include:
- Creation of a temporary file for upload operations.
- Copying existing file content to the temporary file if the upload offset is greater than zero.
- Writing new content to the temporary file and then copying it to the destination file.

This change improves the reliability of file uploads in scenarios where uploads may be interrupted or resumed.

- [x] DO make sure you are making a pull request against the **master branch** (left side). Also, you should start *your branch* off *our master*.

**Further comments**
This change introduces a copy-on-write mechanism in the `tusPatchHandler` to enhance the reliability of file uploads, especially when using `mountpoint-s3`. The primary motivation for this approach was to address the "Operation not permitted" error that users encountered during uploads.

I considered several alternatives, such as directly writing to the destination file or using a different temporary storage strategy. However, these options posed risks of data loss or corruption, especially in scenarios where uploads could be interrupted. The copy-on-write method allows us to safely manage existing file content while appending new data, ensuring that uploads can be resumed without issues.

Additionally, using a temporary file in the `/tmp` directory simplifies the cleanup process and reduces the likelihood of leaving behind orphaned files in the filesystem.


ramiresviana:
Thanks for the contribution! Unfortunately, this is unlikely to be merged, since the issue it addresses is very specific and doesn’t affect most users. If you’d like to push it forward, I recommend opening an issue that clearly describes the problem—gathering feedback from other users will increase the chances of it being considered.

Regarding the proposed changes, there are a few concerns:

* The default temporary path can vary depending on the system configuration (e.g., Windows).
* The temporary path may not have enough space available (e.g., a small system disk with a larger disk exposed to Filebrowser).
* The destination file is rewritten too many times. Uploading 100MB in 1MB chunks would result in gigabytes of disk writes.

ramiresviana:
I'm closing this due to lack of feedback regarding the raised issues.

PR #4907, opened 2025-06-13
chore: updated readme and templates

hacdias:
This PR:

- Updates README to make clear that project is in maintenance-only mode and what that means.
- Disables blank issues. Only bug reports can be filed and steers users to Discussions for other things.
- Updates PR template.

@jniggemann could you take a look?


jniggemann:
> @jniggemann could you take a look?

Looks good :+1:

PR #5173, opened 2025-06-21
feat: add "copy inline link to clipboard" button to Share prompt

kosmosac:
## Description

Whenever a user wanted to share a file for his own use (e.g. to embed it in a forum post), they had to go through the download page to get the inline link or edit the link manually. This adds a 3rd button in the share prompt instead. 

## Additional Information

Might not be the ideal way to do this though. The "open file" link on the download page is shorter but my knowledge of JS and GO is still too limited to implement that.

## Checklist


hacdias:
Hey! I think this can be a good idea, but I want to keep https://github.com/filebrowser/filebrowser/issues/5239 in mind. As such, if you are still interested in adding this option, please update the PR and I will look at it.

kosmosac:
> Hey! I think this can be a good idea, but I want to keep #5239 in mind. As such, if you are still interested in adding this option, please update the PR and I will look at it.

I have reimplemented my proposal, this time even with the "short" link format I originally intended.
In my testing accessing the inline link for a password protected file ( #5239 ) ends up in a 401 Unauthorized.

frontend/src/components/prompts/Share.vue
```diff
@@ -32,6 +33,16 @@
...
+            <td class="small">
+              <button
+                class="action copy-clipboard"
+                :aria-label="$t('buttons.copyInlineToClipboard')"
+                :title="$t('buttons.copyInlineToClipboard')"
+                @click="copyToClipboard(buildInlineLink(link))"
+              >
+                <i class="material-icons">insert_link</i>
+              </button>
+            </td>
```
hacdias, line 46: I think we should do a few things before merging this:

1. Add a new label to indicate that this a direct link, maybe "Copy direct download link to clipboard"
2. Maybe change the icon to a download icon
3. Not show this button if the share is password-protected, otherwise it'll be confusing

kosmosac: > I think we should do a few things before merging this:
> 
> 1. Add a new label to indicate that this a direct link, maybe "Copy direct download link to clipboard"
> 2. Maybe change the icon to a download icon
> 3. Not show this button if the share is password-protected, otherwise it'll be confusing

1 & 2 is simple as its just changing text and icon name, I'll do that right away in my fork.

I agree with 3, even thought about it before renewing the PR, but the only way to do this right I can think of is to extend the api. And thats way out of my comfort zone.

Because as far as I can see even the download page does not really "check" if a file is password protected before trying to present it. Instead ist tries to get the details of the file, and if it runs into a 401 it will present a password prompt.

So we could either implement the check using the same "logic", which would include importing `fetch` from the pub api in addition to the already used share api, and implementing that whole "try - except" logic used in the download page.
Or we could extend the share api to get the logic needed (might even include extending the backend).

Or is there any other way I can't think of right now?

hacdias: I can take over this PR :)

hacdias (approved):
Thanks for this. I updated the PR to reuse a string we already had (Copy download link), updated the icon and made the button disabled for password-protected shares. I will merge!

Image

PR #5225, opened 2025-06-28
fix: add configurable minimum password length

hacdias:


settings/settings.go
```diff
@@ -10,23 +10,25 @@ import (
 )
 
 const DefaultUsersHomeBasePath = "/users"
+const DefaultMinimumPasswordLength = 12
```
evrial, line 13: Why this even a thing? Have you thought about local installations, this is so stupid.

hacdias: 1. This is configurable via either flags, environment variable, or via the interface. 
2. A local installation should also have a reasonable secure password. Local doesn't meant secure per definition.
3. More context: https://github.com/filebrowser/filebrowser/security/advisories/GHSA-cm2r-rg7r-p7gg

evrial: I still don't understand why forcing this issue to the end user. Not even Google, Amazon or Pi-hole has this limit

hacdias: Both Amazon and Google have password requirements. As said before, all of it is customizable. There's no problem here. It's open source and free software. It comes as it is.

PR #5235, opened 2025-06-29
fix: Move folder can be moved into itself causing an error

### LINKED ISSUE #3892: Move folder can be moved into itself causing an error
slect folder A, then click move folder, then select folder A when chosing where to move it

right now this causes an error 400 bad request with a single or multiple folders

the folders you are moving should just not show up in the where to move menu

Image

FoxyHunter7:
## Description

Exclude the to-be-moved folder from the move dialog.

fixes #3892 

## Checklist


hacdias (approved):
Thank you :)

Issue #3794, opened 2025-03-11
Do not show Root directory name in <title>


github-cli:
the directory that the filebrowser uses as root (e.g. /srv) is shown in the window title to all users.
e.g. if a user is logged in, they will see "srv - Files"
if the root directory is something like /JohnDoeHome, that would expose the server owners directory to everyone.

i did not report this as a bug as I was not sure if this might be intended or not

github-cli:
just to clarify, with root directory, i mean the parent directory of the user, the one they will not see anywhere but the title of the window... for those users that dont even know about it, it can be quite confusing... please let me know if I should elaborate or provide screenshots

github-cli:
Why? Has this project been abandoned and is no longer receiving bugfixes?

github-cli:
Any chance this will be looked at?

mboskamp:
I stumbled over the same thing. I think what is displayed in the window title is the name of the current directory. When logging in with the admin user I also see the `srv` directory displayed in the window title (`srv - Files - File Browser`). The title changes based on the current directory, so navigating to a subdirectory called `test` will change the window title to `test - Files - File Browser`.
I think this behavior is fine and I would think the devs would call it expected.

As a workaround, you could change the name of the root directory to something more readable or change the scope of the user in the config.

I hope this helps. :)

github-cli:
Yes, it is the name of the current folder.
I actually use filebrowser for people to be able to edit their Hugo generated websites, so renaming to something user specific or non confusing isn’t always possible.
Maybe something like being able to set static titles per user would be an option?
I guess my use case isn’t the most common, if it were just a file share, it wouldn’t be so confusing.

hacdias:
Action here is changing the display name of the directory before sending it to the UI.

github-cli:
woohoo! thanks :-)

PR #5259, opened 2025-07-03
fix: lookup directory name if blank when downloading shared directory

### LINKED ISSUE #5255: Folder Sharing Packaging Download Naming Error
### Checklist

### Version

```Text
V2.36.0

version amd64 centos7
```

### Description

Share a folder via a link and then download the package of this folder, the name of the package is always fixed to “ZIP”.

Manually rename the package to “XXX.ZIP” and it unpacks fine, it just looks like a naming error.

fixes

### What actually happened?

bug

### Reproduction Steps

bug

### Files

_No response_

Rmiller5466:
## Description

When downloading a shared directory, the fileinfo was being overwritten to where the directory in question would become the new fs root.  This lead to file.Name = "" and file.Path = "/", hence the bug of downloading a file with the name ".zip" ("" + ".zip").  This has been fixed by doing an fs.Stat(".") to get the current directories name under the circumstances where file.Name is empty.  

## Checklist

closes #5255

- [X] I am aware the project is currently in maintenance-only mode. See [README](https://github.com/filebrowser/community/blob/master/README.md)
- [X] I am aware that translations MUST be made through [Transifex](https://app.transifex.com/file-browser/file-browser/) and that this PR is NOT a translation update
- [X] I am making a PR against the `master` branch.
- [X] I am sure File Browser can be successfully built. See [builds](https://github.com/filebrowser/community/blob/master/builds.md) and [development](https://github.com/filebrowser/community/blob/master/development.md).

### DIFF
--- a/http/raw.go
+++ b/http/raw.go
@@ -182,9 +182,10 @@ func rawDirHandler(w http.ResponseWriter, r *http.Request, d *data, file *files.
 			name = file.Name
 		} else {
 			// This should indicate that the fs root is the directory being downloaded, lookup its name
-			actual, err := file.Fs.Stat(".")
-			if err != nil {
-				return http.StatusInternalServerError, err
+
+			actual, statErr := file.Fs.Stat(".")
+			if statErr != nil {
+				return http.StatusInternalServerError, statErr
 			}
 			name = actual.Name()
 		}
--- a/http/raw.go
+++ b/http/raw.go
@@ -177,7 +177,17 @@ func rawDirHandler(w http.ResponseWriter, r *http.Request, d *data, file *files.
 
 	name := filepath.Base(commonDir)
 	if name == "." || name == "" || name == string(filepath.Separator) {
-		name = file.Name
+		// Not sure when/if this will ever be true, though kept incase there is an edge-case where it is
+		if file.Name != "" {
+			name = file.Name
+		} else {
+			// This should indicate that the fs root is the directory being downloaded, lookup its name
+			actual, err := file.Fs.Stat(".")
+			if err != nil {
+				return http.StatusInternalServerError, err
+			}
+			name = actual.Name()
+		}
 	}
 	// Prefix used to distinguish a filelist generated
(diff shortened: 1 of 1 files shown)


http/raw.go
```diff
@@ -177,7 +177,18 @@ func rawDirHandler(w http.ResponseWriter, r *http.Request, d *data, file *files.
...
+			name = file.Name
+		} else {
+			// This should indicate that the fs root is the directory being downloaded, lookup its name
+
+			actual, statErr := file.Fs.Stat(".")
+			if statErr != nil {
+				return http.StatusInternalServerError, statErr
+			}
+			name = actual.Name()
+		}
```
hacdias, line 191: I'm thinking about this. It is possible to download multiple directories at once, and I think that is one of the cases where an empty name is happening.

I think we should use a static name here, maybe just `download`.

I want to avoid showing the root directory for this reason here: https://github.com/filebrowser/filebrowser/issues/3794

Rmiller5466: It's important to distinguish the two ways downloads can be done (atleast that I'm aware of): 

1) selecting file(s) directly in the UI and clicking the download button 
2) selecting a single file and using the "share with" option.

This empty name bug was only occurring on that second method.  When downloading multiple directories / files in the UI I am observing it being saved as "\_{PARENT_DIR}.zip".  When downloading a single file it is downloading as "{FILE_NAME.zip}".  To be consistent with the default (non-share) download button UI, this could also be updated to "\_{PARENT_DIR}.zip", or to your point about avoiding the root directory we could change both of these to be a generic "download.zip".  Regardless I think we should do the same for each for consistency.  

I personally lean more towards leaving the parent directory name as the default download title - as a user I feel it's more immediately clear that I'm downloading what I expected rather than having to unzip / preview the file after it has finished to ensure I didn't download the wrong thing.

hacdias: Let's keep it as-is then. It can always be changed later.

PR #5260, opened 2025-07-03
feat: prevent page change if there are outstanding edits

### LINKED ISSUE #2038: Avoid losing ongoing edits when navigating page by alerting user

Rmiller5466:
## Description

Addresses the linked issue - adds an event listener on the file editor that triggers the browser default "unsaved changes" confirmation dialog if the user tries to exit or refresh the page while there are outstanding unsaved changes to a file.

`event.returnValue` is now deprecated, though kept to ensure legacy browser support.  Inline comment and link to documentation included.  

## Checklist

closes #2038

### DIFF
--- a/frontend/src/views/files/Editor.vue
+++ b/frontend/src/views/files/Editor.vue
@@ -53,7 +53,7 @@ import { getTheme } from "@/utils/theme";
 import { marked } from "marked";
 import { inject, onBeforeUnmount, onMounted, ref, watchEffect } from "vue";
 import { useI18n } from "vue-i18n";
-import { useRoute, useRouter } from "vue-router";
+import { onBeforeRouteUpdate, useRoute, useRouter } from "vue-router";
 
 const $showError = inject<IToastError>("$showError")!;
 
@@ -129,6 +129,15 @@ onBeforeUnmount(() => {
   window.removeEventListener("wheel", handleScroll);
   window.removeEventListener("beforeunload", handlePageChange);
   editor.value?.destroy();
+});
+
+onBeforeRouteUpdate((to, from, next) => {
+  if (!editor.value?.session.getUndoManager().isClean()) {
+    layoutStore.showHover("discardEditorChanges");
+    next(false);
+  } else {
+    next();
+  }
 });
 
 const keyEvent = (event: KeyboardEvent) => {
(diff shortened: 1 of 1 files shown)


hacdias:
@Rmiller5466 this is a good start!

I would also want to show the the prompt when going to another page. I think using `onBeforeUnmount` should do the trick.  think the `close.` function will also need to be reworked a bit to work correctly when navigating to another page.

Rmiller5466:
Thanks!

Could you elaborate more on this?  This change currently protects from page navigation (forward/back buttons), the refresh button, and window closing.  In my testing, the "unsaved changes" confirmation was showing correctly when navigating away from the editor with unsaved changes; could you give a specific case that I can test to better understand the desired behavior?

hacdias:
@Rmiller5466 in my tests, this only works when refreshing. When using the browser's back and forward buttons, I still do not get the prompt. This is because they do not trigger the `unload` event, because it's handled by Vue's navigation and not the browser itself.

You can see in the video below that refreshing triggers the prompt, but navigation doesn't.

https://github.com/user-attachments/assets/b4004d57-e014-49d5-be22-88929333e90a

hacdias:
Maybe you can use the `beforeRouteLeave`: https://router.vuejs.org/guide/advanced/navigation-guards.html#Using-the-Options-API

Rmiller5466:
Gotcha - I follow now.  I tested refresh and closing the window but I indeed seem to have missed testing the back button.  I could have sworn I tested the buttons though I guess not!  I'll take a look at this as soon as I can!

Rmiller5466:
Sorry that took so long for me to circle back to!  When I first tried to implement that `beforeRouteLeave` I wasn't able to get it working and had to put it down for a bit as life got in the way.  Realized `beforeRouteUpdate` might be needed instead and was able to get it working as intended right away.  Let me know if there is anything else I've missed!

hacdias:
Thank you so much!

PR #5269, opened 2025-07-06
fix: Move and Copy file conflict check race condition

### LINKED ISSUE #3312: move file to dst with same name: modal window disappears on its own
* FB 2.30
* Reproduce as follows:
  * Folder A has a file test.txt
  * Folder B has a file test.txt
  * Go into folder B, mark file test.txt and try to move it to folder A. The modal window (see screenshot below) very briefly appears and immediately disappears again without user interaction
* Can reproduce with Chrome and Edge
* No error message on browser console

image

jagadam97:
## Description

This PR introduces a fix for race condition that happens in move or copy file conflict check by converting the check function to promised based. this also makes the async flow more explicit.

##Proof of Testing

https://github.com/user-attachments/assets/bcdc5d0e-b995-44d4-96de-7ceaa5233bcf

## Additional Information
I have tested this by adding console log and the `upload.checkConflict` sometimes returns false. then this conflict popup is not shown and replace method is used to copy / Move the file.

if i add console.log before `upload.checkConflict` then the race condition is not at all reporducible.

Closes #3312 

## Checklist

### DIFF
--- a/frontend/src/components/prompts/Copy.vue
+++ b/frontend/src/components/prompts/Copy.vue
@@ -80,7 +80,7 @@ export default {
   },
   methods: {
     ...mapActions(useLayoutStore, ["showHover", "closeHovers"]),
-    copy: async function (event) {
+    copy: function (event) {
       event.preventDefault();
       const items = [];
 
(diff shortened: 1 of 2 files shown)


hacdias:
The closing modals issue seems to have been fixed in https://github.com/filebrowser/filebrowser/pull/3537, and I cannot reproduce it.

I don't see how changing the method from async/await to a Promise would solve it the issue.

Thanks for the contribution, but I'm not accepting this at the moment, since I can't reproduce the issue, or see how this refactor would even solve the issue if it existed.

jagadam97:
@hacdias 

Ho ok. i was able to reproduce the above issue. but very rarely . so i taught it was race condition. anyway we can close the PR if issue is not reproducible

PR #5270, opened 2025-07-07
fix: drop modify permission for uploading new file

### LINKED ISSUE #2993: Upload Event Command Runner Not Triggered

### LINKED ISSUE #5202: Upload fails when user has Create but not Modify permissions
### Checklist

### Version

```Text

```

### Description

When uploading a file:

The initial POST request succeeds, and the file is created.

The subsequent PATCH request fails with 403 Forbidden.

As a result, the uploaded file remains in the system as a 0-byte file.

After checking the code, it appears that the PATCH handler checks for modify permission, and denies access if it's not granted. However, since uploading is a two-phase process (POST to create, then PATCH to write), this breaks the upload flow for users who are allowed to create but not modify.

_No response_

### What actually happened?

_No response_

### Reproduction Steps

_No response_

### Files

_No response_

ramiresviana:
The modify permission is currently required, as the upload process consists of creating a new file using the POST endpoint and then writing its data via the PATCH endpoint. To remove the need for the modify permission, we need to ensure that the PATCH endpoint is used only for uploading new files—thus preventing malicious use of the PATCH endpoint to edit existing files.

The current upload system has several issues: it is not fully compliant with the TUS protocol, it is missing features that were previously available, and the old implementation is still present in the codebase.

In an effort to bring the upload functionality back to at least a usable state, I’m keeping the scope and number of changes in this PR to a minimum. I've opted to use in-memory tracking of new file uploads. While this is far from ideal, it gets the job done. Users with the create permission will now be able to use the PATCH endpoint after the initial POST request.

The behavior of the override parameter didn’t make sense to me. It was destructive—modifying files in an unpredictable way and without user knowledge. I took the opportunity to fix this by restoring its original behavior.

Now that we can track uploads, I’ve also added the upload event.

fixes #5202
fixes #2993


hacdias:
I will take a look at this during this week. If this PR does what it promises, it also supersedes #3899.

http/tus_handlers.go
```diff
@@ -148,12 +208,26 @@ func tusPatchHandler() handleFunc {
 			return http.StatusInternalServerError, fmt.Errorf("could not write to file: %w", err)
 		}
 
-		w.Header().Set("Upload-Offset", strconv.FormatInt(uploadOffset+bytesWritten, 10))
+		newOffset := uploadOffset + bytesWritten
+		w.Header().Set("Upload-Offset", strconv.FormatInt(newOffset, 10))
+
+		if newOffset >= uploadLength {
+			unregisterUpload(file.RealPath())
```
hacdias: This is the only place where we call `unregisterUpload`. What happens if the upload fails/is canceled?

ramiresviana: Good point, i forgot to handle uploads that were gracelly terminated the frontend. I can implement `tusDeleteHandler `to handle the case where the frontend cancels the upload.

In the event of an upload failure, I assume the user would retry, making it to naturally unregister, effectively avoiding any negative impact. Properly handling upload failures would require implementing some form of background processing, given the limited benefit in our current context, I’ve decided to skip it for now. However, if desirable, i can give it a try. One approach I’ve considered is using goroutines to count down the timeout.

ramiresviana: Good point, i forgot to handle uploads that were gracelly terminated the frontend. I can implement `tusDeleteHandler` to handle the case where the frontend cancels the upload.

In the event of an upload failure, I assume the user would retry, making it to naturally unregister, effectively avoiding any negative impact. Properly handling upload failures would require implementing some form of background processing. Given the limited benefit in our current context and seeing that even the current implementation doesn't handle failures, I’ve decided to skip it for now.

hacdias: A possibility would be to use some sort of caching like [otter](https://pkg.go.dev/github.com/maypok86/otter/v2@v2.2.0), with automatic deletion of entries if there haven't been any update after _n_ time

ramiresviana: It looks like otter is what we need, but it requires Go version 1.24, whereas the current project is using 1.23. What do you thing about using an alternative like [ristretto](https://pkg.go.dev/github.com/dgraph-io/ristretto) or [ttlcache](https://pkg.go.dev/github.com/jellydator/ttlcache) which, by the way, appear to be more widely adopted?

hacdias: That's also fine.

frontend/src/api/tus.ts
```diff
@@ -207,6 +205,9 @@ export function abortAllUploads() {
     }
     if (CURRENT_UPLOAD_LIST[filePath].upload) {
       CURRENT_UPLOAD_LIST[filePath].upload.abort(true);
+      CURRENT_UPLOAD_LIST[filePath].upload.options!.onError!(
+        new Error("Upload aborted")
```
ramiresviana, line 201: After aborting an upload, the confirmation prompt is being wrongly displayed when closing the browser tab. This change fixes this issue by rejecting the upload promise, which was hanging previously.

hacdias:
Will take a look at this tmr

hacdias (commented):
Very happy with this so far, thank!

I took it out for a spin but I noticed that when you upload multiple files at once, the progress bar above and the progress where you can the % uploaded goes back and forth.

I think there may be some issue on the frontend that is miscalculating the totals. Could you take a look at that?

frontend/src/stores/upload.ts
```diff
@@ -168,7 +159,7 @@ export const useUploadStore = defineStore("upload", {
       this.processUploads();
     },
     finishUpload(item: UploadItem) {
-      this.setProgress({ id: item.id, loaded: item.file.size > 0 });
+      this.setProgress({ id: item.id, loaded: item.file.size });
```
ramiresviana, line 162: The upload size was being discarded after finishing, making the total progress to go backwards. The size will now be retained, allowing the total progress to be properly calculated.

hacdias (approved):

Nmrawe:
Upgraded to v2.39.0 and now uploads give zero-byte files. This fix might be the culprit.

ramiresviana:
@Nmrawe thank you for trying the new update! Could you please create a issue with much detail as possible so we can work on a fix? Also, are you by any chance using a custom base url?"

iesv:
> 升级到 v2.39.0 后，上传文件时会显示零字节。这可能是此修复导致的。

I have the same.

Nmrawe:
@ramiresviana Fixed in v2.40.0. Thanks!

PR #5280, opened 2025-07-12
feat: Improve Docker entrypoint and config handling

outlook84:
## Description

- Simplify Dockerfile ENTRYPOINT to delegate config handling to init.sh.
- Enhance init.sh to automatically add `--config=/config/settings.json` if no config argument is provided.

## Additional Information

This would enable container user to set configfile via --config arg.

## Checklist

### DIFF
--- a/docker/alpine/init.sh
+++ b/docker/alpine/init.sh
@@ -11,7 +11,7 @@ fi
 has_config_arg=0
 for arg in "$@"; do
   case "$arg" in
-  --config=*)
+  --config|--config=*|-c|-c=*)
     has_config_arg=1
     break
     ;;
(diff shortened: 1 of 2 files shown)


docker/alpine/init.sh
```diff
@@ -7,4 +7,19 @@ if [ ! -f "/config/settings.json" ]; then
   cp -a /defaults/settings.json /config/settings.json
 fi
 
-exec "$@"
+# Deal with the case where user does not provide a config argument
+has_config_arg=0
+for arg in "$@"; do
+  case "$arg" in
+  --config=*)
```
hacdias: A few things that need to be improved here:

- `-c` is also possible as a short version of `--config`
- What if the user uses `--config /path`? There's no `=` and it stillx would work, but now we'd set it twice

hacdias (changes requested):

docker/alpine/init.sh
outlook84: OK, I change script to match `--config|--config=*|-c|-c=*`.

I think this should cover all cases now.

hacdias (approved):

PR #5281, opened 2025-07-12
build: improve docker image and binary sizes

### LINKED ISSUE #5215: Docker images should be slimmer
### Checklist

### Version

```Text
All
```

### Description

> Container images should be as slim as possible to reduce resources and the attack surface. The current Filebrowser image contains files (like the apk package cache or the curl and jq package) which are not necessary for operation. This is especially problematic in conjunction with the security problems of the command execution feature. The default image should be replaced with a distroless build and Alpine image should be stripped of all files not strictly required.
 
Thanks @mtausig for the report.

_No response_

### What actually happened?

_No response_

### Reproduction Steps

_No response_

### Files

_No response_

jagadam97:
## Description

This PR introduces improvements to docker release build sizes.
Improvements that help in achieving small docker sizes are
1. using upx to shrink the binary size of filebrowser
2. removing curl command from health check and using wget which is included in alpine image

the final docker build file using these above steps is 27.2 MB compared to initial 69.6MB
image

## Additional Information

closes #5215 
removing `jq` gives only benifit of 1.3 MB and I felt like its not work the effort to replicate what `jq` is doing with sed or `grep`

## Checklist

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -85,6 +85,8 @@ jobs:
           node-version: "22.x"
           cache: "pnpm"
           cache-dependency-path: "frontend/pnpm-lock.yaml"
+      - name: Install upx
+        run: sudo apt-get install -y upx
       - name: Set up QEMU
         uses: docker/setup-qemu-action@v1
       - name: Set up Docker Buildx
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -40,6 +40,17 @@ archives:
       - goos: windows
         formats: ["zip"]
 
+upx:
+  - enabled: true
+    goos:
+      - linux
+      - darwin
+    goarch:
+      - amd64
+      - arm64
+    compress: "best"
+    lzma: true
+
 dockers:
   # Alpine docker images
   - dockerfile: Dockerfile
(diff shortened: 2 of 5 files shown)


hacdias (approved):
These seem to be good improvements. I am a bit on the fence about `upx`, but it seems to be wildly used. If we end up seeing problems, we can always change it back.

outlook84:
upx is commonly seen in embeded system like router for a reason.
It needs to decompress file to ram, so it trades ram for storage space and has slower start.

On normal computing devices, ram is a more scarce resource than storage space. 

Replace alpine with busybox as base-image, image shrinks to less than 20MB without upx.

图片

hacdias:
If you make a PR to change to Busy Box (and remove upx), I will take a look at it.

outlook84:
> If you make a PR to change to Busy Box (and remove upx), I will take a look at it.

Busybox won't have tini. And I don't like the idea of copying third-party static binary into container.
So if init is needed, users need to set --init=true via docker/podman.

Is this acceptable?

outlook84:
> If you make a PR to change to Busy Box (and remove upx), I will take a look at it.

Let's talk about added packages in alpine container since busybox doesn't have a package manager.
`ca-certificates mailcap jq tini`
ca-certificates and mailcap can be copied into busybox image.
jq can be replaced by busybox sed/grep.
I don't think tini is needed anymore, Docker introduced --init option in 2017.

Issue #5289, opened 2025-07-13
Get "filebrowser  | ln: /tmp/FB_CONFIG: Read-only file system" after update


knubew:
### Checklist

- [x] This is a bug report, not a question.
- [x] I have searched on the [issue tracker](https://github.com/filebrowser/filebrowser/issues?q=is%3Aissue) for my bug.
- [x] I am running the latest [FileBrowser version](https://github.com/filebrowser/filebrowser/releases) or have an issue updating.

### Version

```Text
2.39.0
```

### Description

After pulling and updating the Filebrowser Compose File from 2.38.0 to 2.39.0 (in OpenMediaVault 7.7.11-1) I cannot start filebrowser anymore. The log shows multiple identical entries: filebrowser  | ln: /tmp/FB_CONFIG: Read-only file system

This is the first time I have problems after pulling a new FB version.

### What did you expect to happen?

Filebrowser 2.39.0 should start.

### What actually happened?

Filebrowser 2.39.0 does not start.
The logfile shows:
filebrowser  | ln: /tmp/FB_CONFIG: Read-only file system

### Reproduction Steps

none

outlook84:
Do you add --read-only flag to docker run? I overlooked the possibility that someone would set it up this way.

Normally this shouldn't happen.  And this shouldn't prevent filebrowser to start, I'll work out a fix.

knubew:
> Do you add --read-only flag to docker run? I overlooked the possibility that someone would set it up this way.
> Normally this shouldn't happen. And this shouldn't prevent filebrowser to start, I'll work out a fix.

This is my composer file:

---
services:
  filebrowser:
    image: filebrowser/filebrowser:latest
    container_name: filebrowser
    user: "1003:100"  
    read_only: true   
    cap_drop:
      - ALL           
    ports:
      - "82:80"
    volumes:
      - /srv/dev-disk-by-uuid-XXXXXXXX-XXXX-XXXX-XXXX-XXXXXXXXXXXX/MyCloud:/srv/MyCloud
      - /data/filebrowser/config:/config
      - /data/filebrowser/config:/database
    networks:
      - cloudnet
    restart: unless-stopped

networks:
  cloudnet:
    external: true

hacdias:
Should be fixed in 2.40.0 which is being released at the moment

knubew:
v2.40.0 works fine. Thank you.

PR #5291, opened 2025-07-13
fix: ln command fail workaround for read-only container 

outlook84:
## Description

https://github.com/filebrowser/filebrowser/issues/5289

I'm sorry that I overlooked the possibility of read-only container. 
I couldn’t find a way to pass the config argument to the healthcheck script in a read-only container, so I ensure the ln command always exits successfully.

## Checklist

### DIFF
--- a/docker/alpine/healthcheck.sh
+++ b/docker/alpine/healthcheck.sh
@@ -2,8 +2,13 @@
 
 set -e
 
-PORT=${FB_PORT:-$(cat /tmp/FB_CONFIG | sh /JSON.sh | grep '\["port"\]' | awk '{print $2}')}
-ADDRESS=${FB_ADDRESS:-$(cat /tmp/FB_CONFIG | sh /JSON.sh | grep '\["address"\]' | awk '{print $2}' | sed 's/"//g')}
+CONFIG_FILE="/tmp/FB_CONFIG"
+if [ ! -f "/tmp/FB_CONFIG" ]; then
+  CONFIG_FILE="/config/settings.json"
+fi
+
+PORT=${FB_PORT:-$(cat $CONFIG_FILE | sh /JSON.sh | grep '\["port"\]' | awk '{print $2}')}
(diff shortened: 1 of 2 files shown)


hacdias:
Another idea could be to just always assume the config file to be where we intended it to be (/config/settings.json). 

By adding more options, we're just making it more likely that misconfigurations happen. 

Wdyt?

outlook84:
> Another idea could be to just always assume the config file to be where we intended it to be (/config/settings.json).
> 
> By adding more options, we're just making it more likely that misconfigurations happen.
> 
> Wdyt?

To be frank, I think it's the healthcheck that complicates things.

I started to use filebrowser container when port in healthcheck is hard-coded, I have to disable healthcheck becaue 80 port is used by caddy in the same podman pod.

Then healthcheck needs to read config file for port, so config file path needs to be the same place.

Maybe if there is no build-in healthcheck, these problems will go away? There are lots of options to setup healthcheck externally.

I know nothing. If maintainers decide healthcheck is important, I will revert to hard-coded config file path.

hacdias:
I'd argue the problem is a lot of people using the Docker image in a different way than it is intended. 

The Docker image is supposed to be a way to run the service with minimal configuration and reproducibility. Hence, there's an expected place for the configuration (which is completely optional) and the database.

When people start introducing variability and wanting to put these config and database somewhere else for no reason, then the problems begin. The image has never been design to be used for maximum configurability. If someone wants that, they can make their own image and build file browser from source. 

It is common to set a health check in a Dockerfile, and having it built in makes it easier for most users. 

I'd rather have everything streamlined with an expected path instead of letting to all be configurable, than removing the health check

hacdias:
I'm now releasing 2.40.0 which reverts the `ln` and keeps the hardcoded `/config/settings.json` in the health check, which means it'll just keep the behavior of previous versions.
 
If more people need this, with good argument, we can think about adding it.

outlook84:
Upon reflection, I regret the disruption caused by my implementation.

PR #5292, opened 2025-07-13
fix: invalid path when uploading files

ramiresviana:
Trying to upload a file to an instance running on Windows — or when baseurl is configured — results in errors and empty files. This issue stems from the `tusPostHandler`, which uses an absolute path for the Location header. Unfortunately, the `tus-js-client` library doens't handle absolute paths properly and causes the request url to have an additional slash:
```
http://localhost:8080/api/tus//example/file.jpg
```
While Linux handles that without issues, Windows returns an error:
```
404 127.0.0.1 lstat //example/file.jpg: file does not exist
```
Additionally, the baseurl configuration was being ignored, leading to further problems. The introduced changes will ensure that the upload url is now properly constructed and follows the expected format:
```
http://localhost:8080/baseurl/api/tus/example/file.jpg
```

### DIFF
--- a/http/tus_handlers.go
+++ b/http/tus_handlers.go
@@ -147,7 +147,8 @@ func tusPostHandler() handleFunc {
 		// Enables the user to utilize the PATCH endpoint for uploading file data
 		registerUpload(file.RealPath(), uploadLength)
 
-		w.Header().Set("Location", "/api/tus/"+r.URL.Path)
+		// Signal the frontend to reuse the current request URL
+		w.Header().Set("Location", "")
 
 		return http.StatusCreated, nil
 	})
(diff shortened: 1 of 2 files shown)


hacdias (approved):

http/tus_handlers.go
```diff
@@ -147,7 +147,8 @@ func tusPostHandler() handleFunc {
 		// Enables the user to utilize the PATCH endpoint for uploading file data
 		registerUpload(file.RealPath(), uploadLength)
 
-		w.Header().Set("Location", "/api/tus/"+r.URL.Path)
```
hacdias, line 150: I think we should add this back. The problem from the double slash is that r.URL.Path contains the slash, or not? That creates the double slash issue.

hacdias: @ramiresviana maybe `"/api/tus" + r.URL.Path` would do the trick

PR #5302, opened 2025-07-17
fix: Location header on TUS endpoint

hacdias:
I think this should correctly fix the issue with the uploads creating empty files described in https://github.com/filebrowser/filebrowser/discussions/5299.

On https://github.com/filebrowser/filebrowser/pull/5292, we changed the `Location` header to an empty string because there was a double slash in the path. However, that was because we were concatenating `/api/path/` (with slash) with `r.URL.Path` which is prefixed with the slash.

### DIFF
--- a/http/tus_handlers.go
+++ b/http/tus_handlers.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"io"
 	"net/http"
+	"net/url"
 	"os"
 	"path/filepath"
 	"strconv"
@@ -147,8 +148,12 @@ func tusPostHandler() handleFunc {
 		// Enables the user to utilize the PATCH endpoint for uploading file data
 		registerUpload(file.RealPath(), uploadLength)
 
-		w.Header().Set("Location", "/api/tus"+r.URL.Path)
+		path, err := url.JoinPath("/", d.server.BaseURL, "/api/tus", r.URL.Path)
+		if err != nil {
+			return http.StatusBadRequest, fmt.Errorf("invalid path: %w", err)
+		}
 
+		w.Header().Set("Location", path)
 		return http.StatusCreated, nil
 	})
 }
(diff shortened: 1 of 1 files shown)


allixx:
Maybe url joining should be done in more robust way instead of naive string concatenation?

Like

```
package main

import (
	"fmt"
	"net/url"
	"os"
)

func main() {
	u, err := url.JoinPath("/api/tus", "bar/")
	if err != nil {
		os.Exit(1)
	}
	fmt.Println(u)
}
```

This will give correct results regardless of leading slash presence or absence in joining value.

hacdias:
That's true, but I don't see the need since `r.URL.Path` is guaranteed to be the same as the URL, stripped of the `/api/tus`, since that's the path matcher for this HTTP route.

http/tus_handlers.go
```diff
@@ -147,8 +147,7 @@ func tusPostHandler() handleFunc {
 		// Enables the user to utilize the PATCH endpoint for uploading file data
 		registerUpload(file.RealPath(), uploadLength)
 
-		// Signal the frontend to reuse the current request URL
-		w.Header().Set("Location", "")
+		w.Header().Set("Location", "/api/tus"+r.URL.Path)
```
ramiresviana:
```suggestion
		w.Header().Set("Location", d.server.BaseURL+"/api/tus"+r.URL.Path)
```

hacdias: Is there a need to add the server URL? Does the JS library not resolve the absolute URL in relation to the root URL?

ramiresviana: The frontend will resolve the URL by joining origin (http://localhost:8080) with the location value (/baseurl/api/tus/file.test).

hacdias: Absolutely right, thanks!

PR #5319, opened 2025-07-21
feat: Allow file and directory creation modes to be configured

### LINKED ISSUE #5316: Feature request: Allow configuring the default file/directory creation modes
### Checklist

### Version

```Text
v2.40.2
```

### Description

While I'm aware that the default user/group file **ownership** can be configured simply by running Filebrowser as a different user, the file/directory creation **modes** are not configurable at the moment. Although they can be influenced by the `umask` setting, that setting subtracts bits (bitwise AND), which is not the same as setting the creation mode.

Therefore, this is a feature request to allow configuring the file and directory creation modes.

Related discussion: #5200 

See above.

### What actually happened?

See above.

### Reproduction Steps

See above.

### Files

_No response_

williewillus:
The defaults remain the same as before.
For now, the config options are global instead of per-user.
Note also that the BoltDB creation maintains the old default mode of 0640
since it's not really a user-facing filesystem manipulation.
Fixes #5316, #5200

### DIFF
--- a/fileutils/copy.go
+++ b/fileutils/copy.go
@@ -1,6 +1,7 @@
 package fileutils
 
 import (
+	"io/fs"
 	"os"
 	"path"
 
@@ -8,7 +9,7 @@ import (
 )
 
 // Copy copies a file or folder from one place to another.
-func Copy(fs afero.Fs, src, dst string) error {
+func Copy(afs afero.Fs, src, dst string, fileMode, dirMode fs.FileMode) error {
 	if src = path.Clean("/" + src); src == "" {
 		return os.ErrNotExist
 	}
@@ -26,14 +27,14 @@ func Copy(fs afero.Fs, src, dst string) error {
 		return os.ErrInvalid
 	}
 
-	info, err := fs.Stat(src)
+	info, err := afs.Stat(src)
 	if err != nil {
 		return err
 	}
 
 	if info.IsDir() {
-		return CopyDir(fs, src, dst)
+		return CopyDir(afs, src, dst, fileMode, dirMode)
 	}
 
-	return CopyFile(fs, src, dst)
+	return CopyFile(afs, src, dst, fileMode, dirMode)
 }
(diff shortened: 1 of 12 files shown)


fileutils/copy.go
```diff
@@ -1,14 +1,15 @@
...
+	"io/fs"
 	"os"
 	"path"
 
 	"github.com/spf13/afero"
 )
 
 // Copy copies a file or folder from one place to another.
-func Copy(fs afero.Fs, src, dst string) error {
+func Copy(afs afero.Fs, src, dst string, fileMode, dirMode fs.FileMode) error {
```
williewillus, line 12: NB: Here and throughout I've had to rename these args to `afs` since the linter otherwise complains I'm conflicting with the io/fs import.

williewillus:
Vasil Rangelov ***@***.***> writes:

> boenrobot left a comment (filebrowser/filebrowser#5319)
>
> @williewillus Hello.
>
> At work, we have an instance of filebrowser 2.43.0, running in a docker container. It was previously 2.36.1, and we updated specifically for this feature.
>
> How do you actually configure these file permission settings though? I am not seeing them in the UI even though I'm an admin user. I tried to add it in settings.json, like
> ```json
> {
>   "port": 80,
>   "baseURL": "",
>   "address": "",
>   "log": "stdout",
>   "database": "/database/filebrowser.db",
>   "root": "/srv",
>   "file-mode": "0664",
>   "dir-mode": "0775"
> }
> ```
>
> but even after restarting the container, that didn't work. What am I doing wrong?

so in the JSON, the fileMode and dirMode are stored in their raw integer
form, because JSON does not have octal syntax. So it would be, in your
case:

```json
{
  "fileMode": 436,
  "dirMode": 509
}
```

This is slightly unintuitive, which is why I'd recommend using the
`filebrowser config set ...` CLI command to manipulate the config
instead of editing the json manually.

PR #5330, opened 2025-07-26
fix: show file upload errors

### LINKED ISSUE #5327: Filebrowser does not return errors correctly
### Checklist

### Version

```Text
File Browser 2.41.0
```

### Description

Hello,

The issue is that Filebrowser does not properly return errors when it is behind a proxy.

For example, I have restricted file uploads to certain extensions: .txt, .pdf, and .csv.
If I try to upload a .rpm file, the request is indeed blocked with a 403 by Filebrowser, as **expected**.

Image

However, the client is not informed — nothing happens in the browser.
The upload simply does not start, but no error message is displayed to the user.

Here is a screenshot that illustrates the issue (file upload silently fails despite being rejected with a 403).

Image

I've also observed the same behavior in other cases:

- Read-only file system

- Permission denied errors

In all these situations, Filebrowser silently fails without any UI feedback when it is behind a reverse proxy.

expected Filebrowser to properly return an error to the client when an upload fails — for example, by displaying a message such as “Upload not allowed” or “Permission denied,” so that the user understands why the action failed.

### What actually happened?

Filebrowser silently fails without any UI feedback when it is behind a reverse proxy.

### Reproduction Steps

Set up Filebrowser behind a reverse proxy (e.g. Apache or Nginx).

Configure Filebrowser to allow only certain file types for upload (e.g. .txt, .pdf, .csv).

Attempt to upload a file with a disallowed extension (e.g. .rpm).

Observe that:

Filebrowser returns a 403 Forbidden (can be seen in network logs).

But the browser shows no error or feedback — the upload silently fails.

### Files

_No response_

ramiresviana:
## Description
An error notification is now shown when a file upload fails, as the implementation was previously missing. Additionally, I removed error notifications displayed by the sidebar usage indicator. These errors were redundant and often irrelevant—for example, accessing a non-existent path would trigger an error toast, even though the file listing already displays a message. These errors were also difficult to trace, as the component is updated as a side effect of the folder navigation.

fixes #5327

## Checklist


frontend/src/api/tus.ts
```diff
@@ -55,12 +55,20 @@ export async function upload(
...
         }
         delete CURRENT_UPLOAD_LIST[filePath];
-        reject(new Error(`Upload failed: ${error.message}`));
+
+        const message =
+          error instanceof tus.DetailedError
+            ? error.originalResponse === null
+              ? "000 No connection"
+              : error.originalResponse.getBody()
+            : "Upload failed";
```
hacdias, line 69: And if it's a regular `Error`? Shouldn't we test for that and use the message in that casE?

ramiresviana: I've chosen to display a generic "Upload failed" message, since the messages from that class are quite technical ([take a look](https://github.com/search?q=repo%3Atus%2Ftus-js-client+%22throw+new+Error%22+upload.ts&type=code)). As an addition, I could print the original error on the console instead, so users can still access it if needed. What do you think?

hacdias: I think logging it to the console is a good idea. It'll make it easier when someone reports a bug.

azdfzshffg:
Hello,

up !  @hacdias  😊

thx !

azdfzshffg:
Thanks both for this change !

PR #5470, opened 2025-10-05
fix(auth): prevent integer overflow in logout timer using safeTimeout

### LINKED ISSUE #5461: Inadvertently logged out because of timeout integer overflow
### Checklist

### Version

```Text
v2.44.0
```

### Description

After deploying Filebrowser v2.44.0 I am automatically logged out and the error "You have been logged out due to inactivity." is shown.
When trying to login with the correct credentials the message is shown again. When trying to login with the wrong credentials it correctly shows "Wrong credentials".
Redeploying a previous version (v2.43.0) fixes the issue.

Expected a successful logjn.

### What actually happened?

The error "You have been logged out due to inactivity." is shown and no login happens. The call to /api/login returns 200 OK.
I might be related to commit e6c674b (https://github.com/filebrowser/filebrowser/commit/e6c674b3c616831942c4d4aacab0907d58003e23).

### Reproduction Steps

Update to current version and try to login. I am using Firefox Browser on MacOs.

### Files

_No response_

ArielLeyva:
## Description
Closes https://github.com/filebrowser/filebrowser/issues/5461
This PR introduces a new setSafeTimeout utility to replace the direct use of setTimeout when scheduling automatic user logout based on JWT expiration.

## Additional Information
- Replaced the existing setTimeout call in parseToken() with a safer implementation that avoids integer overflow issues when the delay exceeds the 32-bit limit (~24.8 days).
- The new setSafeTimeout function divides long timeouts into smaller intervals (e.g., 1 day each) and checks the remaining time recursively until the total delay has passed.
- Ensures consistent behavior across browsers, mobile WebViews, and other runtime environments that may impose shorter timeout limits.

## Checklist

### DIFF
--- a/frontend/src/api/utils.ts
+++ b/frontend/src/api/utils.ts
@@ -91,3 +91,23 @@ export function createURL(endpoint: string, searchParams = {}): string {
 
   return url.toString();
 }
+
+export function setSafeTimeout(callback: () => void, delay: number): number {
+  const MAX_DELAY = 86_400_000;
+  let remaining = delay;
+  let timerId: number;
+
+  function scheduleNext() {
+    if (remaining <= MAX_DELAY) {
+      timerId = window.setTimeout(callback, remaining);
+    } else {
+      timerId = window.setTimeout(() => {
+        remaining -= MAX_DELAY;
+        scheduleNext();
+      }, MAX_DELAY);
+    }
+  }
+
+  scheduleNext();
+  return timerId;
+}
--- a/frontend/src/api/utils.ts
+++ b/frontend/src/api/utils.ts
@@ -95,19 +95,17 @@ export function createURL(endpoint: string, searchParams = {}): string {
 export function setSafeTimeout(callback: () => void, delay: number): number {
   const MAX_DELAY = 86_400_000;
   let remaining = delay;
-  let timerId: number;
 
-  function scheduleNext() {
+  function scheduleNext(): number {
     if (remaining <= MAX_DELAY) {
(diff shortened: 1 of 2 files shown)


frontend/src/api/utils.ts
```diff
@@ -91,3 +91,21 @@ export function createURL(endpoint: string, searchParams = {}): string {
 
   return url.toString();
 }
+
+export function setSafeTimeout(callback: () => void, delay: number): number {
+  const MAX_DELAY = 86_400_000;
```
hacdias, line 96: Another idea would be to use `Number.MAX_SAFE_INTEGER`. Is there a reason not to? Then it'll be architecture dependent

ArielLeyva: It's a good idea but the value stored in the `Number.MAX_SAFE_INTEGER` constant exceeds the limit imposed by `setTimeout` (which is **32 bits** in most cases or **16 bits** in specific contexts like _microcontrollers_). In the [Node.js documentation (25.0.0)](https://nodejs.org/docs/latest/api/timers.html#settimeoutcallback-delay-args) it talks about the 32-bit limit (`2147483647`), if this value is exceeded Node.js internally sets the timer to 1 and that's why it executes instantly. I decided to use the value `86_400_000` since it is a fairly safe maximum limit even in small environments.

hacdias: Ah, you're correct. MDN says browsers do the same; https://developer.mozilla.org/en-US/docs/Web/API/Window/setTimeout#maximum_delay_value

PR #5549, opened 2025-11-15
feat: consistent flags and environment variables

hacdias:
This PR has a few goals:

- Simplify the handling of config/flags/env to use Viper in root. We were not using it before because of an issue with Viper which wasn't properly identifying when a flag was set or not (https://github.com/spf13/viper/pull/331).
- Move all `kebab-case` flags to `camelCase` in a backwards compatible way, in order to stay consistent. This is done using `SetGlobalNormalizationFunc`.
- Allows all flags as snake cased environment variables in the root command. This was possible for some, but not all options. This is done with `SetEnvKeyReplacer` by generating custom replacements for all flags. This allows transforming `camelCase` options to `UPPER_SNAKE_CASE` environment variables, not breaking compatibility. Until now, only some flags were available as env variable.
	- ⚠️ **Caveat**: we should not set two variables that would transform to the same name. Example: `foo.barCar` and `fooBar.car` would both transform to `FOO_BAR_CAR`. I added a test to ensure this doesn't happen.

@ramiresviana I would like to ask a second set of eyes on this, just to make sure I'm not breaking something. I think the change makes sense, as it allows people to easily use flags in Docker as environment variables, and strives for consistency, but I don't want to break anything.


cmd/config_init.go
```diff
@@ -34,135 +35,43 @@ override the options.`,
...
 			AuthMethod:            authMethod,
 			Defaults:              defaults,
 			Branding: settings.Branding{
-				Name:                  brandingName,
-				DisableExternal:       brandingDisableExternal,
-				DisableUsedPercentage: brandingDisableUsedPercentage,
-				Theme:                 brandingTheme,
-				Files:                 brandingFiles,
+				Name:                  v.GetString("branding.name"),
+				DisableExternal:       v.GetBool("branding.disableexternal"),
```
hacdias: ℹ️ Internally, all Viper config key names are lower cased.

cmd/config_init.go
```diff
@@ -24,145 +25,52 @@ override the options.`,
...
-			MinimumPasswordLength: minLength,
-			Shell:                 convertCmdStrToCmdArray(shell),
+			Key:                   generateKey(),
+			Signup:                v.GetBool("signup"),
+			HideLoginButton:       v.GetBool("hideloginbutton"),
+			CreateUserDir:         v.GetBool("createuserdir"),
+			MinimumPasswordLength: v.GetUint("minimumpasswordlength"),
+			Shell:                 convertCmdStrToCmdArray(v.GetString("shell")),
 			AuthMethod:            authMethod,
 			Defaults:              defaults,
```
hacdias: ℹ️ In a perfect world, we would just use `v.Unmarshal`, but unfortunately we have some custom types and I think this is easier for now to avoid breaking things.

cmd/upgrade.go
```diff
@@ -23,15 +23,15 @@ this version.`,
...
-		oldConf, err := getString(flags, "old.config")
+		oldConf, err := flags.GetString("old.config")
 		if err != nil {
 			return err
 		}
-		db, err := getString(flags, "database")
+		db, err := flags.GetString("database")
 		if err != nil {
 			return err
 		}
```
hacdias: ℹ️ This is the only public command were we're not using `python` and therefore the env variables would **not** work. See https://github.com/filebrowser/filebrowser/pull/5550

cmd/cmds_ls.go
```diff
@@ -2,6 +2,7 @@ package cmd
 
 import (
 	"github.com/spf13/cobra"
+	v "github.com/spf13/viper"
```
ramiresviana: **note:** although [discouraged](https://github.com/spf13/viper?tab=readme-ov-file#should-viper-be-a-global-singleton-or-passed-around), using viper as a global singleton made the code simpler and easier to maintain.

cmd/cmds_ls.go
```diff
@@ -19,11 +20,7 @@ var cmdsLsCmd = &cobra.Command{
 		if err != nil {
 			return err
 		}
-		evt, err := getString(cmd.Flags(), "event")
-		if err != nil {
-			return err
-		}
-
+		evt := v.GetString("event")
```
ramiresviana: error handling? silent error?

cmd/config_init.go
```diff
@@ -24,145 +25,52 @@ override the options.`,
...
 		s := &settings.Settings{
-			Key:                   key,
-			Signup:                signup,
-			HideLoginButton:       hideLoginButton,
-			CreateUserDir:         createUserDir,
-			MinimumPasswordLength: minLength,
-			Shell:                 convertCmdStrToCmdArray(shell),
+			Key:                   generateKey(),
+			Signup:                v.GetBool("signup"),
+			HideLoginButton:       v.GetBool("hideloginbutton"),
```
ramiresviana: **suggestion:** use camelCase when accessing viper values to keep consistent with the keys from the cobra flags.

cmd/utils.go
```diff
@@ -89,11 +76,34 @@ func dbExists(path string) (bool, error) {
...
+
+	return replacements
+}
+
 func python(fn pythonFunc, cfg pythonConfig) cobraFunc {
 	return func(cmd *cobra.Command, args []string) error {
+		v.SetEnvKeyReplacer(strings.NewReplacer(generateEnvKeyReplacements(cmd)...))
+
+		// Bind the flags
+		err := v.BindPFlags(cmd.Flags())
```
ramiresviana: precedence changes?

cmd/config_set.go
```diff
@@ -29,65 +28,67 @@ you want to change. Other options will remain unchanged.`,
...
 			case "baseurl":
-				ser.BaseURL, err = getString(flags, flag.Name)
+				ser.BaseURL = v.GetString(key)
 			case "root":
-				ser.Root, err = getString(flags, flag.Name)
+				ser.Root = v.GetString(key)
 			case "socket":
-				ser.Socket, err = getString(flags, flag.Name)
+				ser.Socket = v.GetString(key)
 			case "cert":
```
ramiresviana: **question:** is it desirable to use viper values on commands other than the root? It can cause changes in the config without user input. Eg.:  simply running `filebrowser config set` with `FB_SIGNUP=true`  environment variable will update the config value.

cmd/docs.go
```diff
@@ -40,7 +40,7 @@ var docsCmd = &cobra.Command{
 	Hidden: true,
 	Args:   cobra.NoArgs,
 	RunE: func(cmd *cobra.Command, _ []string) error {
-		dir, err := getString(cmd.Flags(), "path")
+		dir, err := cmd.Flags().GetString("path")
```
ramiresviana: not viper?

cmd/root.go
```diff
@@ -119,24 +144,21 @@ user created with the credentials from options "username" and "password".`,
...
-		}
+		workersCount := v.GetInt("imageprocessors")
 		if workersCount < 1 {
 			return errors.New("image resize workers count could not be < 1")
 		}
 		imgSvc := img.New(workersCount)
 
 		var fileCache diskcache.Interface = diskcache.NewNoOp()
-		cacheDir, err := cmd.Flags().GetString("cache-dir")
+		cacheDir, err := cmd.Flags().GetString("cachedir")
```
ramiresviana: **issue:** the correct key for accessing this value on the flagSet is `cacheDir`. Also, shouldn't viper be used instead?

cmd/root.go
```diff
@@ -37,8 +36,33 @@ import (
 
 var (
 	cfgFile string
+
+	flatNamesMigrations = map[string]string{
```
ramiresviana: **suggestion:** implement `flags.MarkDeprecated` to notify users using the old flag names.

cmd/utils.go
```diff
@@ -89,11 +76,34 @@ func dbExists(path string) (bool, error) {
...
+		if err != nil {
+			panic(err)
+		}
```
ramiresviana: **suggestion**: keep this viper initialization code on `root.go` by adding a  `PersistentPreRun` command property.

cmd/config_set.go
hacdias: That's a good question. I think using Viper everywhere is fine. The question is: do we want to use the environment variables everywhere?

All flags are prefixed in order to make it explicit that they belong to this program. `FB_SIGNUP` would only exist in the `config init` and `config set` commands. Why else would you have set `FB_SIGNUP`?

We could choose to only use the environment variable parsing in the root command for now, and then see later if it's something desirable to expand to other commands. What do you think?

cmd/config_init.go
hacdias: Done

cmd/root.go
hacdias: Done

cmd/root.go
hacdias: From my testing, I don't think that's possible while keeping compatibility with the new flag names. Instead, I added a message when the user tries to use an old flag name, which should be practically the same.

cmd/cmds_ls.go
hacdias: I have refactored `python` to initialize viper and pass it around.

cmd/utils.go
hacdias: I moved this to a function to init viper in order to pass it around as an instance, which would make the `PersistentPreRun` a bit harder unless we would make it a global variable.

cmd/config_set.go
ramiresviana: > All flags are prefixed in order to make it explicit that they belong to this program. FB_SIGNUP would only exist in the config init and config set commands. Why else would you have set FB_SIGNUP?

Note that viper uses values from config file in addition to the ones from env. What could happen is a user having mistakenly set a value accessible to viper and forgot about it. Since this behavior is not common on cli apps, I would choose to use only flag values.

cmd/root.go
```diff
@@ -35,11 +33,34 @@ import (
...
+		"img-processors":                   "imageProcessors",
+		"cache-dir":                        "cacheDir",
+		"token-expiration-time":            "tokenExpirationTime",
+		"baseurl":                          "baseURL",
+	}
 )
 
+func migrateFlagNames(f *pflag.FlagSet, name string) pflag.NormalizedName {
+	if newName, ok := flagNamesMigrations[name]; ok {
+		fmt.Printf("Flag --%s has been deprecated, use --%s instead\n", name, newName)
```
ramiresviana: **nitpick:**  running `filebrowser --img-processors` (no argument) prints the deprecation message multiple times.

cmd/config_set.go
hacdias: So your suggestion is to keep behavior mostly as-is:

- Use Viper only to parse the server settings (`getRunParams`), using flags > env > config.
- All other commands just use flags.

I'd agree with that approach, but I'd need to make a few changes. Right?

ramiresviana: Correct, feel free to leave for another time if too troublesome.

cmd/root.go
hacdias: Addressed

cmd/config_set.go
hacdias: This is now addressed

PR #5643, opened 2025-12-28
docs: clarify scope of config file and environment variables

bjackman:
I got pretty confused about why I can't set auth.method from the config file (and also why there were no docs saying what the JSON/YAML/etc key would even be).

Document that this doesn't work to try to prevent future confusion here.

Caveat: I wrote this documentation by hand, but I did not actually read the code carefully enough to have figured this out myself. I am trusting the claims made by an LLM coding agent (which do at least seem to match empirical evidence).

## Description

... Hopefully docs improvements is compatible with maintenance-only mode?

## Checklist

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -135,6 +135,11 @@ file named .filebrowser.{json, toml, yaml, yml} in the following directories:
 - $HOME/
 - /etc/filebrowser/
 
+**Note:** Only the options listed below can be set via the config file or
+environment variables. Other configuration options live exclusively in the
+database and so they must be set by the "config set" or "config
+import" commands.
+
 The precedence of the configuration values are as follows:
 
 - Flags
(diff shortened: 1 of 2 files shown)


hacdias (changes requested):
The CLI docs get generated from the CLI code itself. To change the CLI description, you need to update this file: https://github.com/filebrowser/filebrowser/blob/master/cmd/root.go#L113

bjackman:
Ah, sorry. Let me try to do this properly...

- Held my nose and [installed Docker](https://wiki.nixos.org/wiki/Docker#Rootless_Docker) (needn't have been so reluctant, looks like they fixed the thing where it runs everything as root!)
- `nix develop nixpkgs#filebrowser`
- `task docs:cli:generate`
- Yep that undid my changes
- Edit the Go that [hacdias](https://github.com/hacdias) pointed to
- `task docs:cli:generate` again

hacdias (approved):
Thanks!

PR #5655, opened 2026-01-01
fix: download path encoding file paths

### LINKED ISSUE #5651: When sharing a folder, the top download button does not work for a single file
### Checklist

### Version

```Text
2.53.0
```

### Description

There seems to be an escaping bug in the top bar menu `Download` button in a shared folder view when selecting a single file.

Image

The `Download` button works fine when selecting several file work fine and downloading a zip for example.

If you double-click the file and then use the `Download` button in the middle of the screen, it works fine.

It also does not seem to impact the main view top bar `Download` button, only the shared folder view is, apparently.

I'm not sure when this bug started exactly, but it's definitely older than 2.53.0.

The top bar menu `Download` button leads to https://files.mortagne.org/api/public/dl/FAIalCgD/1260831.png.

### What actually happened?

The top bar menu `Download` button leads to https://files.mortagne.org/api/public/dl/FAIalCgD%2F1260831.png.

### Reproduction Steps

* select a folder
* click the `Share` button from the top bar
* use the `Copy to clipboard` button to get the link (you are getting something like https://files.mortagne.org/share/FAIalCgD, which should work for 1 year if you want to see it in action)
* go to the URL
* select a file
* click the `Download` button from the top bar menu

### Files

_No response_

jagadam97:
## Description

in download function we are encoding file paths and double encoding files query params value.
Apache HTTP server by default doesn't allow encoded slashes in path and throws 404 error. Nginx by default sends the request to server.

this PR fixes the encoding so the filebrowser works seamlessly with Apache HTTP server and Nginx server without any manual configuration.

## Additional Information
I have tested this change locally with NGINX and Apache server 

closes #5651 

## Checklist

### DIFF
--- a/frontend/src/api/pub.ts
+++ b/frontend/src/api/pub.ts
@@ -41,12 +41,12 @@ export function download(
   let url = `${baseURL}/api/public/dl/${hash}`;
 
   if (files.length === 1) {
-    url += encodeURIComponent(files[0]) + "?";
+    url += files[0] + "?";
   } else {
     let arg = "";
 
     for (const file of files) {
-      arg += encodeURIComponent(file) + ",";
+      arg += file + ",";
     }
 
     arg = arg.substring(0, arg.length - 1);


hacdias (approved):
Thanks. I'm not sure why we were double encoding them, but it doesn't seem to make sense.

PR #5678, opened 2026-01-10
feat: add 'Open direct' button to images

imbaczek:
## Description

Add a new button 'Open direct' which opens images in a new tab directly in the browser, not in the UI.

## Additional Information

This is helpful when working with multiple large images - I'm dealing with \~3000x\~10000px images and need multiple tabs open with each.

## Checklist


hacdias:
Thanks. What's the use case here? And would it make sense adding it for other file types (e.g. audio, video)?

imbaczek:
> Thanks. What's the use case here?

I'm dealing with \~3000x\~10000px images - the HTML/JS viewers break in subtle ways on dimensions this big and the native browser viewers actually handle this better.

Also need multiple tabs open with each image and a one-click version helps.

> And would it make sense adding it for other file types (e.g. audio, video)?

It probably makes sense, I only tested with images since that's what I needed.

frontend/src/i18n/en.json
```diff
@@ -42,6 +42,7 @@
     "update": "Update",
     "upload": "Upload",
     "openFile": "Open file",
+    "openDirect": "Open direct",
```
hacdias: Could we maybe call this "View raw" or "Open raw in new tab", as well as put the button after/before the download button?

imbaczek: Took a while to get back to this, but both done

hacdias (approved):

PR #5691, opened 2026-01-15
fix: support for multiple encodings in CSV files

### LINKED ISSUE #5670: Accented characters are not handled correctly in CSV Viewer
### Checklist

### Version

```Text
2.51.2
```

### Description

Hello,

The CSV Viewer feature does not correctly handle accented characters.

When opening the following file, accented characters are displayed as invalid symbols (�).

Observed example

Image

In the CSV header:

`nom �tat date �dition p�riode libell� offre produit r�f�rence facture
`

Instead of:

`nom état date édition période libellé offre produit référence facture
`
Impact

Data is difficult to read, Loss of clarity for business fields (labels, addresses, descriptions, etc.), Potential issues during data analysis or export

- The issue is likely related to CSV file encoding (UTF-8 not properly detected, or missing support for ISO-8859-1 / Windows-1252).

### Additional Improvement – Column Separator

The Column Separator option is set to comma by default, which often leads to incorrect column parsing.

In addition
The option is located at the bot of the page

When opening large CSV files, it requires a lot of scrolling to reach it and change the value

Suggested improvement :

Set BOTH as the default separator or provide an option configurable by the administrator

Make the Column Separator option more accessible (sticky header, toolbar, or closer to the table) Or allow changing it without scrolling back to the top

Thank you for looking into this.

Expected behavior

Automatic encoding detection

### What actually happened?

Data is difficult to read, Loss of clarity for business fields (labels, addresses, descriptions, etc.), Potential issues during data analysis or export

### Reproduction Steps

File with accent

### Files

_No response_

ArielLeyva:
## Description
1. The column separator selector in the CSV file viewer was changed to the top.
2. An encoding selector was added to the CSV file viewer.
3. Support was added for various encoding types when reading file contents in the backend.
4. A new endpoint was added to the API that returns the different types of encoding available.
5. Set `Both (, and ;)` as the default column separator when opening a CSV file.
6. An existing error was corrected when editing CSV files a second time- Steps to reproduce: 
- Open CSV file
- Click on edit as text
- Close editor
- Reopen CSV file
- Click on edit as text 
- The window keeps loading and an error appears in the console)

PD:
7. I've replaced the native CSV parsing function with the `csv-parser` library.

This file can be used for testing:
[test.csv](https://github.com/user-attachments/files/24950800/test.csv)

## Additional Information
Closes https://github.com/filebrowser/filebrowser/issues/5670
image

## Checklist


hacdias:
At this point, I wonder if it's not easier to just use something out of the shelf, like https://github.com/adaltas/node-csv/tree/master/packages/csv-parse, which can handle delimiters, encodings, etc. We're sort of reinventing the wheel and keeping our own CSV parsing implementation, but does it make sense? Wdyt?

ArielLeyva:
> At this point, I wonder if it's not easier to just use something out of the shelf, like https://github.com/adaltas/node-csv/tree/master/packages/csv-parse, which can handle delimiters, encodings, etc. We're sort of reinventing the wheel and keeping our own CSV parsing implementation, but does it make sense? Wdyt?

I've replaced the native CSV parsing function with the `[csv-parser](https://github.com/adaltas/node-csv/tree/master/packages/csv-parse)` library; it seems clearer that way. However, it's necessary to maintain the endpoints for encodings and the function to specify the encoding of a text resource in the backend, since it always uses `UTF-8` by default, and some encodings don't allow for the resolution of certain special characters from the frontend. 

I think this same function could be extended when editing a file (since it currently always uses `UTF-8`, and special characters appear in the text editor in certain files with encodings specific to certain languages ​​not supported by `UTF-8`). 

Do you think passing the encoding as a "query parameter" in the URL would cause conflicts with other functions? Would it be better to specify the encoding through a header in the request, for example, `X-Resource-Encoding`?

hacdias:
I haven't looked very in depth to the code, but I think a separate endpoint is too much. I think using headers would be the best way. Just a thought:

1. Requests `/api/resource/whatever` with `X-Resource-Encoding: true`
2. Response with `X-Resource-Encoding: {encoding}`

Would something along those lines make sense?

ArielLeyva:
> I haven't looked very in depth to the code, but I think a separate endpoint is too much. I think using headers would be the best way. Just a thought:
> 
> 1. Requests `/api/resource/whatever` with `X-Resource-Encoding: true`
> 2. Response with `X-Resource-Encoding: {encoding}`
> 
> Would something along those lines make sense?

The idea behind creating a new endpoint, `/api/encodings`, would be to keep the list of supported encodings (which are managed by the package `golang.org/x/text`) in one place (backend). Later, if you want to remove or add support for any encoding, it would only require a single change to https://github.com/filebrowser/filebrowser/pull/5691/changes#diff-adf70ed3bd777e5310666a7602f040e2254ba94cdb846f94fea96b5d29559a51R12.

However, you could eliminate this endpoint and create a constant in the frontend with the encoding identifiers, while maintaining the dictionary in the backend to parse the encoding identifier coming from the frontend.

Regarding handling the encoding in the headers, it could look something like this: 
1. Requests `/api/resource/whatever` with `X-Resource-Encoding: windows-1252`
2. Response with `X-Resource-Encoding: windows-1252`

The idea is to be able to send from the frontend the encoding you want the backend to use to read the text file and return the content.

Does this approach seem correct to you?

hacdias:
@ArielLeyva that seems like a good idea. But I wonder... what if we get the byte stream _raw_ from the backend and use the frontend's `TextDecoder` interface and do all the encoding shenanigans just in the frontend: https://developer.mozilla.org/en-US/docs/Web/API/Encoding_API/Encodings - and then save it back as a byte stream. That way it'd be completely transparent to the backend. But only if it's widely available.

Otherwise we need to keep a list somewhere, whether it is on the frontend or backend, I'm fine with both. Just need to ensure that remains consistent.

ArielLeyva:
> @ArielLeyva that seems like a good idea. But I wonder... what if we get the byte stream _raw_ from the backend and use the frontend's `TextDecoder` interface and do all the encoding shenanigans just in the frontend: https://developer.mozilla.org/en-US/docs/Web/API/Encoding_API/Encodings - and then save it back as a byte stream. That way it'd be completely transparent to the backend. But only if it's widely available.
> 
> Otherwise we need to keep a list somewhere, whether it is on the frontend or backend, I'm fine with both. Just need to ensure that remains consistent.

That sounds interesting. We could start by integrating it only into the CSV file viewer (which isn't a critical feature), and if it works as expected, perhaps we can extend it to other text files. I'll get to work on it.

PR #5704, opened 2026-01-20
fix: allow moving and copying files to root directory

### LINKED ISSUE #5683: files cannot be moved/copied to root
### Checklist

### Version

```Text
v2.54.0
```

### Description

The error 403 Forbidden is displayed when trying to copy/move a file to the configured File Browser root directory.

The operation should be completed with no errors, like when uploading and renaming a file on the root directory.

### What actually happened?

A error is displayed and the files remain unchanged.

### Reproduction Steps

1. Create /folder/file
2. Move /folder/file to /

### Files

_No response_

dharunashokkumar:
fixes issue where users got 403 forbidden when trying to move or copy files to the configured root, even though uploads and renames worked fine there

Fixes #5683

This PR allows files and folders to be moved/copied to the root directory.

Previously, users received a 403 Forbidden error when trying to move or copy files to the configured FileBrowser root, even though uploads and renames worked fine there.

Changes:
- Removed the `dst == "/"` restriction in `fileutils/copy.go`
- Removed the `dst == "/"` check in `http/resource.go`
- Still prevents moving/copying FROM root for safety

The fix makes move/copy operations consistent with other file operations like upload and rename.

### DIFF
--- a/fileutils/copy.go
+++ b/fileutils/copy.go
@@ -18,8 +18,8 @@ func Copy(afs afero.Fs, src, dst string, fileMode, dirMode fs.FileMode) error {
 		return os.ErrNotExist
 	}
 
-	if src == "/" || dst == "/" {
-		// Prohibit copying from or to the virtual root directory.
+	if src == "/" {
+		// Prohibit copying from the virtual root directory.
 		return os.ErrInvalid
 	}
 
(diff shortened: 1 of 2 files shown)


ramiresviana:
This is not a backend bug but rather the frontend not including the filename in the destination path as expected by the API.

Issue #5481, opened 2025-10-10
409


gokturkmehmet:
### Checklist

- [x] This is a bug report, not a question.
- [x] I have searched on the [issue tracker](https://github.com/filebrowser/filebrowser/issues?q=is%3Aissue) for my bug.
- [x] I am running the latest [FileBrowser version](https://github.com/filebrowser/filebrowser/releases) or have an issue updating.

### Version

```Text
2.44.0
```

### Description

Got 409 conflict while uploading very small json files but more than 13000 in numbers. total size of files were around 50mb. Picture is attached. 

<img width="805" height="254" alt="Image" src="https://github.com/user-attachments/assets/cbe9506f-aa96-45b2-be17-b0c1b8b1edc7" />

### What did you expect to happen?

Expected that file uploading will succeed.

### What actually happened?

İt is still uploading. but how do i know what file was related to it? No info is given in dialog. 

### Reproduction Steps

Try to upload over 10 k small files.

gokturkmehmet:
A "more" button revealing what was the conflict may be sufficient to live with.

ramiresviana:
Could you please share the exact steps used to install and configure File Browser so I can try to reproduce the issue?

fumiama:
Same issue. It happens when I try to continue uploading hundreds of files within some folders without overriding. I have used F12 and checked the network status, while the "create folder" POST requests successfully returned OK, all the file POSTs returned `409 Conflict`. I remembered that I have upload those file and folders once and the network connection was down. After I resuming the WebUI and uploading it again, it starts to show these messages on completed files. I have also manually checked some md5s of those "conflict" files, it turns out that they are the same as the original file that I upload, meaning that the upload process should be succeeded.

P.S. I have clicked `Abort Upload` once when the network was down.

ramiresviana:
@fumiama if you think this affects our general users, please open a new issue with much detail as possible so we can reproduce and fix it.

Closing this one due to lack of feedback.

Issue #5453, opened 2025-09-22
409 CONFLICT


Oceans11Valentine:
### Checklist

- [x] This is a bug report, not a question.
- [x] I have searched on the [issue tracker](https://github.com/filebrowser/filebrowser/issues?q=is%3Aissue) for my bug.
- [x] I am running the latest [FileBrowser version](https://github.com/filebrowser/filebrowser/releases) or have an issue updating.

### Version

```Text
FILE BROWSER 8080
```

### Description

Keeps giving the 409 conflict error whilst downloading a flux model. happens after mid download running for about 15 to 20 minutes.

### What did you expect to happen?

i expected to download the files faster since i am running runpod.

### What actually happened?

Keeps giving the 409 conflict error whilst downloading a flux model. happens after mid download running for about 15 to 20 minutes.

### Reproduction Steps

Keeps giving the 409 conflict error whilst downloading a flux model. happens after mid download running for about 15 to 20 minutes.

morrisonmatt993-oss:
Having the same issue while trying to transfer files off a windows 7 device. Recently did a chassis swap will my drives and thought it had some thing misconfigured. I am also on the latest production of scale. have been attempted multiple times and file transfer fails every time.

ramiresviana:
Could you please provide detailed steps to reproduce the issue? Without this information, it’s not possible to work on a fix. Additionally, you haven’t specified which Filebrowser version you’re using. Note that a `409 Conflict` error is unrelated to file downloads—it only occurs during file creation or upload.

ramiresviana:
Closing due to lack of feedback.

PR #5724, opened 2026-01-26
feat: Add Redis upload cache for multi-replica deployments

arranpx:
## Description
This PR adds optional Redis support for the upload cache used by the tus handlers. This resolves `409 conflict` errors that can often occur in multi-replica deployments of FileBrowser when uploading files.

## Additional Information

**Background:**

I currently run filebrowser in K8s with more than one replica. Following a recent FileBrowser release, I noticed that on file upload I started to see a bunch of `409 conflict` errors when uploading files in my multi-replica environment.

I believe this issue has been reported previously, e.g. in #5453 and #5481, but I am not sure if these users were using multiple replicas like me.

**Root Cause:**

The issue was introduced in PR #5270, which implemented an in-memory cache to track uploads. In multi-replica deployments, this causes `409 conflict` errors when a completion API request is routed to a different replica than the one handling the upload, as each replica maintains its own separate in-memory cache without sharing state.

**Solution:**

This PR introduces optional Redis support for the upload cache, enabling cache sharing across replicas.

I have implemented this so that the Redis cache is optional, meaning backward compatibility is maintained - single-instance deployments continue to use the in-memory update cache.

To test this change, I have also added some docker compose files - one for running filebrowser on its own, and one for running it with the Redis cache.

**Next Steps**

Currently, I have only used Redis for the upload cache - however, for multi-replica deployments like mine, we probably also want to use it for the file cache and possibly configuration.

This being said, as I am only a single user that does not use file cache or update my configuration once initially run, I have not added this capability. 

It is possible I might add this at a future point in another PR - but this specific PR is targeted to the issue I am seeing to keep it small and focussed, and because it is blocking me from upgrading to the latest version of FileBrowser.

## Checklist

- [X] I am aware the project is currently in maintenance-only mode. See [README](https://github.com/filebrowser/community/blob/master/README.md)
- [X] I am aware that translations MUST be made through [Transifex](https://app.transifex.com/file-browser/file-browser/) and that this PR is NOT a translation update
- [X] I am making a PR against the `master` branch.
- [X] I am sure File Browser can be successfully built. See [builds](https://github.com/filebrowser/community/blob/master/builds.md) and [development](https://github.com/filebrowser/community/blob/master/development.md).

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -46,6 +46,7 @@ var (
 		"disable-type-detection-by-header": "disableTypeDetectionByHeader",
 		"img-processors":                   "imageProcessors",
 		"cache-dir":                        "cacheDir",
+		"redis-cache-url":                  "redisCacheUrl",
 		"token-expiration-time":            "tokenExpirationTime",
 		"baseurl":                          "baseURL",
 	}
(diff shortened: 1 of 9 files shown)


hacdias (approved):
Thanks! I think this is a valid case and a good solution for it.

PR #5731, opened 2026-01-30
fix: avoid showing 409 conflict when overwrite is false when uploading file

### LINKED ISSUE #5727: Uploading a file already existing leads to 409 errors
### Checklist

### Version

```Text
2.55.0 (Docker image filebrowser/filebrowser behind Traefik)
```

### Description

Same here v2.55.0

1. I partially upload a folder (connection interrupted or whatever)
2. I reupload the same folder, and I would like to ignore the already uploaded files, so I click on Continue

Image

3. I have a shitstorm of 409 errors

Image

Note : it is the same if I try to upload just one file already existing. I get a 409 if I click on CONTINUE.

Hence, clicking on continue always leads to 409.

The upload should Continue without errors.

### What actually happened?

I have 409 errors

### Reproduction Steps

Upload a file twice and click "CONTINUE" when prompted for Replace.

### Files

_No response_

ArielLeyva:
## Description
Adjustment to load error handling to prevent a 409 conflict message from being displayed when the user selects "Continue" (i.e., overwrite = false). In this case, the conflict is expected and should not be considered an error, so the user interface now resolves the load flow without displaying a misleading error message.

## Additional Information
Closes https://github.com/filebrowser/filebrowser/issues/5727

## Checklist


frontend/src/api/tus.ts
```diff
@@ -56,12 +56,22 @@ export async function upload(
...
-              : error.originalResponse.getBody()
-            : "Upload failed";
+        let message = "Upload failed";
+        if (error instanceof tus.DetailedError) {
+          const response = error.originalResponse;
+
+          if (response === null) {
+            message = "000 No connection";
+          } else {
+            if (response.getStatus() === 409 && !overwrite) {
```
ramiresviana, line 66: In a scenario that the frontend data is stale, the user is not be prompted of file conflicts and the error will not be displayed. The error should not be silenced unless explicitly chosen by the user.

frontend/src/api/tus.ts
```diff
@@ -56,12 +56,22 @@ export async function upload(
...
+              //do not display error if "continue" was pressed
+              resolve();
```
ramiresviana, line 68: Mocking the upload completion will still compute the bytes as sent, even that it didn't, making the speed indicator inaccurate. Consider updating the upload store instead.

ramiresviana (changes requested):

frontend/src/api/tus.ts
ArielLeyva: I could do something similar to how Windows Explorer handles conflicts when copying/moving files, where it displays a list of conflicting files (with information like the last modification date and size of both the source and destination files) and a checkbox so the user can decide which file to keep (basically whether to overwrite the file or not). I'm going to work on it.

daxid: Hello and thanks for your amazing work on Filebrowser !

It would be awesome to have a fix for this ASAP, without going into some new UI design (that can of course come later).

The use case at the moment is to continue an interrupted transfer with many files. Having a global "Continue" button is perfect for that. Just, if I clic on continue, I do not want to see the 409 errors. This is preventing us to ship the most wanted Filebrowser at the moment :-(

ArielLeyva: I am closing this pull request because I have opened another one with the agreed changes: https://github.com/filebrowser/filebrowser/pull/5765/

Issue #5723, opened 2026-01-26
Can unzip a file


tharilak37:
### Checklist

- [x] This is a bug report, not a question.
- [x] I have searched on the [issue tracker](https://github.com/filebrowser/filebrowser/issues?q=is%3Aissue) for my bug.
- [x] I am running the latest [FileBrowser version](https://github.com/filebrowser/filebrowser/releases) or have an issue updating.

### Version

```Text
latest as of 2026.01.26
```

### Description

I uploaded a zip file and can't find an unzip option. I checked github issues and saw that I have to enable a command. but, i didn't find such an option to enable commands

<img width="1280" height="1024" alt="Image" src="https://github.com/user-attachments/assets/3ad0351b-77b1-4e9f-b2c5-d71b0e1cf5ea" />

<img width="1280" height="1024" alt="Image" src="https://github.com/user-attachments/assets/15edea07-a237-4497-8ee1-7bb7baa72c7b" />

### What did you expect to happen?

give an option to unzip a zip file

### What actually happened?

tried unzipping a zip file and i found no option

### Reproduction Steps

upload a zip file, and try unzipping

ramiresviana:
There's no bult-in feature for uncompressing files and command execution is disabled by default, see https://filebrowser.org/command-execution.html

tharilak37:
Thank you for letting me know!

Issue #3529, opened 2024-10-11
Compress/Decompress files


rcarmo:
**Is your feature request related to a problem? Please describe.**

I'd like to be able to compress and decompress files and folders. There's always an instance where you need to compress a bunch of things to save space, or to get only a small specific file out of a .tar.gz or a .zip

**Describe the solution you'd like**

Add commands to do the above. Probably include common archive formats like .zip, .tar.gz, .7z, .rar, etc.

**Describe alternatives you've considered**

**Additional context**

Issue #3210, opened 2024-05-08
Does it support extract zip or archived files?


rikhtehgaran:
**Is your feature request related to a problem? Please describe.**
All file managers should be able to extract or zip files. it's supports zip only when we want to download

**Describe the solution you'd like**
Add features to zip or extract on the disk (not only for download or upload)

**Describe alternatives you've considered**
we have to upload or download 

**Additional context**
maybe it has a plugin or something. I just ask if it has these features

seternate:
Would like that feature as well 👍

Issue #2768, opened 2023-10-16
Add Ability to compress, extract files (zip, tar)


gzxiexl:
**Is your feature request related to a problem? Please describe.**

When I need only a small file in a huge archive, I can only download the archive and extract it locally.

**Describe the solution you'd like**

I would like to be able to support compression and decompression on the server side, and decompression to folders, so that there is no need to download the entire archive to the mobile side.

**Describe alternatives you've considered**

**Additional context**

PR #5746, opened 2026-02-06
feat: add zip extraction support

ArielLeyva:
## Description
This PR introduces a secure and configurable unzip handler to FileBrowser. It adds server‑level configuration options with sensible defaults and enforces multiple safety checks to prevent ZIP bombs, path traversal, and denial‑of‑service scenarios.
The zip file extraction function is disabled by default and can be enabled by using the `--unzipEnabled` flag or by enabling it through `filebrowser config set`.

### Key changes
1. New server configuration fields with defaults:
- `MaxZipFileSize` (default: 5GB): Maximum size of zip files that can be extracted (large files may cause DoS attacks).
- `MaxZipFileEntries` (default: 100k files): Maximum number of files in the zip archive (zip archives with too many files can trigger DoS attacks).
- `MaxTotalUncompressedSize` (default: 20GB): This is the disk space that the file will occupy once it is fully decompressed (a necessary check to avoid completely filling up the available disk space and crashing the system).
- `MaxUncompressedSizeRate` (default: 0.01 → 1%): The ratio between the size of each compressed and uncompressed file (prevents zip bomb attacks). Some malicious files (zip bombs) can "lie" about their uncompressed size, so every time a file is extracted from the zip archive, it is checked that the size it occupies on the disk matches the size declared in the zip file (if they don't match, the zip file "lies" and the extraction is stopped immediately because it is considered a corrupt or malicious file).
- `MaxUncompressedFileSize` (default: 5GB): Maximum size that each uncompressed file can occupy (avoids using up all available disk space and crashing the system).
2. Unzip handler implementation:
- Validates permissions and destination paths.
- Prevents Zip Slip attacks by sanitizing and rejecting absolute/relative escape paths.
- Enforces limits on compressed size, uncompressed size, number of entries, and compression ratio.
- `Uses io.LimitReader` to guard against infinite streams and oversized files.
- Provides clear error handling with user feedback.
3. Defaults integrated into server initialization so they can be overridden via configuration. 
- In other words, in older instances of filebrowser configurations where newer configurations do not exist, default values ​​are used to avoid incompatibilities with databases or configuration files from previous versions.

## Additional Information
Extracting compressed files is a highly requested feature in the project, and an optional, clear, and secure implementation is necessary for some users.

## Checklist


hacdias:
> Extracting compressed files is a highly requested feature in the project

Could you send me some examples?

ArielLeyva:
> > Extracting compressed files is a highly requested feature in the project
> 
> Could you send me some examples?

These issues discuss this function (https://github.com/filebrowser/filebrowser/issues/5723, https://github.com/filebrowser/filebrowser/issues/3210, https://github.com/filebrowser/filebrowser/issues/2768, https://github.com/filebrowser/filebrowser/issues/2316, https://github.com/filebrowser/filebrowser/issues/1541, and https://github.com/filebrowser/filebrowser/issues/3529).

I know that a partial solution has been found using "command execution," but its use hasn't been very clear and is somewhat risky. The idea I propose with this pull request is a decompression tool handled directly by Go, making it a bit safer and allowing this function to be readily available from the Filebrowser interface.

settings/settings.go
```diff
@@ -47,21 +53,27 @@ func (s *Settings) GetRules() []rules.Rule {
...
+	Address                  string  `json:"address"`
+	Log                      string  `json:"log"`
+	EnableThumbnails         bool    `json:"enableThumbnails"`
+	ResizePreview            bool    `json:"resizePreview"`
+	EnableExec               bool    `json:"enableExec"`
+	TypeDetectionByHeader    bool    `json:"typeDetectionByHeader"`
+	ImageResolutionCal       bool    `json:"imageResolutionCalculation"`
+	AuthHook                 string  `json:"authHook"`
+	TokenExpirationTime      string  `json:"tokenExpirationTime"`
+	UnzipEnabled             bool    `json:"unzipEnabled"`
```
hacdias, line 71: For consistency with other settings:

```suggestion
	UnzipEnabled             bool    `json:"enableUnzip"`
```

cmd/root.go
```diff
@@ -111,6 +111,7 @@ func addServerFlags(flags *pflag.FlagSet) {
 	flags.Bool("disableExec", true, "disables Command Runner feature")
 	flags.Bool("disableTypeDetectionByHeader", false, "disables type detection by reading file headers")
 	flags.Bool("disableImageResolutionCalc", false, "disables image resolution calculation by reading image files")
+	flags.Bool("unzipEnabled", false, "enable zip file extraction")
```
hacdias, line 114: All other flags are `--disable{Functionality}`. I'd rather stay consistent here, unless there's a good reason not to.

ArielLeyva: Oh right, I know it doesn't really follow the structure that has been used so far in the project; I would like to be able to implement it as you suggest. I did it the other way around because I didn't want to 'affect' users who were already running filebrowser and might not want this zip extraction feature for security reasons, but I don't think it will bother them. In the way you mention, if a user doesn't want this functionality, they can disable it with the flag `--disable{Functionality}`. If you tell me that the `--disable{Functionality}` approach is correct for the project, I will do it that way.

hacdias: Hmm... you do have a point. The only `disable` that is on by default is `--disableExec`, so that could be technically done.

However, making it `--disableX` would not affect existing installations I think. Because the `--disable` flag is an override. The default setting for **new** users would be True, but for existing users, the new field in the settings would still be False.

I think if we go with that approach, it'd work and be consistent. Other features have been added with time, making them enabled by default for new installations only.

hacdias:

PR #5753, opened 2026-02-12
fix: wrap response text in Error before reject

aarondill:
This was causing the error notification not to appear when failing to create a file (due to permissions) because the result was a string, not an Error

## Description

Fix a place where a string is rejected (thrown) and an Error is expected.

## Checklist


aarondill:
The Error is expected here: https://github.com/filebrowser/filebrowser/blob/master/frontend/src/components/prompts/NewFile.vue#L80
No error message is shown to the user because `e instanceof Error` is false, since it's the literal string `"403 Forbidden"`.

hacdias (approved):

Issue #5699, opened 2026-01-19
user dateFormat option seems not to affect the UI


mszalbach:
### Checklist

- [x] This is a bug report, not a question.
- [x] I have searched on the [issue tracker](https://github.com/filebrowser/filebrowser/issues?q=is%3Aissue) for my bug.
- [x] I am running the latest [FileBrowser version](https://github.com/filebrowser/filebrowser/releases) or have an issue updating.

### Version

```Text
2.55.0
```

### Description

I am trying to set the UI Option "Set exact date format" with the users add command.

For full reference I am creating filebrowser and the user the following way:

```#!/bin/sh
/bin/filebrowser config init --auth.method noauth -a [::] --root /srv -d /database/filebrowser.db -p 8080 --branding.name "Operations Log Access" --dateFormat
/bin/filebrowser users add admin adminadminadmin --perm.modify=false \
                                    --perm.create=false \
                                    --perm.delete=false \
                                    --perm.execute=false \
                                    --perm.rename=false \
                                    --perm.share=false \
                                    --lockPassword \
				    --dateFormat=true \
                                    -d /database/filebrowser.db
exec /bin/filebrowser -d /database/filebrowser.db
```

I also tried --dateFormat without anything.  As reference the option --hideDotfiles works as expected.

### What did you expect to happen?

I expect the UI option "Set exact date format" in the admin profile to be enabled and the Last Modified UI date to be absolute and not relative anymore.

### What actually happened?

The server starts with no errors or warnings the "Set exact date format" is not clicked and all modified times are relative instead of absolute.

### Reproduction Steps

You can use the example repo with the docker compose file:
https://github.com/mszalbach/filebrowser-dateFormat

mszalbach:
Installed new version v2.61.2 and works for me.
Thanks for the fast review and merge.

PR #5804, opened 2026-03-05
fix: added dateFormat to getUserDefaults so this is respected in the …

mszalbach:
This pull request should solve https://github.com/filebrowser/filebrowser/issues/5699

The problem was that users.go getUserDefaults method did not copy the dateFormat from the flags.
I also searched for other locations which uses hideDotfiles but did not have dateFormat.

I did not find a suitable test to extend to show it works, so if you can point me to the right file I can try to add one.
I tested it with the steps described in my issue and afterwards the user setting in the UI was set and files where shown as absolute.

I am not sure I understand all impact of the files I changed and if it is always correct so also use the dateFormat where the hideDotfiles was used.


azdfzshffg:
@hacdias  Can you check pls ?

I have the same issue with my users. They are asking for a more precise date format, and each of them has to enable the option manually. Unfortunately, I can't set it as the default.

PR #5822, opened 2026-03-08
fix: properly surface config parse errors

### LINKED ISSUE #5821: Malformed config file incorrectly reported as "no config being used"
### Checklist

### Version

```Text
v2.61.2
```

### Description

If the configuration file exists but is invalid, filebrowser reports that no configuration file is being used instead of surfacing the parsing error from Viper.

If the configuration file exists but is invalid, the application should report the configuration parsing error.

### What actually happened?

The application reports that no configuration file is being used, masking the underlying parse error.

### Reproduction Steps

1. Run `echo 'root = /unquoted/path' > .filebrowser.toml`
2. Run `go run .`
3. Observe output

### Files

_No response_

apkatsikas:
## Description

When the configuration file was malformed, the application would incorrectly report that no configuration was being used.

This happened because the error check used `errors.Is` instead of `errors.As`, which prevented `viper.ConfigParseError` from being properly detected.

This change fixes the error detection so configuration parsing errors are correctly surfaced to the user.

## Additional Information

Closes https://github.com/filebrowser/filebrowser/issues/5821

## Checklist

### DIFF
--- a/cmd/utils.go
+++ b/cmd/utils.go
@@ -118,7 +118,9 @@ func initViper(cmd *cobra.Command) (*viper.Viper, error) {
 
 	// Read in configuration
 	if err := v.ReadInConfig(); err != nil {
-		if errors.Is(err, viper.ConfigParseError{}) {
+
+		var parseErr viper.ConfigParseError
+		if errors.As(err, &parseErr) {
 			return nil, err
 		}
 
--- a/cmd/utils.go
+++ b/cmd/utils.go
@@ -119,8 +119,7 @@ func initViper(cmd *cobra.Command) (*viper.Viper, error) {
 	// Read in configuration
 	if err := v.ReadInConfig(); err != nil {
 
-		var parseErr viper.ConfigParseError
-		if errors.As(err, &parseErr) {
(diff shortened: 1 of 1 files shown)


cmd/utils.go
```diff
@@ -118,7 +118,9 @@ func initViper(cmd *cobra.Command) (*viper.Viper, error) {
 
 	// Read in configuration
 	if err := v.ReadInConfig(); err != nil {
-		if errors.Is(err, viper.ConfigParseError{}) {
+
+		var parseErr viper.ConfigParseError
+		if errors.As(err, &parseErr) {
```
hacdias: Wouldn't this work?

```suggestion
		if errors.Is(err, &viper.ConfigParseError{}) {
```

apkatsikas: @hacdias yes, I tried this previously and it worked - I will push the one-liner update.

apkatsikas: https://github.com/filebrowser/filebrowser/pull/5822/changes/3b69e77d187f2178bb2f4c7346ab2b4c1b1ef339 fixed.

PR #5832, opened 2026-03-14
feat: add optional directory sizes in listings

posalex:
## Description

Add an optional per-user setting to calculate and display directory sizes in file listings.

Today File Browser shows file sizes but not useful directory sizes in the listing view. This change adds a new profile setting, `showDirectorySizes`, which is disabled by default. When enabled, directory sizes are calculated recursively in the backend and shown in the UI. The calculated values are also used when sorting by size.

## Additional Information

The implementation was added to the existing backend listing pipeline instead of shelling out to a platform-specific command such as `du`. That keeps the feature portable, testable, and consistent with the current file metadata flow.

The setting is stored on the user model, passed through defaults/auth payloads, exposed in Profile Settings, and consumed by the listing UI. Directory size calculation respects the existing checker/rule filtering so hidden or blocked entries are not counted. Symlinked directories are not traversed.

The feature is intentionally opt-in because recursive size calculation can be expensive on large directory trees.

I also added backend tests for directory size calculation and size-based sorting.

This PR includes minimal UI string additions in `en.json` and `de.json` only because the new setting requires labels in the interface. It is not intended as a translation-focused change.

## Checklist

### DIFF
--- a/auth/hook.go
+++ b/auth/hook.go
@@ -233,11 +233,12 @@ func (a *HookAuth) GetUser(d *users.User) *users.User {
 			Asc: a.Fields.GetBoolean("user.sorting.asc", d.Sorting.Asc),
 			By:  a.Fields.GetString("user.sorting.by", d.Sorting.By),
 		},
-		Commands:     a.Fields.GetArray("user.commands", d.Commands),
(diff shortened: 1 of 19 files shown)


hacdias:

PR #5848, opened 2026-03-18
fix: double slash in TUS upload path when readEntries returns multiple batches

### LINKED ISSUE #5845: Double slash in TUS upload path when readEntries returns multiple batches
## Bug

When uploading files via drag-and-drop into a folder that contains enough files for the browser to split `FileSystemDirectoryReader.readEntries()` across multiple batches, the constructed upload path contains double slashes (`//`). This causes nginx (and other reverse proxies) to 301-redirect the TUS POST, which converts it to a GET per HTTP spec, resulting in a 404.

## Root cause

In `frontend/src/utils/upload.ts`, `readReaderContent()` appends `/` to the directory on **both** the `readEntry()` call and the recursive `readReaderContent()` call:

```typescript
function readReaderContent(reader, directory) {
  reader.readEntries((entries) => {
    if (entries.length > 0) {
      for (const entry of entries) {
        readEntry(entry, `${directory}/`);        // appends /
      }
      readReaderContent(reader, `${directory}/`);  // appends / AGAIN
    }
  });
}
```

The `readEntries()` API returns entries in batches. On the first call, `directory` is e.g. `"TestFolder"`, so entries get `"TestFolder/"` — correct. But the recursive call passes `"TestFolder/"`, so the **second batch** of entries gets `"TestFolder//"` — double slash.

## Observed behavior

```
POST /api/tus/.../08.07.2025//FeldAI_SampleFiles_P01784/file.xlsx → 301 (nginx merges //)
GET  /api/tus/.../08.07.2025/FeldAI_SampleFiles_P01784/file.xlsx  → 404
```

Upload fails. Only affects directories with enough files to trigger multi-batch `readEntries()`.

## Reproduction test

Extracts the path-building logic from `scanFiles()` and simulates two batches:

```typescript
import { describe, it, expect } from "vitest";

type Entry = { name: string; isFile: boolean; isDirectory: boolean };

function simulateScanFiles(dirName: string, entryBatches: Entry[][]): string[] {
  const paths: string[] = [];
  let batchIndex = 0;

  function readEntry(entry: Entry, directory = ""): void {
    if (entry.isFile) {
      paths.push(`${directory}${entry.name}`);
    }
  }

  // Mirrors readReaderContent() from upload.ts lines 118-138
  function readReaderContent(directory: string): void {
    const entries = batchIndex < entryBatches.length ? entryBatches[batchIndex] : [];
    batchIndex++;
    if (entries.length > 0) {
      for (const entry of entries) {
        readEntry(entry, `${directory}/`);
      }
      readReaderContent(`${directory}/`);
    }
  }

  readReaderContent(dirName);
  return paths;
}

describe("scanFiles path construction", () => {
  it("should not produce double slashes when readEntries returns multiple batches", () => {
    const paths = simulateScanFiles("TestFolder", [
      [
        { name: "file1.xlsx", isFile: true, isDirectory: false },
        { name: "file2.xlsx", isFile: true, isDirectory: false },
      ],
      [
        { name: "file3.xlsx", isFile: true, isDirectory: false },
      ],
    ]);
    expect(paths).toHaveLength(3);
    for (const p of paths) {
      expect(p, `path "${p}" contains double slash`).not.toContain("//");
    }
  });
});
```

```
FAIL: path "TestFolder//file3.xlsx" contains double slash
  expected 'TestFolder//file3.xlsx' not to contain '//'
```

## Suggested fix

The recursive call should not re-append `/` since the first iteration already added it:

```typescript
function readReaderContent(reader, directory) {
  reader.readEntries((entries) => {
    if (entries.length > 0) {
+     const dirWithSlash = directory.endsWith("/") ? directory : `${directory}/`;
      for (const entry of entries) {
-       readEntry(entry, `${directory}/`);
+       readEntry(entry, dirWithSlash);
      }
-     readReaderContent(reader, `${directory}/`);
+     readReaderContent(reader, dirWithSlash);
    }
  });
}
```

## Environment

- filebrowser v2.61.2
- Behind nginx ingress (Kubernetes)
- Triggered by drag-and-drop upload of a folder with ~20+ files

fabboe:
## Summary

Fixes #5845

`readReaderContent()` in `frontend/src/utils/upload.ts` appends `/` to the directory path on **both** the `readEntry()` call and the recursive `readReaderContent()` call. The browser's `FileSystemDirectoryReader.readEntries()` returns entries in batches — you must call it repeatedly until it returns an empty array. On the second batch and beyond, the directory accumulates extra slashes (e.g. `TestFolder//file3.xlsx`).

This causes reverse proxies (nginx, Cloudflare, etc.) to 301-redirect the TUS POST request to normalize the double slash, which per HTTP spec converts POST → GET, resulting in a 404.

## Fix

Normalize the trailing slash once before iterating entries and recursing:

```typescript
const dirWithSlash = directory.endsWith("/") ? directory : `${directory}/`;
```

## Test

Adds vitest as a dev dependency, a `test` script, and a CI job (`Test Frontend`). The test extracts the path-building logic from `scanFiles()` and verifies no double slashes appear when `readEntries()` returns multiple batches.

## Test plan
- [x] `pnpm run test` passes locally (2/2 tests green)
- [x] Verified fix in production behind nginx ingress — large folder uploads no longer produce 301/404

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -27,6 +27,24 @@ jobs:
           pnpm install --frozen-lockfile
           pnpm run lint
 
+  test-frontend:
+    name: Test Frontend
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v6
+      - uses: pnpm/action-setup@v4
+        with:
+          package_json_file: "frontend/package.json"
+      - uses: actions/setup-node@v6
+        with:
+          node-version: "24.x"
+          cache: "pnpm"
+          cache-dependency-path: "frontend/pnpm-lock.yaml"
+      - working-directory: frontend
+        run: |
+          pnpm install --frozen-lockfile
+          pnpm run test
+
   lint-backend:
     name: Lint Backend
     runs-on: ubuntu-latest
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -14,7 +14,8 @@
     "typecheck": "vue-tsc -p ./tsconfig.app.json --noEmit",
     "lint": "eslint src/",
     "lint:fix": "eslint --fix src/",
-    "format": "prettier --write ."
+    "format": "prettier --write .",
+    "test": "vitest run"
   },
   "dependencies": {
     "@chenfengyuan/vue-number-input": "^2.0.1",
@@ -70,6 +71,7 @@
     "typescript": "^5.9.3",
     "vite": "^8.0.0",
     "vite-plugin-compression2": "^2.3.1",
+    "vitest": "^4.1.0",
     "vue-tsc": "^3.1.3"
   },
   "packageManager": "pnpm@10.32.1+sha512.a706938f0e89ac1456b6563eab4edf1d1faf3368d1191fc5c59790e96dc918e4456ab2e67d613de1043d2e8c81f87303e6b40d4ffeca9df15ef1ad567348f2be"
--- a/frontend/pnpm-lock.yaml
+++ b/frontend/pnpm-lock.yaml
@@ -162,6 +162,9 @@ importers:
       vite-plugin-compression2:
         specifier: ^2.3.1
         version: 2.5.1(rollup@4.57.1)
+      vitest:
+        specifier: ^4.1.0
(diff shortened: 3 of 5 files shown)


hacdias (approved):
Thank you!

PR #5850, opened 2026-03-19
fix: touch Redis upload cache key on GetLength to prevent TTL expiry

arranpx:
## Description

When using the Redis upload cache, big file uploads (that take longer than 2 minutes in total) fail with a `409 - Conflict`.

On file upload in FileBrowser, the file is registered in an upload cache with a TTL (of 2 minutes) so FileBrowser can track all the in-progress uploads. The TTL is naturally set so that if the upload fails, the file key is eventually removed from the upload cache.

When using the `ttlcache` (in-memory) upload cache, `GetLength` will also automatically touch the key, extending its expiry. However, with Redis, running a `GET` does not automatically extend the TTL. This means `GetLength` was never touching the key with this upload cache.

Additionally, when uploading, FileBrowser uses tus to upload the file in 10MB chunks via a PATCH. When this PATCH is handled, a goroutine is started which will touch the key in the upload cache every two seconds, extending the TTL to ensure the key does not expire while an upload is in progress.

However, because the keep-alive goroutine only touches the key after an initial two-second delay — if a 10MB chunk uploads in under two seconds (e.g. on a fast network and/or fast storage), the goroutine is cancelled before the key is ever touched. This means the TTL is never extended, leading to eventual key expiry after 2 minutes — which causes a `409 - Conflict` error.

## Additional Information

N/A

## Checklist

### DIFF
--- a/http/tus_handlers.go
+++ b/http/tus_handlers.go
@@ -18,8 +18,6 @@ import (
 
 // keepUploadActive periodically touches the cache entry to prevent eviction during transfer
 func keepUploadActive(cache UploadCache, filePath string) func() {
-	cache.Touch(filePath)
-
 	stop := make(chan bool)
 
 	go func() {
--- a/http/tus_handlers.go
+++ b/http/tus_handlers.go
@@ -18,6 +18,8 @@ import (
 
 // keepUploadActive periodically touches the cache entry to prevent eviction during transfer
 func keepUploadActive(cache UploadCache, filePath string) func() {
+	cache.Touch(filePath)
+
 	stop := make(chan bool)
 
 	go func() {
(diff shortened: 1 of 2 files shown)


ramiresviana:
Isn't the TTL being reset by this code? A cache retrieval should already perform a touch.

https://github.com/filebrowser/filebrowser/blob/6d44b3ae1d8d26484380bd90d8eb90a08a7fc062/http/tus_handlers.go#L186

arranpx:
Ah ha! This is a good point - I just looked at the `ttlcache` docs and indeed it does say the following for `Get()` used by `GetLength()`:

>  it also extends/touches an item's expiration timestamp on successful retrieval. If the item is not found, a nil value is returned.

But this has actually highlighted the actual source of my issue then! I am using the redis upload cache, whose `Get()` does not touch the key. This means the fix I need to implement is different. 

Let me update the PR

arranpx:
Updated the PR - please can I request a new review @ramiresviana 🙏

ramiresviana (approved):

PR #5858, opened 2026-03-22
fix: language change issue

LiDaoKuan:
## fix: language change issue

When the user changes the language on the frontend, the interface language does not change.

## Checklist

1. fixed: When the user changes the language on the frontend, the interface language does not change.
    now it can be changed when user click the Update Button in setting page.
2. fixed: Every time the user login, the interface language is English, not the language previously set by the user.
    I found out that the reason is that when the user login, the locale field in the data passed from the backend to the frontend is null, so the frontend uses the default locale field: English. Now I have changed it so that the locale field in the settings is passed from the backend to the frontend every time the user logs in.
    
## Additional Information

there is a same parblem in theme change：when use click the Update Button to change page theme, the theme will indeed be changed. But if user reflesh the page or relogin, the theme will revert to the default setting (Dark).

<mark>**I didn't fix the issue of theme changes, but only fixed the problem of language changes. **</mark>

### DIFF
--- a/frontend/src/views/settings/Global.vue
+++ b/frontend/src/views/settings/Global.vue
@@ -251,6 +251,7 @@ import { StatusError } from "@/api/utils";
 import Rules from "@/components/settings/Rules.vue";
 import Themes from "@/components/settings/Themes.vue";
 import UserForm from "@/components/settings/UserForm.vue";
+import { setHtmlLocale, setLocale } from "@/i18n";
 import { useLayoutStore } from "@/stores/layout";
 import { enableExec } from "@/utils/constants";
 import { getTheme, setTheme } from "@/utils/theme";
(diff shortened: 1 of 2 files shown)


ramiresviana:
Values from "User default settings" are used when new users are created and should have no impact on existing users. I'm also unable to confirm that "data passed from the backend to the frontend is null", could you please give more information?

LiDaoKuan:
Like you said, the default user settings are only for new users – they shouldn't affect existing ones at all. My bad, I mixed up the new user settings with existing user settings, which is why I thought "the backend sent nothing to the frontend."
I re-cloned the master branch code and tested the frontend again. Here's what's actually happening:
When logging in, the frontend sends a GET /api/settings request. The backend response does have the locale field:
2ebecbc65e39f31c54cee3ccf2d184cf
c15ab114c9992a7af7d6e5f31e12a4b1
So I'm guessing maybe we forgot to call setLocale() after getting the settings? Or maybe it's not triggered at the right time?

ramiresviana:
I'm closing this PR as the proposed changes are invalid. Feel free to open an issue report if you're facing a problem.

Issue #5877, opened 2026-03-31
Error on docker compose run


geourjoa:
### Checklist

- [x] This is a bug report, not a question.
- [x] I have searched on the [issue tracker](https://github.com/filebrowser/filebrowser/issues?q=is%3Aissue) for my bug.
- [x] I am running the latest [FileBrowser version](https://github.com/filebrowser/filebrowser/releases) or have an issue updating.

### Version

```Text
2.62.2
Commit 860c19ddf58665aafcf976f5299a907738214226 on master branch

Ubuntu 24
```

### Description

Hello,

I was unable to launch project with docker compose (I was able by running vite + go).

To launch I do : 

```bash 
task build
docker compose up
```

Redis container start but filebrowser container failed with this error :
```
filebrowser  | /init.sh: exec: line 35: filebrowser: not found
filebrowser exited with code 127
```

I t seem related to `docker/alpine/alpine.sh` script.

I have find a way to fix it, PR [5878](https://github.com/filebrowser/filebrowser/pull/5878)

 

### What did you expect to happen?

Application start

### What actually happened?

Filebrowser container not start 

### Reproduction Steps

To launch I do : 

```bash 
task build
docker compose up
```

hacdias:
See comment on #5878

PR #5878, opened 2026-03-31
fix: Fix docker compose deployment

geourjoa:
Related to #5877

### DIFF
--- a/Taskfile.yml
+++ b/Taskfile.yml
@@ -19,6 +19,8 @@ tasks:
 
   build:backend:
     desc: Build backend binary
+    env:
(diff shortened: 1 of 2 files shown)


hacdias:
Thanks.

The Dockerfile that we have is really meant to be used on the CI when cross-building the app for multiple platforms, and it may be confusing for people when building using `task build` on e.g. macOS, and then running Docker with a Linux environment.

As such, I just updated the compose file to use the published Docker image (see 2f805de5274514c627f61e2b648fc7a9b8eadb2c).

No other changes are needed.

PR #5890, opened 2026-04-04
fix: restrict default permissions for proxy-auth auto-provisioned users

kodareef5:
## Summary

- Apply the same permission restrictions to proxy-auth auto-created users as the signup handler (commit b6a4fb1)
- Strip `Admin`, `Execute`, and `Commands` after applying defaults, since auto-provisioned accounts should not inherit execution capabilities
- Adds test verifying restricted permissions on proxy-auth user creation

Ref: GHSA-7526-j432-6ppp

### DIFF
--- a/auth/proxy_test.go
+++ b/auth/proxy_test.go
@@ -14,8 +14,7 @@ type mockUserStore struct {
 }
 
 func (m *mockUserStore) Get(_ string, id interface{}) (*users.User, error) {
-	switch v := id.(type) {
-	case string:
+	if v, ok := id.(string); ok {
 		if u, ok := m.users[v]; ok {
 			return u, nil
 		}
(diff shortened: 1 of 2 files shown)


auth/proxy_test.go
```diff
@@ -0,0 +1,80 @@
...
+	users map[string]*users.User
+}
+
+func (m *mockUserStore) Get(_ string, id interface{}) (*users.User, error) {
+	switch v := id.(type) {
+	case string:
+		if u, ok := m.users[v]; ok {
+			return u, nil
+		}
+	}
```
hacdias:
```suggestion
	if v, ok := id.(string); ok {
		if u, ok := m.users[v]; ok {
			return u, nil
		}
	}
```

PR #5892, opened 2026-04-05
fix: add header to prevent the creation of new users in proxy authentication

### LINKED ISSUE #3900: Proxy Auth creates user even if "Allow users to signup" is disabled
**Description**
When using proxy authentication, if a user logs in without an account, a new account is automatically created even if the "Global Settings" -> "Allow users to signup" is NOT selected.

**Expected behaviour**
HTTP Error 404 Not Found, or something similar.

The user account is created, despite "Allow users to signup" being disabled.

**Additional context**
You could argue this is by design, because the user is already authenticated, but not for my use case. And why have an "Allow users to signup" option, but not check it.

**How to reproduce?**
If you already have auth set to proxy, then it is easy to test, just login with a non-existent user and one will automatically get created. You need a reverse proxy server to test this, or adding a HTTP host header to simulate the proxy.

**Files**

ArielLeyva:
## Description
This PR allows a header to be sent in proxy authentication so that user creation can be prevented if the username sent by the proxy does not exist in the database.

The proxy can send the `X-Auth-No-Create-User` header with the value `true` to prevent the user from being created if it does not exist in the database; if the user does not exist, filebrowser authentication is not passed.

The explanation of this behavior and the use of the `X-Auth-No-Create-User` header was added to the documentation.

## Additional Information
Closes https://github.com/filebrowser/filebrowser/issues/3900

## Checklist


hacdias:
@ArielLeyva thanks for this. Can we also make it clear that "Allow users to signup" doesn't apply to Proxy auth?

ArielLeyva:
> @ArielLeyva thanks for this. Can we also make it clear that "Allow users to signup" doesn't apply to Proxy auth?

I've added a note to the proxy authentication method documentation to explain this behavior.

hacdias:
I've been thinking about this. Wouldn't it be maybe better to add it as a flag configuration for the proxy authentication? I'm not sure if the header is the right way to go.

PR #5940, opened 2026-05-17
fix: preserve ASCII filenames in downloads

### LINKED ISSUE #5939: Download a .gpx file, FileBrowser adds .xml extension
### Checklist

### Version

```Text
2.63.3
```

### Description

When I download files from an Android browser with the .gpx extension, they are being renamed to have .gpx.xml as the extension.

The file should be downloaded with the exact same name it had when I uploaded it.

### What actually happened?

The file is being renamed to abc.gpx.xml

This breaks an app on my phone, because it will only import files with the .extension.

### Reproduction Steps

1. Login to FileBrowser on my desktop PC (Ubuntu24.04, latest Waterfox browser).
2. Upload a file named abc.gpx. (file is correctly named in FB at this point)
3. Login to FileBrowser on my phone (Android 16, latest IronFox browser).
4. Locate the file, select Download. 
5. I am asked if I want to save the file named abc.gpx.xml

### Files

This is a XML file (in its contents) but the file name should not change when I download it.

mturac:
## Description

Fixes #5939.

Adds a classic `filename="..."` fallback for ASCII download names while keeping the existing RFC5987 `filename*` value for UTF-8 filenames. Some Android browsers appear to fall back to MIME/content sniffing when only `filename*` is present, which can make XML-based `.gpx` downloads save as `.gpx.xml`.

Non-ASCII filenames continue to use `filename*` only.

## Verification

- `go test ./http -run TestSetContentDisposition`
- `git diff --check`

### DIFF
--- a/http/raw.go
+++ b/http/raw.go
@@ -8,6 +8,7 @@ import (
 	"net/url"
 	gopath "path"
 	"path/filepath"
+	"strconv"
 	"strings"
 
 	"github.com/mholt/archives"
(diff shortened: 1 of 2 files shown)


hacdias (approved):

hacdias:
The other fix has been applioed (#5942), which I hope suffices.

PR #5944, opened 2026-05-17
feat: support custom share keys

kakadais:
## Context
This change stays within the existing share-link workflow and does not introduce a new subsystem, config surface, or deployment/runtime behavior.

The goal is to let operators create stable, human-readable share URLs when needed, while preserving the current autogenerated-key behavior for all existing callers.

## What changes
- accept an optional caller-provided share key during share creation
- keep autogenerated share keys as the default when no key is provided
- reject duplicate active keys with `409 Conflict`
- allow reuse of a key that only belongs to an expired share
- validate keys before persistence and keep collision handling for autogenerated keys

## Why this should be low-maintenance
- scope is limited to the existing share creation path
- no auth model change, migration, config, or storage backend expansion
- backward compatible for existing clients that do not send a key
- frontend change is a single optional input in the existing share prompt
- backend behavior is covered with focused tests for invalid, duplicate, and expired-key reuse cases

## Validation
- `go test ./...`
- `npm exec --prefix frontend -- vue-tsc -p frontend/tsconfig.app.json --noEmit`
- `npm exec --prefix frontend -- vite build`

## Scope / privacy check
- PR diff only contains files required for this share-link behavior
- no deployment, compose, local path, webhook, or personal/server-specific config is included


kakadais:
Superseded by #6011, which rebases the custom share-key feature onto current master and removes the unrelated compose change that made this PR dirty/conflicting.

PR #5960, opened 2026-05-24
feat: Archive support for Filebrowser

Kosztyk:
# Changelog: Archive Browsing and Extraction Support

## Summary

This pull request adds archive support to File Browser so users can open supported archive files directly in the web UI, browse their contents like a virtual read-only folder, preview supported files inside the archive, download individual files from inside the archive, and explicitly extract archives or selected archive entries.

The implementation uses the existing `github.com/mholt/archives` dependency already present in the project, avoiding shelling out to external tools such as `unzip`, `tar`, `7z`, or `rar`.

### Archive browsing

Users can now open supported archive files from the normal file listing. When an archive is opened, File Browser displays its contents as a virtual folder structure.

Supported first-pass archive formats include:

- `.zip`
- `.tar`
- `.tar.gz`
- `.tgz`
- `.tar.bz2`
- `.tbz2`
- `.tar.xz`
- `.txz`
- `.tar.zst`
- `.tzst`
- `.7z`
- `.rar`

Archive browsing uses URL query state instead of treating archive entries as real filesystem paths. Example:

```text
/files/frontend.zip?archive=/frontend/src/
```

This keeps archive paths separate from real filesystem paths and avoids ambiguity between a real folder and a file inside an archive.

### Virtual archive API

A new archive API layer was added for listing, streaming, and extracting archive contents.

New API behavior:

```text
GET  /api/archive/list/<archive-path>?inner=<archive-inner-path>
GET  /api/archive/raw/<archive-path>?inner=<archive-inner-file>
POST /api/archive/extract/<archive-path>
```

The list endpoint returns archive contents in a resource-like shape so the existing frontend listing components can render archive folders and files with minimal changes.

The raw endpoint streams a single file from inside an archive so existing preview and download flows can work with archive entries.

The extract endpoint extracts either:

- the full selected archive when the user selects an archive file outside archive view
- a selected file or folder from inside an already-opened archive

### Preview support inside archives

Supported file types can now be previewed directly from inside archives, including:

- images
- audio/video
- PDF files
- CSV files
- text files
- source/config files

Preview URLs are routed through the new archive raw endpoint when the current view is inside an archive.

### Download support inside archives

Users can download individual files from inside an archive without extracting the full archive first.

When the selected item is inside an archive, the download action now targets the archive raw endpoint with the selected inner path.

### Extraction actions

The UI now exposes extraction actions in both toolbar and context-menu flows.

Outside an archive:

- selecting an archive file shows **Extract this archive**
- right-clicking an archive file shows **Extract this archive**

Inside an archive:

- selecting a file shows **Extract this file**
- selecting a folder shows **Extract this folder**
- right-clicking an archive entry shows the matching extract action
- previewing a file inside an archive also exposes **Extract this file**

Extraction is explicit. Opening or browsing an archive never writes files to disk.

### Frontend routing

Archive state is represented by the `archive` query parameter.

Example:

```text
/files/frontend.zip?archive=/frontend/
```

This allows the existing file browser route to remain the main navigation entry point while distinguishing normal filesystem browsing from virtual archive browsing.

### Read-only behavior inside archives

Archive contents are treated as read-only.

The following write actions are disabled while browsing inside an archive:

- upload
- edit/save
- rename
- move
- copy into archive
- delete from archive

This prevents File Browser from implying that archive contents can be modified in place.

### Toolbar behavior

The toolbar now adapts depending on context:

- normal selected archive file: shows archive extraction action
- selected file inside archive: shows file extraction action
- selected folder inside archive: shows folder extraction action
- archive entries keep download behavior where valid

### Context menu behavior

The right-click menu now includes extraction actions for:

- archive files selected outside archive view
- files selected inside archive view
- folders selected inside archive view

### Breadcrumb behavior

Breadcrumbs now understand archive navigation and keep the archive file as the root of the virtual archive path.

## Security and safety

The archive implementation includes safeguards to reduce archive-related risks.

Implemented protections:

- archive browsing is read-only
- inner archive paths are normalized
- path traversal attempts are rejected
- unsafe `../` paths are blocked
- extraction validates output paths before writing
- extracted entries must remain inside the selected destination
- symlinks, device files, named pipes, sockets, and irregular files are skipped during extraction
- preview size is capped for text-like files
- archive file access still goes through File Browser's existing user permission model

Archive browsing and file streaming require the user to have permission to access/download the archive file itself.

## Files added

```text
http/archive.go
```

## Files modified

```text
files/file.go
files/utils.go
http/http.go

frontend/src/api/files.ts
frontend/src/types/file.d.ts
frontend/src/views/Files.vue
frontend/src/views/files/FileListing.vue
frontend/src/views/files/Preview.vue
frontend/src/views/files/Editor.vue
frontend/src/components/Breadcrumbs.vue
frontend/src/components/files/ListingItem.vue
frontend/src/css/listing-icons.css
```

## Implementation notes

The implementation reuses `github.com/mholt/archives`, which was already present in the project for archive creation/download functionality.

This keeps archive support inside the Go application and avoids relying on system-level binaries that may not be available in all deployments or containers.

Archive browsing is implemented as a virtual filesystem layer. Archive entries are not extracted during navigation.

Extraction remains a separate user-triggered operation.

## Validation performed

The following validation was performed during patch preparation:

- Go files were formatted with `gofmt`
- the generated source archive was checked for ZIP integrity
- the full patch applies cleanly to the original source tree
- the incremental UI patch applies cleanly to the previous archive-support version

## Validation not performed

A full local build/test run was not completed in the sandbox environment because this File Browser source requires Go `>= 1.25.0`, while the available sandbox Go version is older and external toolchain download is blocked.

Recommended maintainer-side validation:

```bash
go test ./...
npm install
npm run build
```

or the equivalent frontend package-manager commands used by the project.

## User-facing result

After this change, users can:

1. click an archive file to open it
2. browse folders inside the archive
3. preview supported files inside the archive
4. download individual files from inside the archive
5. extract a full archive from the normal file list
6. extract a selected file or folder while browsing inside an archive

### DIFF
--- a/files/file.go
+++ b/files/file.go
@@ -35,22 +35,25 @@ var (
 // FileInfo describes a file.
 type FileInfo struct {
 	*Listing
-	Fs         afero.Fs          `json:"-"`
-	Path       string            `json:"path"`
-	Name       string            `json:"name"`
-	Size       int64             `json:"size"`
-	Extension  string            `json:"extension"`
-	ModTime    time.Time         `json:"modified"`
-	Mode       os.FileMode       `json:"mode"`
-	IsDir      bool              `json:"isDir"`
-	IsSymlink  bool              `json:"isSymlink"`
-	Type       string            `json:"type"`
-	Subtitles  []string          `json:"subtitles,omitempty"`
-	Content    string            `json:"content,omitempty"`
-	Checksums  map[string]string `json:"checksums,omitempty"`
-	Token      string            `json:"token,omitempty"`
-	currentDir []os.FileInfo     `json:"-"`
-	Resolution *ImageResolution  `json:"resolution,omitempty"`
+	Fs               afero.Fs          `json:"-"`
+	Path             string            `json:"path"`
+	Name             string            `json:"name"`
+	Size             int64             `json:"size"`
+	Extension        string            `json:"extension"`
+	ModTime          time.Time         `json:"modified"`
+	Mode             os.FileMode       `json:"mode"`
+	IsDir            bool              `json:"isDir"`
+	IsSymlink        bool              `json:"isSymlink"`
+	Type             string            `json:"type"`
+	Subtitles        []string          `json:"subtitles,omitempty"`
+	Content          string            `json:"content,omitempty"`
+	Checksums        map[string]string `json:"checksums,omitempty"`
+	Token            string            `json:"token,omitempty"`
+	currentDir       []os.FileInfo     `json:"-"`
+	Resolution       *ImageResolution  `json:"resolution,omitempty"`
+	Archive          bool              `json:"archive,omitempty"`
+	ArchivePath      string            `json:"archivePath,omitempty"`
+	ArchiveInnerPath string            `json:"archiveInnerPath,omitempty"`
 }
 
 // FileOptions are the options when getting a file info.
@@ -241,6 +244,9 @@ func (i *FileInfo) detectType(modify, saveContent, readHeader bool, calcImgRes b
 	}
 
 	switch {
+	case IsArchiveName(i.Name):
+		i.Type = "archive"
+		return nil
 	case strings.HasPrefix(mimetype, "video"):
 		i.Type = "video"
 		i.detectSubtitles()
--- a/files/utils.go
+++ b/files/utils.go
@@ -2,6 +2,7 @@ package files
 
 import (
 	"os"
+	"strings"
 	"unicode/utf8"
 )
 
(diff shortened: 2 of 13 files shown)


lapisxalpha:
This is incredible, and I hope it's accepted, but I do have one thing to say... for my use case, not being able to extract file structures with intact symlinks renders this feature unworkable (I can't create hundreds of unknown symlinks by hand), so I'd have to use something else on-device. It would be really nice to be able to sit at a desktop browser and do it. I understand that anyone's effort working on this comes at a premium, and it is appreciated very much as-is; but perhaps a checkbox to for permission to allow it or other special files makes sense? Thank you very much for your consideration.

viscory:
so uh what needs to happen for this to be merged in?

hacdias:
> so uh what needs to happen for this to be merged in?

I'd kindly refer you to the project status: https://github.com/filebrowser/filebrowser#project-status

hacdias:

PR #5996, opened 2026-06-16
[codex] Feature 1: optimize runtime performance

aeltawela:
## Summary

- avoid file header reads for known extensions in file type detection
- avoid duplicate image resolution work while building directory listings
- pass request context into preview resizing so canceled requests stop resize work

## Why

These are hot paths on low-resource deployments such as Raspberry Pi 5 systems. The changes keep existing functionality intact while reducing repeated disk I/O, image decode work, and wasted CPU after clients disconnect.

## Validation

- `go test ./files ./http`


aeltawela:
Closing because these feature PRs should live in the forked repository aeltawela/filebrowser instead.

PR #6011, opened 2026-07-02
feat: support custom share keys

kakadais:
## Context
This replaces the closed/conflicting #5944 with the same feature rebased onto current `master`. It stays within the existing share-link workflow and does not introduce a new subsystem, config surface, or deployment/runtime behavior.

The goal is to let operators create stable, human-readable share URLs when needed, while preserving the current autogenerated-key behavior for all existing callers.

## What changes
- accept an optional caller-provided share key during share creation
- keep autogenerated share keys as the default when no key is provided
- reject duplicate active keys with `409 Conflict`
- allow reuse of a key that only belongs to an expired share
- validate keys before persistence and keep collision handling for autogenerated keys

## Why this should be low-maintenance
- scope is limited to the existing share creation path
- no auth model change, migration, config, or storage backend expansion
- backward compatible for existing clients that do not send a key
- frontend change is a single optional input in the existing share prompt
- backend behavior is covered with focused tests for invalid, duplicate, and expired-key reuse cases

## Validation
- `go test ./...`
- `CI=true pnpm --config.dangerouslyAllowAllBuilds=true run build`

## Scope / privacy check
- PR diff only contains files required for this share-link behavior
- no deployment, compose, local path, webhook, or personal/server-specific config is included

## Routing note
The root catch-all routing issue is not included here because upstream already fixed it in #5955 (`fix(router): handle undefined catchAll param on root redirect`).


hacdias:

PR #6028, opened 2026-07-16
feat: add a refresh button to the folder listing header

superafun:
## Summary
Add a "Refresh" action to the folder listing header so users can manually reload the current directory.

- Reuses the existing `fileStore.reload` pipeline (the same mechanism used after an upload), so no new state or API call is introduced.
- Uses the existing `buttons.refresh` translation key and the `refresh` icon.

## Test plan
- Open any directory in the file listing.
- Click the new refresh button in the header; the current directory is reloaded.

## Checklist
- [x] Targets `master`
- [x] Frontend-only change, no backend/i18n additions needed (`buttons.refresh` already exists)

### DIFF
--- a/frontend/src/views/files/FileListing.vue
+++ b/frontend/src/views/files/FileListing.vue
@@ -59,6 +59,12 @@
           @action="switchView"
         />
         <action
+          id="refresh-folder-button"
+          icon="refresh"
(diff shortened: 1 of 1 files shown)


hacdias:

PR #6033, opened 2026-07-16
fix: accessibility and security improvements

superafun:
## Summary
A collection of accessibility and security fixes.

**Accessibility (frontend)**
- Add `alt` text to images: header logo, file thumbnails, share preview and the extended image viewer, so screen readers can describe them.
- Add a visually-hidden `<h1>` landmark in `App.vue`.
- Use `aria-pressed` instead of `aria-selected` for toggleable list items (`ListingItem`, `FileList`).
- Allow user zoom by removing `user-scalable=no` from the viewport meta.
- Add `-webkit-user-select: none` alongside existing `user-select: none` rules.

**Security (backend)**
- Set `X-Content-Type-Options: nosniff` on static responses to prevent MIME sniffing.

**Maintenance**
- Replace the deprecated `router.beforeResolve` `next()` callback with return values (Vue Router 4).

## Test plan
- Run the app with a screen reader / axe; images now expose names.
- Zoom the page (Ctrl/Cmd +); it is no longer blocked.
- Visit a static asset response and confirm the `X-Content-Type-Options: nosniff` header is present.

## Notes
- `frontend/index.html` (an unused legacy entry) was intentionally left untouched; the canonical Vite entry is `frontend/public/index.html`.

### DIFF
--- a/frontend/public/index.html
+++ b/frontend/public/index.html
@@ -5,7 +5,7 @@
     <meta http-equiv="X-UA-Compatible" content="IE=edge" />
     <meta
       name="viewport"
-      content="width=device-width, initial-scale=1, user-scalable=no"
+      content="width=device-width, initial-scale=1"
     />
 
     [{[ if .ReCaptcha -]}]
--- a/frontend/src/App.vue
+++ b/frontend/src/App.vue
@@ -1,5 +1,6 @@
 <template>
   <div>
+    <h1 class="sr-only">{{ name }}</h1>
     <router-view></router-view>
   </div>
 </template>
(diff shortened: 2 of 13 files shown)


hacdias:
Is `-webkit-user-select: none` necessary nowadays?

superafun:
Good catch — `-webkit-user-select` was only needed for Safari ≤ 12 and very old iOS WebKit, well below our supported baseline. Dropped the prefix in all 5 spots; the standard `user-select` declarations stay. Pushed as `0f2955a6` — ready for re-review.

PR #6053, opened 2026-07-27
fix: enforce rules on recursive operations and expired proxy tokens

hacdias:
Fixes two reported advisories, one commit per fix with its regression test alongside.

### Recursive operations ignore rules denying descendants (GHSA-77x8-73f4-5485)

Copy, rename and delete authorized only the root of the operation and then recursed through the whole subtree, so a rule denying a descendant of an allowed directory was bypassed by operating on the parent: denied files were readable at the copy destination, or gone after the delete. The archive download walker already checks every descendant.

The handlers now walk the source tree — and the location that tree would land on — before mutating anything, and reject the whole operation with 403 if any path is denied. The walk consults the rules only: `HideDotfiles` is a display preference and must not stop a user from moving a directory that contains a dotfile. Deployments with no rules configured skip the walk entirely.

### Expired JWTs accepted under proxy auth (GHSA-v3jv-rmh2-635j)

Under proxy auth with a non-default logout page the JWT expiration is waived, because the proxy — not the token — owns the session lifetime. The waiver was granted on the strength of the token alone, inside the shared `withUser` middleware, so a token stolen before it expired kept authenticating every route, administrative ones included, and could be refreshed through `/api/renew` indefinitely.

The expiration is now waived only when the configured proxy header still names the user the token was issued for, which the trusted proxy asserts on any genuine request. The seamless-session behaviour from #3884 is unaffected.

### Verification

Both regression tests fail on the pre-fix code and pass after. `go build ./...`, `go vet ./...` and `go test ./...` are clean.

### DIFF
--- a/http/auth.go
+++ b/http/auth.go
@@ -66,7 +66,7 @@ func (e extractor) ExtractToken(r *http.Request) (string, error) {
 	return "", request.ErrNoTokenInRequest
 }
 
-func renewableErr(err error, d *data) bool {
+func renewableErr(err error, r *http.Request, d *data, tk *authToken) bool {
 	if d.settings.AuthMethod != fbAuth.MethodProxyAuth || err == nil {
 		return false
 	}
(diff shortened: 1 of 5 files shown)


mholt:
🫡 Great work Henrique. This project leaves a legacy in its wake. It was great collaborating with you all those years ago and seeing the progress and your skills as a developer. Good job and my hat's off to you! God bless you in your future endeavors :)

